PR #16, opened 2018-11-06
Demo mode tweaks

### LINKED ISSUE #15: FlowAuth demo mode usability tweaks
Couple of tweaks we've identified that would make the FlowAuth demo mode a bit nicer:

- [ ] Check all route and aggregation unit boxes by default when adding a new server
- [ ] Set the default latest token expiry as tomorrow, rather than _now_

greenape:
Closes #15

### I have:

### Description

This makes two changes to the create server page: 
- Routes and aggregation unit checkboxes checked by default when running in demo mode
- Latest expiry date for tokens defaults to tomorrow, not now


flowauth/backend/flowauth/admin.py
```diff
@@ -52,12 +52,22 @@ def list_all_capabilities():
     -----
     Responds with {<capability_name>: {"id":<capability_id>, "permissions":{<right>:False}}}
     """
+    default_capacity_setting = current_app.config["DEMO_MODE"]
```
maxalbert: Minor suggestion: just to keep the terminology consistent, how about renaming this to `default_permission` (or `default_permission_setting`), because the JSON key is also called `"permissions"`? Or is "capacity" used elsewhere in the code?

greenape: Capability is used elsewhere (e.g. `flow` is a capability), but this would be better named `default_permission_setting` 👍

flowauth/backend/flowauth/admin.py
```diff
@@ -52,12 +52,22 @@ def list_all_capabilities():
     -----
     Responds with {<capability_name>: {"id":<capability_id>, "permissions":{<right>:False}}}
     """
+    default_permission_setting = current_app.config["DEMO_MODE"]
```
maxalbert: One more observation: there is no indication of the _type_ here. One has to go to `flowauth/backend/config.py` to see that this is actually a boolean value - and thus be able to verify that it's ok to insert this into the permissions dict further below in the file.

I know it's more verbose, but how about
```
default_permission_setting = bool(current_app.config["DEMO_MODE"])
```
(or even be as explicit as
```
default_permission = True if (current_app.config["DEMO_MODE"] == True) else False
```
)?

I know it's redundant and less elegant and but it requires less cognitive overhead for the reader to verify the code. Just a suggestion though.

PR #23, opened 2018-11-06
Nicer formatting of example notebook in the docs

### LINKED ISSUE #22: No distinction between inputs and outputs in docs notebook
The notebook `example_usage.ipynb` is automatically included in the docs (at https://flowminder.github.io/FlowKit/analyst/example_usage/), but cell inputs and outputs are rendered identically.

**Expected behavior**
There should be a distinction between inputs and outputs.

jch1g10:
Closes #22 

### I have:

### Description

I have added headings and comments to `example_usage.ipynb`, and formatted the outputs of Jupyter notebook cells as blockquotes when they appear in the docs. I have also added a `get_status` function to the client, as a more human-friendly function for checking the status of a query.

@greenape


greenape:
Didn't we just take the demo folder _out_ in #4?

maxalbert:
Nice additions, thank you! For the time being I suggest only keeping a single version of the example usage notebook because the existing docs one and the new demo one are fairly similar (the only difference is how flowclient is installed and how the token is generated). I expect we will evolve the existing one into a more full walkthrough in the near future anyway.

jch1g10:
> Nice additions, thank you! For the time being I suggest only keeping a single version of the example usage notebook because the existing docs one and the new demo one are fairly similar (the only difference is how flowclient is installed and how the token is generated). I expect we will evolve the existing one into a more full walkthrough in the near future anyway.

Good suggestion. I've removed the demo notebook, and updated the title to better describe what this pull request is actually doing.

docs/source/analyst/example_usage.ipynb
```diff
@@ -103,36 +185,41 @@
...
    "cell_type": "code",
-   "execution_count": 9,
+   "execution_count": 10,
    "metadata": {},
    "outputs": [
     {
      "data": {
       "text/plain": [
-       "(False, <Response [202]>)"
+       "(True, <Response [303]>)"
```
maxalbert: Isn't the output of `get_status()` supposed to be a string now? Looks like some of the cells need to be re-executed maybe?

docs/source/analyst/example_usage.ipynb
```diff
@@ -314,23 +402,33 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "dl1 = flowclient.daily_location(date=\"2016-01-01\", aggregation_unit=\"admin3\", daily_location_method=\"last\", subscriber_subset=None)\n",
-    "dl2 = flowclient.daily_location(date=\"2016-01-02\", aggregation_unit=\"admin3\", daily_location_method=\"last\", subscriber_subset=None)"
+    "# dl1 = flowclient.daily_location(\n",
```
greenape: Hmm. Actually, can we either remove this cell, or uncomment it?

docs/source/analyst/example_usage.ipynb
jch1g10: Yes, that output was left over from before I added `get_status()`. I've cleared all outputs now, and checked that the output of `get_status()` is correct when it runs during the docs build.

docs/source/analyst/example_usage.ipynb
jch1g10: Good point. I think it's useful to show there are two ways to create a modal location, so I've kept this cell in and uncommented it.

PR #28, opened 2018-11-07
Use require for logo instead of static serving. Fixes #25

### LINKED ISSUE #25: Flowauth does not display logo in login screen
**Product**
Flowauth

**Expected behavior**
The flowauth login screen should display the logo.

**Screenshots**
screen shot 2018-11-06 at 17 51 35

greenape:
Closes #25

### I have:

### Description

This makes the login screen logo be properly included in the build process, rather than stuck in a location not accessible via the web server when running in the docker container.


flowauth/frontend/package-lock.json
```diff
@@ -221,7 +221,7 @@
     },
     "ansi-escapes": {
       "version": "3.1.0",
-      "resolved": "https://registry.npmjs.org/ansi-escapes/-/ansi-escapes-3.1.0.tgz",
+      "resolved": "http://registry.npmjs.org/ansi-escapes/-/ansi-escapes-3.1.0.tgz",
```
maxalbert: Is there a reason that these URLs are not using https any more? Given several recent incidents in the Javascript world I'd really prefer these to use secure connections for package downloads if at all possible. Will things break if we switch this back?

greenape: Seems to be this: https://github.com/npm/npm/issues/20719

Which the NPM people at least think isn't worth worrying about.

maxalbert: Hmm, how is it _not_ worth worrying about someone potentially feeding you compromised versions of npm packages via a man-in-the-middle attack? 🤔

Again, if the installation still works with the https versions of the URLs I suggest simply reverting this change and not including it in the PR.

maxalbert: Oh hang on, I may have mis-read what these links represent... What the "resolved" links actually used for?

greenape: This is a pretty good explainer: https://github.com/npm/npm/issues/16849#issuecomment-312442508 but the short version would seem to be that these links aren't actually used to download the packages, and the protection is afforded via the integrity hash.

Fwiw, there are ~70 http:// links in the lock file before this as well ;)

maxalbert: Fair enough. ;) Ok, with this background information I'm ok to leave it in as this is not security relevant (although still slightly annoying).

PR #29, opened 2018-11-07
Don't expose ports externally (except API) in `docker-compose.yml`

jch1g10:
### I have:

### Description

It's not necessary to expose any ports externally, other than the API. I have updated `docker-compose.yml` to expose FlowDB, FlowMachine and Redis ports only to other Docker containers in the network.

@greenape


docker-compose.yml
```diff
@@ -13,8 +13,8 @@ services:
     flowdb-testdata:
         container_name: flowdb_testdata
         image: flowminder/flowdb:testdata-master 
-        ports:
-            - 9000:5432
+        expose:
```
greenape: Do we need the expose? From https://docs.docker.com/compose/compose-file/#expose it sounds like this is for use with `links`, but not needed when using `networks`?

jch1g10: Ah, maybe not. I'll try without.

PR #34, opened 2018-11-07
Fix overeager server page transition

### LINKED ISSUE #32: FlowAuth doesn't always show newly added or edited servers after clicking Save
**Describe the bug**
If I add a new server and click save, I usually need to click away from the Servers page of the admin menu to get it to show up. Same issue when changing the name of a server.

**Product**
FlowAuth

**Version**
master

**Expected behavior**
Server is immediately present, changes to the name show up right away.

**Additional context**

Related to this, if the server creation fails for some reason, you get no notification of that.

greenape:
Closes #32 

### I have:

### Description

This fixes newly added servers not showing up, and edits to server names not being visible right away.

#### Backend

- Adds an `InvalidUsage` exception, to be used in place of a generic 400 error which allows setting an error message and supplying a json payload.
- Uses the `InvalidUsage` exception in place of all `abort(400, ..)`s
- Uses the `InvalidUsage` exception to return errors when creating a server with no name, a name that already exists, or a name that is too long. Also flags the specific field that's causing the problem in the json payload.
- Makes the server's name unique at db level
- Adds tests of these failures

#### Frontend

- Waits for the create or edit to succeed before hopping back to the server list
- Highlights the server name field red when that's caused an error
- Doesn't go back to the server list if there _was_ an error

The basic pattern here can then be used wherever else validation of the input data is going to happen server side, and an obvious immediate next step would be to display the long form message sent back in the frontend UI. Haven't done that here, because we should give a little though as to how that should look.


flowauth/backend/flowauth/admin.py
```diff
@@ -235,6 +236,18 @@ def add_server():
     json["latest_token_expiry"] = datetime.datetime.strptime(
         json["latest_token_expiry"], "%Y-%m-%dT%H:%M:%S.%fZ"
     )
+    if "name" not in json:
+        raise InvalidUsage("Must provide server name", payload={"bad_field": "name"})
+    if len(json["name"]) == 0:
+        raise InvalidUsage("Must provide server name", payload={"bad_field": "name"})
```
maxalbert: This is a duplicate message to above. Do you want to mention "non-empty" here?

maxalbert: Sorry, wasn't thinking straight... Of course this message makes sense here and is sufficient. ;)

flowauth/backend/flowauth/admin.py
```diff
@@ -499,9 +512,9 @@ def edit_group_servers(group_id):
             )
             limits.longest_life = server["max_life"]
         if limits.longest_life > server_obj.longest_token_life:
-            abort(400, "lifetime_too_long")
+            raise InvalidUsage("lifetime_too_long")
```
maxalbert: Minor nitpick, but do we want to remove the underscores to be consistent (and slightly more human-readable) with the invocations above? No strong opinion though.

greenape: Oh definitely - basically all of these need to be changed to `InvalidUsage("Verbose error message", payload={"bad_field":"blah"})` or similar, but I felt like this PR was already getting too big ;)

maxalbert:
Very nice changes, thanks a lot!

Also super useful to have the nice breakdown of changes in the PR description, makes it very easy to follow what's going on. 👍 

Looks great, just one inline question about how non-401 errors are handled, otherwise I'm happy to merge (and we should change the `InvalidUsage` messages in a separate PR - or feel free to include in this one if you like, since we're changing the relevant lines already anyway).

PR #46, opened 2018-11-15
Remove large rasters

greenape:
- Removes two large raster files, and an extra set of admin2 boundaries from the FlowDB test data container.
- Adds a tiny raster for testing raster statistics functionality
- Converts test_raster_statistics to pytest form

I've also reordered the image builds on CircleCI, to avoid not finishing builds in the background before hitting the point where they need to be uploaded to Docker cloud.

After merging this, we'll also want to clean the large rasters out of the git history.


flowmachine/tests/test_raster_statistics.py
```diff
@@ -5,106 +5,92 @@
...
+def test_computes_expected_clipping_values(get_dataframe):
     """
-    Series of tests for the RasterStatistics() utility
-    methods.
+    RasterStatistics() returns correct values when clipping vector and raster layers.
     """
+    G = "admin2pcod"
+    vector = Table(schema="geography", name="admin2")
+    r = RasterStatistics(
+        raster="population.worldpop_2015_un", vector=vector, grouping_element=G
```
maxalbert: I'd very much prefer to not call this table `worldpop_2015_un` (since it's now just a test raster). Also, maybe we should move this raster into a specific `testing` schema as opposed to keeping it in the `population` schema that's meant for real data? In any case, I'm happy to defer these changes (if they are desired) to a separate PR.

flowmachine/tests/test_raster_statistics.py
```diff
@@ -5,106 +5,92 @@
...
-    def test_computes_expected_clipping_values(self):
-        """
-        RasterStatistics() returns correct values when clipping vector and raster layers.
-        """
-        G = "district_c"
-        vector = Table(schema="public", name="gambia_admin2")
+    result = get_dataframe(r)  # Should have only _one_ entry
+    assert 1 == len(result)
+    assert "524 3 07" in result.admin2pcod.values
+    assert 2500000.0 == result.set_index("admin2pcod").loc["524 3 07"][0]
```
maxalbert: It's a bit of a bummer that with this simple raster we're not really testing clipping (since the result is the same as for the total value below). But again, I'm happy to defer any changes to this to a separate PR (e.g. happy to use my existing example from last week that I haven't turned into a PR yet).

greenape: Clipping is tested in the test circles ones, which use a circle geom to clip. Could rename the test - the expected behaviour is that we only get rows where there’s overlap between the raster and the vector.

flowmachine/tests/test_raster_statistics.py
```diff
@@ -5,106 +5,92 @@
...
+    vector = Table(schema="geography", name="admin2")
+    with pytest.raises(ValueError):
         r = RasterStatistics(
-            raster="worldpop_gambia", vector=vector, grouping_element="district_c"
+            "population.worldpop_2015_un", vector=vector, grouping_element=None
         )
-        assert r.head(0).columns.tolist() == r.column_names
+
+
+def test_failure_with_no_grouping_layer():
```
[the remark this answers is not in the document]
greenape: I must not have really looked at it when swapping it to pytest- this one should just be removed I think,

greenape:
@maxalbert happy for you to make any tweaks you think are necessary here. :+1:

maxalbert:
@greenape Just saw your message. Thanks for the quick reply! I'll add a couple of minor tweaks and will then merge.

flowmachine/tests/test_raster_statistics.py
maxalbert: Hmm, I see that the circles test does exercise the clipping behaviour, but I think we _also_ want to test this in the RasterStatistics class because (i) the fact that Circles uses RasterStatistics internally is an implementation detail which could in principle change in the future; (ii) if we ever decide to change or remove Circles we'd lose this test for RasterStatistics (and it's not guaranteed developers will know/remember to add it there). I suggest to use a slightly larger raster which overlaps with multiple admin regions, but again let's do this in a separate PR.

PR #50, opened 2018-11-19
Tweak flowdb ports

### LINKED ISSUE #49: Multiple flowdb instances on different ports in docker-compose-dev.yml
During development, it can be useful to have more than one flowdb instance running on different ports (e.g. for benchmarking with different data sizes, or to have access to the test data but also have a flowdb instance available that isn't pre-populated with any data).

I suggest having three flowdb services available in `docker-compose-dev.yml`:
```
- flowdb:  port 9000, without any pre-populated data
- flowdb-testdata:  port 9001, with the data currently used for tests
- flowdb-synthetic-data:  port 9002, with the larger synthetic daa
```
These ports should also be controllable using environment variables (`FLOWDB_PORT`, `FLOWDB_TESTDATA_PORT`, `FLOWDB_SYNTHDATA_PORT`).

maxalbert:
Closes #49

### I have:

### Description

This PR makes the following tweaks to `docker-compose-dev.yml`:

- Adds a `flowdb` service which runs a bare-bones flowdb instance (on port 9000 by default) without any pre-populated data.
- Changes the default ports for `flowdb-testdata` and `flowdb-synthetic-data` to 9001 and 9002, respectively.
- Allows setting the ports for the three flowdb instances via environment variables `FLOWDB_PORT`, `FLOWDB_TESTDATA_PORT`, `FLOWDB_SYNTHDATA_PORT`.
- The `flowdb` service is now the only one that persists its postgres data to a volume (the testdata and synthdata ones simply store them inside the container, i.e. any changes to the data are lost if the container is removed).

I don't think any changes are _required_ for the CircleCI config because we're setting all the env variables in `.circleci/config.yml` explicitly, but should we tweak some of the settings to reflect the changes in this PR? @greenape


docker-compose-dev.yml
```diff
@@ -10,11 +10,33 @@ networks:
...
+      - API_PASSWORD=${API_PASSWORD:-foo}
+    tty: true
+    stdin_open: true
+    restart: always
+    volumes:
+      - data_volume_flowdb:/var/lib/postgresql/data
+    networks:
+      db:
+        aliases:
+          - flowdb
```
greenape: Is these both having the same alias going to cause chaos if you bring both up at once?

maxalbert: Doesn't look like it. I can bring them all up and access them on the various ports via flowmachine without apparent.

maxalbert: Actually, on second look it appears that if I `ping flowdb` from within the flowmachine container (while all three flowdb instances are spun up), it is not predictable _which_ of them is being pinged (and this can change randomly after a while).

Do we rely on the alias being available somewhere? I think I'd prefer to take these aliases out for the time being because otherwise it may cause some _really_ hard to track down bugs... Any objections? Happy to re-think if we _do_ need the alias name somewhere.

josh-gree:
Why would you ever run both at the same time?

maxalbert:
@josh-gree I can see multiple useful scenarios. For example, if you want to run benchmarks against different data sizes without having to tear down and re-create the flowdb container all the time. Also, if you're experimenting (or benchmarking against) different schema layouts using the same flowmachine code, it's useful to just be able to swap out the database underneath by connecting to a different port.

.circleci/config.yml
```diff
@@ -60,7 +60,7 @@ defaults:
         username: $DOCKER_CLOUD_USER
         password: $DOCKER_CLOUD_PASSWORD
       environment:
-        N_DAYS: 3
+        N_DAYS: 7
```
greenape: Can we change this back? It was dropped to 3 because:

- It allows us to check the env var is getting passed on correctly (7 is the default when you spin up)
- It reduces the time to run the tests

maxalbert: I don't think the test runtime is noticeably affected (since we still create only a relatively small amount of rows per day), but checking that the env var is passed in is a good point. I'll change it back.

maxalbert:
Ok, I have updated the PR. Any further suggestions @greenape?

PR #52, opened 2018-11-19
Fix available dates

### LINKED ISSUE #48: Connection.available_dates fails to deal with tuples
**Describe the bug**
Passing in a list raises a type error, because lists aren't hashable. Passing in a tuple raises an SQL error because the tuple is treated as if it were a string.

**Product**
FlowMachine

**Version**
master

**To Reproduce**
```python
con = flowmachine.connect()
con.available_dates(table=["calls", "sms"])
```

or 

```python
con = flowmachine.connect()
con.available_dates(table=("calls", "sms"))
```

greenape:
Closes #48 

### I have:

### Description

- `Connection.available_dates` now takes a string, or a tuple rather than a string or a list to specify tables to check
- `Connection.available_dates` is now type annotated, to make it clear that _only_ a string or tuple may be used
- Adds a test of checking multiple tables


flowmachine/flowmachine/core/connection.py
```diff
@@ -356,8 +362,8 @@ def available_dates(
         """
 
         if isinstance(table, str) and table.lower() == "all":
-            table = self.subscriber_tables
-        tables = table if isinstance(table, list) else [table]
+            table = tuple(self.subscriber_tables)
+        tables = table if isinstance(table, tuple) else [table]
```
maxalbert: Since we're doing type checking anyway, do we want to add a check that `table` can only be a string or tuple of strings? E.g.:
```python
if not isinstance(table, str) or not isinstance(table, tuple) or not all([isinstance(x, str) for x in table]):
    raise TypeError(f"Argument 'table' must be a string or tuple of strings. Got: {table}")
```

flowmachine/tests/test_class_database.py
```diff
@@ -138,6 +138,14 @@ def test_strict_max_date(flowmachine_connect):
...
 
 
+@pytest.mark.parametrize(
+    "tables", [("calls",), ("calls", "sms"), ("calls", "sms", "mds", "topups")]
+)
+def test_multitable_availability(tables, flowmachine_connect):
+    """Dict returned by available_dates should have keys for all requested tables."""
+    assert tables == tuple(flowmachine_connect.available_dates(table=tables).keys())
+
+
```
maxalbert: Maybe also add a test that we get expected failures for the wrong input types?

PR #53, opened 2018-11-19
Standardise usage of InvalidUsage error

### LINKED ISSUE #37: Reconcile InvalidUsage exceptions in `flowauth/backend/flowauth/admin.py`
Spinning out from a comment on PR #34.

Some error messages in `flowauth/backend/flowauth/admin.py` still have the style "`lifetime_too_long`".

The error messages should be changed to a proper sentence and the exceptions should also include a payload, e.g. `InvalidUsage("Verbose error message", payload={"bad_field":"blah"})`.

greenape:
Closes #37 
### I have:

### Description

This harmonises the way `InvalidUsage` exceptions are raised on the FlowAuth backend. They now _always_ return a long form error message suitable for direct display, and where a specific field is the cause of the issue, give the name of the field with the `bad_field` key in the returned json.


flowauth/backend/flowauth/admin.py
```diff
@@ -540,7 +541,10 @@ def edit_group_servers(group_id):
                     for name, val in right["permissions"].items():
                         setattr(perm, name, val)
                         if val and not getattr(server_capability, name):
-                            raise InvalidUsage("capability_not_allowed_on_server")
+                            raise InvalidUsage(
+                                f"{name} not enabled for this server.",
```
maxalbert: Very minor note: I often find it helpful to be very explicit in the phrasing of the error messages, to give an indication how to fix them. Here I suggest rephrasing this as `f"Permission (or capability) {name} not enabled for this server."

Example: "poll not enabled for this server" leaves me wondering where to look in order to enable it. (Is it something I can change in the web interface? Or is it something that needs to be configured in the server itself when flowkit is installed?)

By contrast, "Permission 'poll' not enabled for this server" makes it more obvious that I need to look in the permissions tab of the web interface. If we like we can even add a hint such as "please check the 'permissions' tab".

Anyway, I'm ok to leave it as it is but I have stumbled across a couple of these things in the past so thought I'd mention it.

PR #58, opened 2018-11-20
Refactor test_query_subscribers

greenape:
This is another pytest-conversion.

test_query_subscribers was actually hiding away tests for _ten_ distinct things. I've split these out into separate files, and converted each one to pytest.

Adds:

- test_home_location
- test_daily_location
- test_dataframe_caching (this one should go away as we remove more code intended for flowmachine-as-library use)
- test_most_frequent_location
- test_last_location
- test_event_table_subset
- test_joined_aggregate
- test_spatial_aggregate
- test_subscriber_location_subset
- test_unique_subscribers

Removes:

test_query_subscribers


flowmachine/tests/test_dataframe_caching.py
```diff
@@ -0,0 +1,71 @@
...
+    sd = EventTableSubset("2016-01-01", "2016-01-02")
+    sd.get_dataframe()
+    sd.turn_off_caching()
+    with pytest.raises(AttributeError):
+        sd._df
+
+
+def test_turn_off_caching_handles_error():
+    """
+    *.turn_off_caching() handles AttributeError.
```
maxalbert: Minor note: there is no hint in the test code that an AttributeError is raised anywhere in this code snippet. Might be worth thinking about what this test actually tries to test from the view of the interface that `EventTableSubset` (and also other classes?) present to the outside and either change the docstring to reflect that, or even rewrite the test to only use public-facing properties. Suggestion for the docstring: "*.turn_off_caching() works without `._df` attribute being present"

flowmachine/tests/test_event_table_subset.py
```diff
@@ -0,0 +1,171 @@
...
+    assert 4 == Range
+    # Also check that all the dates are still there
+    assert 3 in df.day
+    assert 2 in df.day
+    assert 1 in df.day
+
+
+def test_handles_backwards_dates(get_dataframe):
+    """
+    If the subscriber passes dates that are 'backwards' this will be interpreted as spanning midnight.
```
maxalbert: This should be "hours" instead of "dates", right? (Similarly in the test name.)

PR #92, opened 2018-11-27
Convert subscriber subsetting tests to pytest

greenape:
This converts test_subscriber_subsetting (tests for working with subsets of subscribers) to pytest form.

### DIFF
--- a/flowmachine/tests/test_subscriber_subsetting.py
+++ b/flowmachine/tests/test_subscriber_subsetting.py
@@ -172,7 +172,7 @@ def test_can_subset_by_sampler(get_dataframe):
     assert len(su_set) == 10
 
 
-def test_ommitted_subscriber_column(get_dataframe):
+def test_ommitted_subscriber_column(get_dataframe, subscriber_list):
     """Test that a result is returned and warning is raised when ommitting a subscriber column."""
     with pytest.warns(UserWarning):
         su_omit_col = get_dataframe(
--- a/flowmachine/tests/test_subscriber_subsetting.py
+++ b/flowmachine/tests/test_subscriber_subsetting.py
@@ -7,8 +7,6 @@ Tests for the subscriber subsetting functionality
 """
 
 
-from unittest import TestCase
-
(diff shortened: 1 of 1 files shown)


flowmachine/tests/test_subscriber_subsetting.py
```diff
@@ -47,165 +45,151 @@ def test_events_table_subscriber_ident_substitutions(ident):
...
-                daily_location(d, subscriber_subset=Table("subscriber_list"))
-                for d in list_of_dates("2016-01-01", "2016-01-03")
-            ]
-        )
-        rog_df = rog.get_dataframe()
-        hl_df = hl.get_dataframe()
-        sql = "DROP TABLE IF EXISTS subscriber_list"
-        con.execute(sql)
+def test_subset_correct(subscriber_list, get_dataframe):
+    """Test that pushed in subsetting matches .subset result"""
```
maxalbert: What does "pushed in" mean here?

flowmachine/tests/test_subscriber_subsetting.py
```diff
@@ -47,165 +45,151 @@ def test_events_table_subscriber_ident_substitutions(ident):
...
+    su = EventTableSubset(
+        "2016-01-01", "2016-01-03", subscriber_subset=unique_subs_sample
+    )
+    su_set = set(get_dataframe(su).subscriber)
+    uu_set = set(get_dataframe(unique_subs_sample).subscriber)
+    assert su_set == uu_set
+    assert len(su_set) == 10
+
+
+def test_ommitted_subscriber_column(get_dataframe, subscriber_list):
```
maxalbert: Nitpick: `s/ommitted/omitted/` (also in docstring)

flowmachine/tests/test_subscriber_subsetting.py
greenape: Pushed in, in the sense that the subset is shoved down to the initial select from an events table, rather than being done on the home location.

PR #94, opened 2018-11-27
Pytest test spatial activity

greenape:
- Renames test_spatial_activity.py to test_total_location_events.py and converts the content to pytest.

 - Relocates a test for the column names of this class to the same file


flowmachine/tests/test_spatial_activity.py
```diff
@@ -1,152 +0,0 @@
...
-    def test_events_daily(self):
-        """
-        TotalLocationEvents() can get activity on a daily level.
-        """
-        te = TotalLocationEvents(
-            "2016-01-01", "2016-01-04", level="versioned-site", interval="day"
-        )
-        df = te.get_dataframe()
-        self.assertIs(type(df), DataFrame)
-        self.assertEqual(
```
maxalbert: I presume this columns check was removed in the converted version because it already exists elsewhere?

greenape: Covered here: https://github.com/Flowminder/FlowKit/pull/94/files#diff-152b8e433482382027342c580027bd09R16

maxalbert:
@greenape Ping on the missing(?) three tests?

greenape:
Captured by the one parametrised column names test 👍

PR #101, opened 2018-11-27
Pytest subsetting

greenape:
Converts the tests for subsetting and numeric subsetting to pytest form.


flowmachine/tests/test_subsetting.py
```diff
@@ -6,251 +6,241 @@
...
+    flowmachine.SubscriberFeature allows for getting items
+    """
+    rog = RadiusOfGyration("2016-01-01", "2016-01-03")
+    dl = daily_location("2016-01-03")
+    single_subscriber = list(get_dataframe(dl).head(8).subscriber)[3]
+    sub = get_dataframe(rog[single_subscriber])
+    assert set(sub.subscriber) == {single_subscriber}
+
+
+def test_calling_parent_does_break_subsetting(get_dataframe):
```
maxalbert: Hmm, I'm not sure I see how this test checks that subsetting is _broken_. Doesn't it check that it works, or am I misinterpreting the test?

greenape: I do believe there should be a not in there now you mention it!

PR #116, opened 2018-11-28
Convert subscriber location cluster tests to pytest

greenape:
Converts the tests for Hartigan clustering to pytest form.


flowmachine/tests/test_subscriber_location_cluster.py
```diff
@@ -78,203 +77,171 @@ def _geo_augmented_query(self):
...
+    """
+    Test whether a call threshold above 1 limits the number of clusters.
+    """
+    cd = CallDays("2016-01-01", "2016-01-04", level="versioned-site")
+
+    hartigan = HartiganCluster(cd, 50)
+    har_df = hartigan.to_geopandas()
+    assert any(har_df.calldays == 1)
+    har_df = get_dataframe(HartiganCluster(cd, 50, call_threshold=2))
+    assert all(har_df.calldays > 1)
```
jch1g10: Is this correct? It appears to be the opposite of the old assertion.

greenape: So it does.. Looking at the original test tho, it is actually broken - the second assert isn't about the Dataframe produced in the test.

PR #129, opened 2018-11-30
Display error messages helpfully on user details screen

### LINKED ISSUE #127: FlowAuth user details screen error messages
Subcategory of #54.

The FlowAuth user details screen should display an error message when attempting to change the password if any of the details entered are incorrect, or something unexpected happened behind the scenes. Additionally, a success message should be displayed when the password is successfully changed.

jch1g10:
Closes #127

### I have:

### Description

* Display error dialog on user details screen
* Define a snackbar component for displaying messages
* Display a success snackbar on successful password change
* Change behaviour so that user is not logged out after changing password
* Refactor `api.js`
* Turn password change component into a form, so that it will prompt when an input is empty
* Add a new FlowAuth API route to check whether user is signed in
* Check whether user is signed in when `App` component mounts
* Rewrite functions as arrow functions to remove the need to `.bind()` them in the constructor
* Add Cypress tests for user details screen, and command to reset demo data before each test
* Fix a few warnings about deprecated variants of Material UI components

@greenape


flowauth/frontend/src/UserAdminDetails.js
```diff
@@ -117,7 +117,7 @@ class UserAdminDetails extends React.Component {
     return (
       <React.Fragment>
         <Grid xs={12}>
-          <Typography variant="headline" component="h1">
```
greenape: This is because headline etc. are deprecated, right? Do we need to [explicitly turn on the hx variants](https://material-ui.com/style/typography/#migration-to-typography-v2)?

flowauth/frontend/src/App.js
```diff
@@ -1,47 +1,53 @@
...
-  login(is_admin) {
+  setLoggedIn = (is_admin) => {
     this.setState({
       loggedIn: true,
       is_admin: is_admin
     });
   }
   componentDidCatch(error, info) {
     console.log(error);
+    logout();
```
greenape: Why not just call `this.logout()` here? Doing the same thing.

flowauth/backend/flowauth/login.py
```diff
@@ -30,6 +30,18 @@ def signin():
     raise Unauthorized("Incorrect username or password.")
 
 
+@blueprint.route("/is_signed_in")
+def is_signed_in():
```
greenape: Couldn't this be simplified by making it an `@login_required` route (so, you get a 401 if you aren't logged in, and a 200 `{"logged_in":True, "is_admin":bool(current_user.is_admin)}`?

PR #162, opened 2018-12-17
Config for mergify

greenape:
This should turn on automatic merging of PRs, using https://mergify.io

If I've got the config right, this should: 

1. Automerge any PRs from Dependabot that pass CI
1. Automerge PRs with an approving review, that pass all the status checks, and are tagged with a 'ready-to-merge' label
1. Delete the branch after merging

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,0 +1,28 @@
+pull_request_rules:
+  - name: automatic merge for Dependabot pull requests on master
+    conditions:
+      - author=dependabot[bot]
+      - status-success=continuous-integration/travis-ci/pr
+      - status-success=run_build_pipeline
+      - base=master
+    actions:
+      merge:
+        method: rebase
+        strict: true
+  - name: Automatic merge for approved PRs labelled as ready
+    conditions:
+      - base=master
+      - "#approved-reviews-by>=1"
+      - label=ready-to-merge
+      - status-success=codecov/project
+      - status-success=run_build_pipeline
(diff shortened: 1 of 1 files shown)


.mergify.yml
```diff
@@ -0,0 +1,28 @@
+pull_request_rules:
+  - name: automatic merge for Dependabot pull requests on master
+    conditions:
+      - author=dependabot[bot]
+      - status-success=continuous-integration/travis-ci/pr
+      - status-success=run_build_pipeline
+      - base=master
+    actions:
+      merge:
+        method: rebase
```
maxalbert: Are we sure that any PRs can always be merged with a rebase? Do we need a [rebase_fallback](https://doc.mergify.io/actions.html#merge) setting?

maxalbert:
@greenape Do you know what happens if these settings are _almost_ right but fail for some PR in the future because of a corner case? Will mergify simply fail and we can merge the PR manually and tweak the rules accordingly?

greenape:
I think that's the gist of it - we can always override mergify, and tweak as necessary.

.mergify.yml
greenape: Doubtless a good idea..

greenape: Sounds like it defaults to merge actually, so let's leave it as and take action should it break.

maxalbert:
Quick afterthought: I assume mergify will execute the rules in order. Is it worth swapping the rules so that our own PRs take priority over automated ones from dependabot?

PR #175, opened 2018-12-20
Cache metadata

### LINKED ISSUE #125: Cache metadata
At present, when a result is requested via FlowAPI the result is cached into a table within FlowDB, where it can be retrieved multiple times without recalculating it.

FlowDB also stores a small amount of metadata about the caches it holds, at present this is the id of the query, the sql string of the query, the class of the query, the version of FlowMachine that issued the query, the schema and table name of of the cache record, a pickle of the actual query object, and the last time a check was made on whether this record was cached.

I think we should augment this a little, and add at a minimum the original cache date, the number of retrievals, the runtime of _creating_ the cache record and perhaps the _size_ of the cache record.

In addition, I think should look at adding a field that encodes the scoring used for dask's cache eviction (https://github.com/dask/cachey), and is updated on retrievals.

This doesn't have an immediate dividend, because we don't currently implement automated management of the cache, but does form the basis for developing that.

greenape:
Closes #125

### I have:

### Description

This adds additional cache metadata, and some functions for managing and inspecting the cache.

(Might want to break this down into a couple of smaller PRs, but let's see.)

### Changes to FlowDB

Adds a new function, `cache_score`, which takes a query_id and a memory half life and calculates `score = (n_cache_reads+1)*(compute_time / num_bytes * (1 + eps))` (from [cachey](https://github.com/dask/cachey))

#### Changes to cache.cached

- Renames the ts field to last_accessed
- Adds a created field
- Adds an access_count field
- Adds a compute_time field
### Changes to Query, Table, ModelResult

- Records the time taken to compute the result when storing
- Updates access_count, and last_accessed fields when `get_query` is called. Previously this happened when `is_stored` was called, but `get_query` is a better representation of cache reads.
- Removes the ability to store as a view, because explain can't be used and hence we can't calculate a runtime. (Not something that's ever been used to the best of my knowledge)

### Cache Management Functions

- Creates a new `cache` submodule under `core`
- Adds a `get_query_by_id` function, which retrieves a query object from cache by the md5 id (similar to getting at the cache record by creating a `Table` pointing at it).
- Adds `size_of_table`, which gets the total size in bytes (i.e. data & indexes) of a table in the DB
- Adds `size_of_cache`, which gets the total size in bytes of all cache tables, excluding `Table` records
- Adds `compute_time`, which gets the recorded compute time of a cached table given query id
- Adds `score`, which gets the current cache score of a query given a query id and memory bandwidth
- Adds `shrink_one`, which removes the cache record with the lowest `cache_score`
- Adds `shrink_below_size`, which removes cache records until the size of the cache drops below a threshold
- Adds `get_cached_queries_by_score`, which returns cached queries ordered by cache score

In principle, the next step here would be to run `shrink_below_size` in `server.py` to keep the cache manageably sized. Little tricky, because one would need to ensure that cache records didn't get invalidated immediately before they got read, which I guess would be a possibility for fast-to-compute things.


flowmachine/flowmachine/core/cache.py
```diff
@@ -0,0 +1,246 @@
...
+        List of the queries that were removed
+    """
+    initial_cache_size = size_of_cache(connection)
+    removed = []
+    logger.info(
+        f"Shrinking cache from {initial_cache_size} to below {size_threshold}{'(dry run)' if dry_run else ''}."
+    )
+
+    while initial_cache_size > size_threshold:
+        obj_removed, cache_reduction = shrink_one(connection, dry_run)
```
jch1g10: Will this work in dry-run if more than one query would be removed? Since `shrink_one` won't actually remove `obj_removed` in a dry run, it looks like the same query would be picked up again in the next iteration of the loop.

flowmachine/tests/test_cache_utils.py
```diff
@@ -0,0 +1,182 @@
...
+    assert not dl.is_stored
+
+
+def test_shrink_to_size_respects_dry_run(flowmachine_connect):
+    """Test that shrink_below_size doesn't remove anything during a dry run."""
+    dl = daily_location("2016-01-01").store().result()
+    removed_queries = shrink_below_size(
+        flowmachine_connect, size_of_cache(flowmachine_connect) - 1, dry_run=True
+    )
+    assert 1 == len(removed_queries)
```
jch1g10: It would be useful to have this test for a case where `len(removed_queries) > 1` (see earlier comment in `cache.py`)

flowmachine/flowmachine/core/cache.py
greenape: Good point!

flowmachine/flowmachine/core/cache.py
```diff
@@ -0,0 +1,246 @@
...
+    half_life : float
+        Memory decay halflife. Smaller values will decay more slowly.
+
+    Returns
+    -------
+    float
+        Updated cache score
+    """
+    byte_size = size_of_table(connection, *query.table_name.split(".")[::-1])
+    runtime = compute_time(connection, query.md5)
```
josh-gree: runtime = max(some_value_large_enough_for_stuff_to_stick_around, compute_time(connection, query.md5))?

greenape: I don't think that quite resolves it though.. Because the score grows, you'd need `some_value_large_enough_for_stuff_to_stick_around` to change to reflect that, and you end up giving stuff which arguably _shouldn't_ be cached an artificially high score, thereby keeping it around when it should vanish.

flowmachine/tests/test_cache_utils.py
greenape: @jch1g10 this should be sorted now.

greenape:
Closing this for the minute - I’ve not quite got the cache scoring right.

PR #183, opened 2018-12-28
Etl schema

### LINKED ISSUE #182: Book keeping table in etl schema
We need a place to store information about the process of ingestion. This could also become the canonical place to look for available data. I suggest we add a table to the etl schema that allows for this.

josh-gree:
Closes #182

### I have:

### Description

This adds a table to the etl schema. The table will record the process of ingestion for found files.


flowdb/sql/04_schema_other.sql
```diff
@@ -68,3 +68,15 @@ Schema used for temp storage during etl.
 */
 
 CREATE SCHEMA IF NOT EXISTS etl;
+
+CREATE TYPE cdrtype AS ENUM ('voice', 'sms', 'mds');
```
maxalbert: I'm wondering whether it would be better to use a reference table than an enum, because there is a reasonable chance that we may want to support additional data types in the future? (See e.g. https://tapoueh.org/blog/2018/05/postgresql-data-types-enum/)

We can probably also change this later if required, but we'll likely want to move towards using reference tables in other contexts as well, so this might be a good starting point.

Curious what @greenape thinks.

flowdb/sql/04_schema_other.sql
```diff
@@ -68,3 +68,15 @@ Schema used for temp storage during etl.
...
+CREATE TYPE status AS ENUM ('in_process','done', 'quarantine');
```
maxalbert: Nitpick (missing space):
```suggestion
CREATE TYPE status AS ENUM ('in_process', 'done', 'quarantine');
```

flowdb/sql/04_schema_other.sql
```diff
@@ -68,3 +68,15 @@ Schema used for temp storage during etl.
...
+CREATE TABLE etl.etl (
+        id SERIAL NOT NULL,
+        file_name VARCHAR,
+        cdr_type cdrtype,
+	cdr_date DATE,
```
maxalbert: Nitpick (indentation / tab?):
```suggestion
      cdr_date DATE,
```

flowdb/sql/04_schema_other.sql
```diff
@@ -68,3 +68,15 @@ Schema used for temp storage during etl.
...
+        status status,
+        time_stamp TIMESTAMP WITHOUT TIME ZONE,
```
maxalbert: Any specific reason to use timestamp **without** time zone?

flowdb/sql/04_schema_other.sql
josh-gree: I think this is a github bug...As you can see your suggestion is no different...

flowdb/sql/04_schema_other.sql
josh-gree: Irrelevant really when recording etl?

flowdb/sql/04_schema_other.sql
maxalbert: It is... I replaced a tab with a bunch of spaces ;)

josh-gree: Yeah but look at https://github.com/Flowminder/FlowKit/blob/cb126150144d29bdc6181af8ec3e6771e1fb9bae/flowdb/sql/04_schema_other.sql no weird indent?

flowdb/sql/04_schema_other.sql
josh-gree: Yeah perhaps but - not now!?

flowdb/sql/04_schema_other.sql
maxalbert: Sure, so why not use `WITH TIME ZONE` so that timestamps are automatically converted to UTC? Agreed that it's not a big deal, but seems to be good practice to always use timestamps with time zone except in rare cases so we might as well go with the standard?

flowdb/sql/04_schema_other.sql
josh-gree: If this:

```
ALTER TYPE … ADD VALUE (the form that adds a new value to an enum type) cannot be executed inside a transaction block.
```

is the only disadvantage then I think having an enum is simpler...

flowdb/sql/04_schema_other.sql
josh-gree: Yeah ok...

flowdb/sql/04_schema_other.sql
maxalbert: Hmm, it might be. Not sure if there are any issues with tables being locked for a while if an enum is updated, but in any case we can tackle this (if needed) when we re-think the other parts of the schema.

maxalbert:
Yes, I think so

PR #217, opened 2019-01-08
Add cache metadata, and cache management utilities

### LINKED ISSUE #125: Cache metadata

greenape:
Closes #125

### I have:

### Description

This adds additional cache metadata, and some functions for managing and inspecting the cache. Redux of #175 but much more workable.

### Changes to FlowDB

- Adds a new function, `touch_cache`, which takes a query_id updates cache metadata, calculating a cache 'heat' score similar to that from cachey: `score = (n_cache_reads+1)*(compute_time / num_bytes * (1 + half_life))` (from [cachey](https://github.com/dask/cachey))
- Adds a cache_config table, which holds settings for cache half-life, and maximum disk size (set by env vars)
   - These default to 1000, and a tenth of the space available on whatever disk the postgres data store lives on
- Adds cache_half_life and cache_max_size functions as a convenience for getting those values
- Adds a table_size function, which gets the on disk size of a table (including indexes)
- Adds a cache_touches sequence, which tracks the total number of cache retrievals (required for calculating cache score)

#### Changes to cache.cached

- Renames the ts field to last_accessed
- Adds a created field
- Adds an access_count field
- Adds a compute_time field
- Adds a cache_score_multiplier field

#### Changes to Query, Table, ModelResult

- Records the time taken to compute the result when storing
- Updates access_count, last_accessed, and cache_score_multiplier fields when `get_query` is called. Previously this happened when `is_stored` was called, but `get_query` is a better representation of cache reads.
- Removes the ability to store as a view, because explain can't be used and hence we can't calculate a runtime. (Not something that's ever been used to the best of my knowledge)

#### Cache Management Functions

- Creates a new `cache` submodule under `core`
- Adds a `get_query_by_id` function, which retrieves a query object from cache by the md5 id (similar to getting at the cache record by creating a `Table` pointing at it).
- Adds `size_of_table`, which gets the total size in bytes (i.e. data & indexes) of a table in the DB
- Adds `size_of_cache`, which gets the total size in bytes of all cache tables, excluding `Table` records
- Adds `compute_time`, which gets the recorded compute time of a cached table given query id
- Adds `score`, which gets the current cache score of a query given a query id and memory bandwidth
- Adds `shrink_one`, which removes the cache record with the lowest `cache_score`
- Adds `shrink_below_size`, which removes cache records until the size of the cache drops below a threshold
- Adds `get_cached_queries_by_score`, which returns cached queries ordered by cache score
- Adds get/set functions for cache half-life and cache size
- Adds `touch_cache` function, which updates a cache record's score and metadata
- Adds a `reset_cache` function, which wipes the cache clean

Collectively, these changes allow us to check what's being used from cache with much greater ease, and facilitate managing the cache - potentially in a fully automated way.

Because the cache size and half-life are of necessity going to be specific to a given instance of FlowDB, rather than FlowMachine, I've put all the config for them into that. I've also put the score calculation FlowDB-side, because the implementation was a little more elegant without needing to do a bunch of round trips back into python.

The cache_score_multiplier is actually calculated as `cache_score_multiplier + ((1 + ln(2) / half_life) ** total_cache_touches)`, which is equivalent to the implementation here: https://github.com/dask/cachey/blob/master/cachey/score.py and the actual cache score is then the `compute_time_in_seconds/number_of_bytes_used*cache_score_multiplier`.

The total number of cache touches is incremented on any use of `get_query` or a cache write, and obviously an individual cached query's cache multiplier is updated only when _that_ query is read or written.

### DIFF
--- a/docs/source/administrator/cache_management.md
+++ b/docs/source/administrator/cache_management.md
@@ -31,7 +31,19 @@ By default, calling this method will also flush from the cache any cached querie
 
 ### Configuring the Cache
 
-There are two parameters which control FlowKit's cache, both of which are in the `cache.cache_config` table. `half_life` controls how much weight is given to recency of access when updating the cache score. `cache_size` is the maximum size in bytes that the cache tables should occupy on disk. These settings default to 1000, and 10% of available space on the drive where `/var/lib/postgresql/data` is located.
+There are two parameters which control FlowKit's cache, both of which are in the `cache.cache_config` table. `half_life` controls how much weight is given to recency of access when updating the cache score. `half_life` is in units of number of cache retrievals, so a larger value for `half_life` will give less weight to recency and frequency of access. 
+
+!!! example
+    `big_query` and `small_query` both took 100 seconds to calculate. `big_query` takes 100 bytes to store, and `small_query` takes 10 bytes.
+    Their _costs_ are `compute_time/storage_size`, or 1 for `big_query` and 10 for `small_query`. `small_query` is stored first and has an initial cache score of 10.
+    If query `big_query` is stored next, with a `half_life` of 2, it will get an initial cache score of 1.35. 
+    
+    Just in terms of the balance between compute time and storage cost, `small_query` is more valuable in cache because it is relatively cheaper to store.  However, after only four retrievals of `big_query` from cache, `big_query` will have a cache score of 13.3, meaning it is more valuable in cache because it is so frequently used.
+    
+    If `half_life` was instead set to 10, `big_query` would need to be retrieved _seven_ times to exceed the cache score of `small_query`. 
+     
+ 
+`cache_size` is the maximum size in bytes that the cache tables should occupy on disk. These settings default to 1000, and 10% of available space on the drive where `/var/lib/postgresql/data` is located.
 
 These values can be overridden when creating a new FlowDB container by setting the `CACHE_SIZE` and `CACHE_HALF_LIFE` environment variables for the container, set by updating the `cache.cache_config` table after connecting directly to FlowDB, or modified using the cache submodule.
 
--- a/docs/source/administrator/cache_management.md
+++ b/docs/source/administrator/cache_management.md
@@ -1,0 +1,40 @@
+# Caching in FlowKit
+
+FlowKit implements a caching system to enhance performance. Queries requested via FlowAPI are cached in FlowDB, under the `cache` schema.
+
+Once cached, a query will not be recalulated - the cached version will simply be returned instead, which can save significant computation time. In addition to queries which are _directly_ returned, FlowKit may cache queries which are used in calculating other queries. For example, calculating a modal location aggregate, and a daily location aggregate will both use the same underlying query when the dates (and other parameters) overlap. Hence, caching the underlying query allows both the aggregate and the modal location aggregate to be produced faster.
+
+
+This performance boost is achieved at the cost of disk space usage, and management of the cached data will sometimes be required.
+
+## Cache Management
+
+FlowMachine and FlowDB provide tools to inspect and manage the content of FlowKit's cache. FlowDB also contains metadata about the content of cache, in the `cache.cached` table.
+
+Administrators can inspect this table directly by connecting to FlowDB, but in many scenarios the better option is to make use of FlowMachine's [cache management module](../../developer/api/flowmachine/core/cache/).
+
+The cache submodule provides functions to assess the disk usage of the cache tables, and to reduce the disk usage below a desired threshold.
+
+### Shrinking Cache
+
+To identify which tables should be discarded from cache, FlowKit keeps track of how expensive they were to calculate initially, how much disk space they occupy, and how often and recently they have been used. These factors are combined into a cache score, based on the [cachey](https://github.com/dask/cachey) algorithm.
+
+Each cache table has a cache score, with a higher score indicating that the table has more cache value.
+
+FlowMachine provides two functions which make use of this cache score to reduce the size of the cache - [`shrink_below_size`](../../developer/api/flowmachine/core/cache/#shrink_below_size), and [`shrink_one`](../../developer/api/flowmachine/core/cache/#shrink_one). `shrink_one` flushes the table with the _lowest_ cache score. `shrink_below_size` flushes tables until the disk space used by the cache falls below a threshold[^1] by calling `shrink_one` repeatedly.
+
+### Removing a Specific Query from Cache
+
+If a specific query must be removed from the cache, then an administrator can retrieve the query's _object_ from FlowDB by using the [`get_query_by_id`](../../developer/api/flowmachine/core/cache/#get_query_by_id) function of the `cache` submodule, and calling the object's [`invalidate_db_cache`](../../developer/api/flowmachine/core/query/#invalidate_db_cache) method.
+
+By default, calling this method will also flush from the cache any cached queries which used that query in their calculation. This will also cascade to any queries which used _those_ queries, and so on. Setting the `cascade` argument to `False` will remove only the one query.
+
+### Configuring the Cache
+
+There are two parameters which control FlowKit's cache, both of which are in the `cache.cache_config` table. `half_life` controls how much weight is given to recency of access when updating the cache score. `cache_size` is the maximum size in bytes that the cache tables should occupy on disk. These settings default to 1000, and 10% of available space on the drive where `/var/lib/postgresql/data` is located.
+
+These values can be overridden when creating a new FlowDB container by setting the `CACHE_SIZE` and `CACHE_HALF_LIFE` environment variables for the container, set by updating the `cache.cache_config` table after connecting directly to FlowDB, or modified using the cache submodule.
+
+    
+
+[^1]:By default, this uses the value set for `cache_size` in `cache.cache_config`.
--- a/docs/source/administrator/cache_management.md
+++ b/docs/source/administrator/cache_management.md
@@ -38,14 +38,14 @@ There are two parameters which control FlowKit's cache, both of which are in the
 !!! example
     `big_query` and `small_query` both took 100 seconds to calculate. `big_query` takes 100 bytes to store, and `small_query` takes 10 bytes.
     Their _costs_ are `compute_time/storage_size`, or 1 for `big_query` and 10 for `small_query`. `small_query` is stored first and has an initial cache score of 10.
-    If query `big_query` is stored next, with a `half_life` of 2, it will get an initial cache score of 1.35. 
+    If query `big_query` is stored next, with a `half_life` of 2.0, it will get an initial cache score of 1.35. 
     
     Just in terms of the balance between compute time and storage cost, `small_query` is more valuable in cache because it is relatively cheaper to store.  However, after only four retrievals of `big_query` from cache, `big_query` will have a cache score of 13.3, meaning it is more valuable in cache because it is so frequently used.
     
-    If `half_life` was instead set to 10, `big_query` would need to be retrieved _seven_ times to exceed the cache score of `small_query`. 
+    If `half_life` was instead set to 10.0, `big_query` would need to be retrieved _seven_ times to exceed the cache score of `small_query`. 
      
  
-`cache_size` is the maximum size in bytes that the cache tables should occupy on disk. These settings default to 1000, and 10% of available space on the drive where `/var/lib/postgresql/data` is located.
+`cache_size` is the maximum size in bytes that the cache tables should occupy on disk. These settings default to 1000.0, and 10% of available space on the drive where `/var/lib/postgresql/data` is located.
 
 These values can be overridden when creating a new FlowDB container by setting the `CACHE_SIZE` and `CACHE_HALF_LIFE` environment variables for the container, set by updating the `cache.cache_config` table after connecting directly to FlowDB, or modified using the cache submodule.
 
--- a/docs/source/administrator/cache_management.md
+++ b/docs/source/administrator/cache_management.md
@@ -23,11 +23,13 @@ Each cache table has a cache score, with a higher score indicating that the tabl
 
 FlowMachine provides two functions which make use of this cache score to reduce the size of the cache - [`shrink_below_size`](../../developer/api/flowmachine/core/cache/#shrink_below_size), and [`shrink_one`](../../developer/api/flowmachine/core/cache/#shrink_one). `shrink_one` flushes the table with the _lowest_ cache score. `shrink_below_size` flushes tables until the disk space used by the cache falls below a threshold[^1] by calling `shrink_one` repeatedly.
 
+If necessary, the cache can also be completely reset using the [`reset_cache`](../../developer/api/flowmachine/core/cache/#reset_cache) function.
+
 ### Removing a Specific Query from Cache
 
-If a specific query must be removed from the cache, then an administrator can retrieve the query's _object_ from FlowDB by using the [`get_query_by_id`](../../developer/api/flowmachine/core/cache/#get_query_by_id) function of the `cache` submodule, and calling the object's [`invalidate_db_cache`](../../developer/api/flowmachine/core/query/#invalidate_db_cache) method.
+If a specific query must be removed from the cache, then an administrator can use the [`invalidate_cache_by_id`](../../developer/api/flowmachine/core/cache/#invalidate_cache_by_id) function of the `cache` submodule.
 
-By default, calling this method will also flush from the cache any cached queries which used that query in their calculation. This will also cascade to any queries which used _those_ queries, and so on. Setting the `cascade` argument to `False` will remove only the one query.
+By default, this function only removes that specific query from cache. However, setting the `cascade` argument to `True` will also flush from the cache any cached queries which used that query in their calculation. This will also cascade to any queries which used _those_ queries, and so on.
 
 ### Configuring the Cache
 
--- a/flowdb/sql/07_schema_cache.sql
+++ b/flowdb/sql/07_schema_cache.sql
@@ -46,5 +46,5 @@ CREATE TABLE IF NOT EXISTS cache.dependencies
                             );
 
 CREATE TABLE cache.cache_config (key text, value text);
-INSERT INTO cache.cache_config (key, value) VALUES ('half_life', 1000);
-INSERT INTO cache.cache_config (key, value) VALUES ('cache_size', 4950307635);
+INSERT INTO cache.cache_config (key, value) VALUES ('half_life', 0);
+INSERT INTO cache.cache_config (key, value) VALUES ('cache_size', 0);
--- a/flowdb/sql/07_schema_cache.sql
+++ b/flowdb/sql/07_schema_cache.sql
@@ -46,5 +46,5 @@ CREATE TABLE IF NOT EXISTS cache.dependencies
                             );
 
 CREATE TABLE cache.cache_config (key text, value text);
-INSERT INTO cache.cache_config (key, value) VALUES ('half_life', 0);
-INSERT INTO cache.cache_config (key, value) VALUES ('cache_size', 0);
+INSERT INTO cache.cache_config (key, value) VALUES ('half_life', NULL);
+INSERT INTO cache.cache_config (key, value) VALUES ('cache_size', NULL);
--- a/flowdb/sql/07_schema_cache.sql
+++ b/flowdb/sql/07_schema_cache.sql
@@ -17,13 +17,19 @@ CREATE TABLE IF NOT EXISTS cache.cached
                                 query_id CHARACTER(32) NOT NULL,
                                 version CHARACTER VARYING,
                                 query TEXT,
-                                ts TIMESTAMP WITH TIME ZONE,
+                                created TIMESTAMP WITH TIME ZONE,
+                                access_count INTEGER,
+                                last_accessed TIMESTAMP WITH TIME ZONE,
+                                compute_time NUMERIC,
+                                cache_score_multiplier NUMERIC,
                                 class CHARACTER VARYING,
                                 schema CHARACTER VARYING,
                                 tablename CHARACTER VARYING,
                                 obj BYTEA,
                                 CONSTRAINT cache_pkey PRIMARY KEY (query_id)
                             );
+/* Sequence counting total number of retrievals from cache */
+CREATE SEQUENCE cache.cache_touches START 1;
 CREATE TABLE IF NOT EXISTS cache.dependencies
                             (
                                 query_id CHARACTER(32) NOT NULL,
@@ -38,3 +44,7 @@ CREATE TABLE IF NOT EXISTS cache.dependencies
                                     ON UPDATE NO ACTION
                                     ON DELETE CASCADE
                             );
+
+CREATE TABLE cache.cache_config (key text, value text);
+INSERT INTO cache.cache_config (key, value) VALUES ('half_life', 1000);
+INSERT INTO cache.cache_config (key, value) VALUES ('cache_size', 4950307635);
--- a/flowmachine/flowmachine/core/cache.py
+++ b/flowmachine/flowmachine/core/cache.py
@@ -110,7 +110,9 @@ def get_query_object_by_id(connection: "Connection", query_id: str) -> "Query":
         raise ValueError(f"Query id '{query_id}' is not in cache on this connection.")
 
 
-def get_cached_query_objects_ordered_by_score(connection: "Connection") -> List[Tuple["Query", int]]:
+def get_cached_query_objects_ordered_by_score(
+    connection: "Connection"
+) -> List[Tuple["Query", int]]:
     """
     Get all cached query objects in ascending cache score order.
 
--- a/flowmachine/flowmachine/core/cache.py
+++ b/flowmachine/flowmachine/core/cache.py
@@ -169,6 +169,9 @@ def shrink_below_size(
     else:
         shrink = shrink_one
 
+    if size_threshold is None:
+        size_threshold = get_max_size_of_cache(connection)
+
     while initial_cache_size > size_threshold:
         obj_removed, cache_reduction = shrink(connection)
         removed.append(obj_removed)
--- a/flowmachine/flowmachine/core/cache.py
+++ b/flowmachine/flowmachine/core/cache.py
@@ -51,6 +51,11 @@ def reset_cache(connection: "Connection") -> None:
     ----------
     connection : Connection
     """
+    # For deletion purposes, we ignore Table objects. These either point to something
+    # outside the cache schema and hence shouldn't be removed by calling this, or
+    # they point to a cache table and hence are a duplicate of a Query entry which
+    # will also point to that table.
+
     qry = f"SELECT tablename FROM cache.cached WHERE NOT class='Table'"
     tables = connection.fetch(qry)
     with connection.engine.begin() as trans:
--- a/flowmachine/flowmachine/core/cache.py
+++ b/flowmachine/flowmachine/core/cache.py
@@ -184,7 +184,7 @@ def shrink_below_size(
     list of "Query"
         List of the queries that were removed
     """
-    initial_cache_size = size_of_cache(connection)
+    initial_cache_size = get_size_of_cache(connection)
     removed = []
     logger.info(
         f"Shrinking cache from {initial_cache_size} to below {size_threshold}{' (dry run)' if dry_run else ''}."
@@ -209,7 +209,9 @@ def shrink_below_size(
     return removed
 
 
-def size_of_table(connection: "Connection", table_name: str, table_schema: str) -> int:
+def get_size_of_table(
+    connection: "Connection", table_name: str, table_schema: str
+) -> int:
     """
     Get the size on disk in bytes of a table in the database.
 
@@ -236,7 +238,7 @@ def size_of_table(connection: "Connection", table_name: str, table_schema: str) 
         )
 
 
-def size_of_cache(connection: "Connection") -> int:
+def get_size_of_cache(connection: "Connection") -> int:
     """
     Get the total size in bytes of all cache tables.
 
@@ -331,7 +333,7 @@ def set_max_size_of_cache(connection: "Connection", cache_size_limit: int) -> No
         trans.execute(sql)
 
 
-def compute_time(connection: "Connection", query_id: str) -> float:
+def get_compute_time(connection: "Connection", query_id: str) -> float:
     """
     Get the time in seconds that a cached query took to compute.
 
@@ -358,7 +360,7 @@ def compute_time(connection: "Connection", query_id: str) -> float:
         raise ValueError(f"Query id '{query_id}' is not in cache on this connection.")
 
 
-def score(connection: "Connection", query_id: str) -> float:
+def get_score(connection: "Connection", query_id: str) -> float:
     """
     Get the current cache score for a cached query.
 
--- a/flowmachine/flowmachine/core/cache.py
+++ b/flowmachine/flowmachine/core/cache.py
@@ -86,7 +86,7 @@ def invalidate_cache_by_id(
     return query_obj
 
 
-def get_query_by_id(connection: "Connection", query_id: str) -> "Query":
+def get_query_object_by_id(connection: "Connection", query_id: str) -> "Query":
     """
     Get a query object from cache by id.
 
--- a/flowmachine/flowmachine/core/cache.py
+++ b/flowmachine/flowmachine/core/cache.py
@@ -162,7 +162,7 @@ def shrink_one(connection: "Connection", dry_run: bool = False) -> "Query":
         f"{'Would' if dry_run else 'Will'} remove cache record for {obj_to_remove.md5} of type {obj_to_remove.__class__}"
     )
     logger.info(
-        f"Table {obj_to_remove.table_name} ({obj_size} bytes) {'would' if dry_run else 'will'} be removed."
+        f"Table {obj_to_remove.fully_qualified_table_name} ({obj_size} bytes) {'would' if dry_run else 'will'} be removed."
     )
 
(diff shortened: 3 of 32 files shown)


docs/source/administrator/cache_management.md
```diff
@@ -0,0 +1,40 @@
...
+
+Each cache table has a cache score, with a higher score indicating that the table has more cache value.
+
+FlowMachine provides two functions which make use of this cache score to reduce the size of the cache - [`shrink_below_size`](../../developer/api/flowmachine/core/cache/#shrink_below_size), and [`shrink_one`](../../developer/api/flowmachine/core/cache/#shrink_one). `shrink_one` flushes the table with the _lowest_ cache score. `shrink_below_size` flushes tables until the disk space used by the cache falls below a threshold[^1] by calling `shrink_one` repeatedly.
+
+### Removing a Specific Query from Cache
+
+If a specific query must be removed from the cache, then an administrator can retrieve the query's _object_ from FlowDB by using the [`get_query_by_id`](../../developer/api/flowmachine/core/cache/#get_query_by_id) function of the `cache` submodule, and calling the object's [`invalidate_db_cache`](../../developer/api/flowmachine/core/query/#invalidate_db_cache) method.
```
maxalbert: This seems slightly more convoluted than necessary. Worth adding a helper function to make it slightly more convenient for administrators wanting to do this? Happy to do this in a separate PR.

docs/source/administrator/cache_management.md
```diff
@@ -0,0 +1,40 @@
...
+
+By default, calling this method will also flush from the cache any cached queries which used that query in their calculation. This will also cascade to any queries which used _those_ queries, and so on. Setting the `cascade` argument to `False` will remove only the one query.
+
+### Configuring the Cache
+
+There are two parameters which control FlowKit's cache, both of which are in the `cache.cache_config` table. `half_life` controls how much weight is given to recency of access when updating the cache score. `cache_size` is the maximum size in bytes that the cache tables should occupy on disk. These settings default to 1000, and 10% of available space on the drive where `/var/lib/postgresql/data` is located.
```
maxalbert: What's the unit of `half_life`? Seconds?

flowdb/sql/07_schema_cache.sql
```diff
@@ -37,4 +43,8 @@ CREATE TABLE IF NOT EXISTS cache.dependencies
                                     REFERENCES cache.cached (query_id) MATCH SIMPLE
                                     ON UPDATE NO ACTION
                                     ON DELETE CASCADE
-                            );
\ No newline at end of file
+                            );
+
+CREATE TABLE cache.cache_config (key text, value text);
```
maxalbert: Shouldn't the `value` be numeric?

flowdb/sql/07_schema_cache.sql
```diff
@@ -37,4 +43,8 @@ CREATE TABLE IF NOT EXISTS cache.dependencies
...
+INSERT INTO cache.cache_config (key, value) VALUES ('half_life', 0);
```
maxalbert: Worth initialising this with NULL just to be explicit that it'll be changed later?

flowdb/sql/07_schema_cache.sql
greenape: Well there are already two types in there - half_life is a float, cache_size is an integer number of bytes. But yes, also allows for other types if more settings should be added.

maxalbert: Oh sorry, I didn't realise that `half_life` is a float because in the docs (and when it is initialised) this is done with integer values.

docs/source/administrator/cache_management.md
maxalbert: Not very important, but maybe worth replacing `1000` with `1000.0` to hint at the fact that `half_life` is a float?

flowmachine/flowmachine/core/cache.py
```diff
@@ -0,0 +1,384 @@
...
+def reset_cache(connection: "Connection") -> None:
+    """
+    Reset the query cache. Deletes any tables under cache schema, resets the cache count
+    and clears the cached and dependencies tables.
+
+    Parameters
+    ----------
+    connection : Connection
+    """
+    qry = f"SELECT tablename FROM cache.cached WHERE NOT class='Table'"
```
maxalbert: I _think_ the logic here is correct, but it's a bit confusing that the docstring says it deletes any tables under the cache schema and then you select all items where `class` is _not_ `'Table'`. Is there any way of clarifying things a bit more? (Not sure what to suggest because I'm a bit hazy on exactly what lives in the cache schema.

flowmachine/flowmachine/core/cache.py
```diff
@@ -0,0 +1,384 @@
...
+    """
+    qry = f"SELECT obj FROM cache.cached WHERE query_id='{query_id}'"
+    try:
+        obj = connection.fetch(qry)[0][0]
+        return pickle.loads(obj)
+    except IndexError:
+        raise ValueError(f"Query id '{query_id}' is not in cache on this connection.")
+
+
+def get_cached_queries_by_score(connection: "Connection") -> List[Tuple["Query", int]]:
```
maxalbert: See above: rename this?
```suggestion
def get_cached_query_objects_by_score(connection: "Connection") -> List[Tuple["Query", int]]:
```

flowmachine/flowmachine/core/cache.py
```diff
@@ -0,0 +1,384 @@
...
+    """
+    Get all cached queries in ascending cache score order.
```
maxalbert:
```suggestion
    Get all cached query objects in ascending cache score order.
```

flowmachine/flowmachine/core/cache.py
```diff
@@ -0,0 +1,384 @@
...
+
+    Returns
+    -------
+    list of "Query"
+        List of the queries that were removed
+    """
+    initial_cache_size = size_of_cache(connection)
+    removed = []
+    logger.info(
+        f"Shrinking cache from {initial_cache_size} to below {size_threshold}{'(dry run)' if dry_run else ''}."
```
maxalbert: Missing space before `(dry run)`?
```suggestion
        f"Shrinking cache from {initial_cache_size} to below {size_threshold}{' (dry run)' if dry_run else ''}."
```

flowmachine/flowmachine/core/cache.py
```diff
@@ -0,0 +1,384 @@
...
+        cached_queries = iter(get_cached_queries_by_score(connection))
+        shrink = lambda x: cached_queries.__next__()
+    else:
+        shrink = shrink_one
+
+    if size_threshold is None:
+        size_threshold = get_max_size_of_cache(connection)
+
+    while initial_cache_size > size_threshold:
+        obj_removed, cache_reduction = shrink(connection)
```
maxalbert: Should this print debugging info about the removed object, too?

flowmachine/flowmachine/core/cache.py
```diff
@@ -0,0 +1,384 @@
...
+
+    Returns
+    -------
+    int
+        Number of bytes in total used by cache tables
+
+    """
+    sql = """SELECT sum(table_size(tablename, schema)) as total_bytes 
+        FROM cache.cached  
+        WHERE NOT cached.class='Table'"""
```
maxalbert: It occurs to me that almost all of the queries have an explicit exclusion clause of the form `WHERE NOT cached.class='Table'` (or similar). Maybe not for this PR, but is it worth storing `Table` objects somewhere else so that we don't need to continually treat them as an exception?

flowmachine/flowmachine/core/cache.py
```diff
@@ -0,0 +1,384 @@
...
+    Get the upper limit set in FlowDB for the cache size, in bytes.
+
+    Parameters
+    ----------
+    connection : "Connection"
+
+    Returns
+    -------
+    int
+        Number of bytes in total available cache tables
```
maxalbert:
```suggestion
        Number of bytes in total available to cache tables
```

flowmachine/flowmachine/core/cache.py
```diff
@@ -0,0 +1,384 @@
...
+    ----------
+    connection : "Connection"
+    cache_half_life : float
+        Setting for half-life
+
+    Notes
+    -----
+
+    Changing this setting without flushing the cache first may have unpredictable consequences
+    and should be avoided.
```
maxalbert: Depending how severe the potential problems are, maybe we should add an additional required parameter `yes_i_know_what_i_am_doing=True` to protect users from accidental use? ;)

flowmachine/flowmachine/core/cache.py
```diff
@@ -0,0 +1,384 @@
...
+    """
+    sql = (
+        f"UPDATE cache.cache_config SET value='{cache_half_life}' WHERE key='half_life'"
+    )
+    with connection.engine.begin() as trans:
+        trans.execute(sql)
```
maxalbert: What happens (here or in other functions) if the user passes a wrong parameter (e.g. something that can't be cast to a float)? Do we need to raise a meaningful error or are we ok for the `sqlalchemy`/`psycopg2` error to bubble up?

flowmachine/flowmachine/core/cache.py
```diff
@@ -0,0 +1,384 @@
...
+
+    """
+    sql = f"UPDATE cache.cache_config SET value='{cache_size_limit}' WHERE key='cache_size'"
+    with connection.engine.begin() as trans:
+        trans.execute(sql)
+
+
+def compute_time(connection: "Connection", query_id: str) -> float:
+    """
+    Get the time in ms that a cached query took to compute.
```
maxalbert:
```suggestion
    Get the time in seconds that a cached query took to compute.
```

flowmachine/flowmachine/core/cache.py
```diff
@@ -0,0 +1,384 @@
...
+    Returns
+    -------
+    float
+        Current cache score of this query
+
+    """
+    try:
+        return float(
+            connection.fetch(
+                f"""SELECT cache_score_multiplier*((compute_time/1000)/table_size(tablename, schema))
```
maxalbert: This formula is repeated in a couple of places. Worth wrapping it up in a helper function with a meaningful name?

flowmachine/flowmachine/core/query.py
```diff
@@ -188,6 +193,10 @@ def get_query(self):
             schema, name = table_name.split(".")
             with rlock(self.redis, self.md5):
                 if self.connection.has_table(schema=schema, name=name):
+                    try:
+                        touch_cache(self.connection, self.md5)
+                    except ValueError:
+                        pass  # Cache record not written yet
```
maxalbert: It took me a minute to understand why this exception is ok to ignore. Without re-reading the definition of `touch_cache()` I assumed it would create a cache record if it didn't exist yet (similar to how the Unix command `touch` creates an empty file if it doesn't exist). Is it worth actually making this an optional (or required?) argument, e.g. `touch_cache(self.connection, self.md5, ignore_if_not_exists=True)`

Regarding the context of this addition, why does `get_query()` want to touch the cache at all? Is it because something else is trying to make use of this query? Does it make a difference whether `get_query()` is called by some other flowmachine-internal class or by the user?

flowmachine/flowmachine/core/query.py
```diff
@@ -429,12 +437,6 @@ def _make_sql(self, name=None, schema=None, as_view=False, force=False):
             Ordered list of SQL strings to execute.
         """
 
-        table_type = ""
-        if as_view:
```
maxalbert: Thanks for removing the `as_view` option! Definitely simplifies the code. 👍

flowmachine/flowmachine/core/query.py
```diff
@@ -498,63 +491,43 @@ def to_sql_async(
...
                 with con.begin():
-                    sql = """CREATE SCHEMA IF NOT EXISTS {}""".format(schema)
-                    con.execute(sql)
-                    try:
-                        for Q in Qs:
-                            con.execute(Q)
-                    except Exception as e:
-                        logger.error(f"Error executing SQL: '{Q}'. Error was {e}")
-                        raise e
+                    rs = []
```
maxalbert: I'll be honest that this loop does read a little cryptic at first. Maybe it's due to the short names `Qs`, `rs` and the fact that it's not immediately clear what types of objects we're dealing with (e.g. what type of object is `r` and why is `r[0][0][0]` a query plan(?) and how do we know that it has a key called `"Execution Time"`?

Not sure I have any good suggestions other than maybe slightly more explicit names?

Also, why is `Qs` an iterable? Shouldn't `self._make_sql(...)` just return the SQL string?

flowmachine/flowmachine/core/model.py
```diff
@@ -72,7 +73,9 @@ def new_f(self, *args, **kwargs):
         mr = ModelResult(self, run_args=args, run_kwargs=kwargs)
         if mr.is_stored:
             return mr
+        start_time = time.perf_counter()  # TODO: Should use process time here really
```
maxalbert: If you think `time.process_time()` is more appropriate, why not switch it straight away?

In general, is there any use in us recording _both_ wall time and process time? Or is this not possible with timings that are recorded by Postgres?

flowmachine/flowmachine/core/query.py
```diff
@@ -777,6 +752,14 @@ def invalidate_db_cache(self, name=None, schema=None, cascade=True, drop=True):
         """
         with rlock(self.redis, self.md5):
             con = self.connection.engine
+            try:
+                table_form = self.get_table()
```
maxalbert: What does table _form_ mean here?

flowmachine/flowmachine/core/query.py
greenape: Added a comment to explain the pass.

The equivalent to this used to happen in `is_stored`, but I felt like `get_query` is actually a better indicator of the cache record being _used_, since in most cases the only time `get_query` is called is if you're running the query to get the dataframe, or another query is including its results.

Ideally, we'd do this update with a trigger on the actual cache table being read via a SELECTm but unfortunately Postgres doesn't support that at this time.

flowmachine/flowmachine/core/query.py
```diff
@@ -777,6 +752,14 @@ def invalidate_db_cache(self, name=None, schema=None, cascade=True, drop=True):
...
+                if table_form is not self:
```
maxalbert: And what is a possible scenario where `table_form` is _not_ `self`?

flowmachine/flowmachine/core/query.py
greenape: Given these slightly less terse names.

Qs (as was) is a list of ddl ops. Marginally nicer to have this as an iterable because you can report the specific one that failed.

r[0][0][0] should be a (json) query plan which Postgres helpfully returns as a row, with one field, and wrapped in a list.

flowmachine/flowmachine/core/query.py
greenape: This query, in the form of a Table. Happy to change to something else if you've got a clearer way to express that.

flowmachine/tests/test_cache.py
```diff
@@ -12,6 +12,25 @@
...
 
 
+def test_table_records_removed(flowmachine_connect):
+    """Test that removing a query from cache removes any Tables in cache that pointed to it."""
+    dl = daily_location("2016-01-01")
+    dl.store().result()
+    assert dl.is_stored
+    table = dl.get_table()
+    assert bool(
+        flowmachine_connect.fetch(
```
maxalbert: Might be worth having a helper function `has_cache_entry(query_id)`?

flowmachine/flowmachine/core/query.py
greenape:
```python
dl = daily_location("2016-01-01").store().result()
dl_table = dl.get_table()
```

`dl` is a query object, which happens to be cached and hence exists as a table, `dl_table` is a `Table` object which points _to_ that cached table.

flowmachine/flowmachine/core/query.py
```diff
@@ -676,19 +643,27 @@ def _db_store_cache_metadata(self):
                     "SELECT * FROM cache.cached WHERE query_id='{}'".format(self.md5)
                 )
             )
+
             with con.begin():
+                cache_record_insert = """
+                INSERT INTO cache.cached 
+                (query_id, version, query, created, access_count, last_accessed, compute_time,
```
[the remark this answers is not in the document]
greenape: I was going to say that you never compute it again, but that's not necessarily true.. I think we can leave that for a followup tho, as it would require some rethinking of how queries are identified as cached and what the cache.cached table does.

flowmachine/tests/test_cache_utils.py
```diff
@@ -0,0 +1,350 @@
...
+    set_max_size_of_cache,
+    get_cache_half_life,
+    set_cache_half_life,
+    invalidate_cache_by_id,
+)
+from flowmachine.features import daily_location
+
+
+def test_scoring(flowmachine_connect):
+    """Test that score updating algorithm is correct by comparing to cachey as reference implementation"""
```
maxalbert: Purely personal aesthetics-related comment, but I always find it more readable if the triple-quotes are on separate lines:
```suggestion
    """
    Test that score updating algorithm is correct by comparing to cachey as reference implementation
    """
```

I'll add this suggestion to the other test functions as well, but feel free to ignore if you disagree or if there is a style guide that recommends otherwise.

flowmachine/tests/test_cache_utils.py
```diff
@@ -0,0 +1,350 @@
...
+    # Add another unrelated cache record, which should have a higher initial score
+    dl_2 = daily_location("2016-01-02").store().result()
+    dl_time = compute_time(flowmachine_connect, dl_2.md5)
+    dl_size = size_of_table(flowmachine_connect, *dl_2.table_name.split(".")[::-1])
+    cache_score = cachey_scorer.touch("dl_2", dl_time / dl_size)
+    assert cache_score == pytest.approx(score(flowmachine_connect, dl_2.md5))
+
+
+def test_touch_cache_record_for_query(flowmachine_connect):
+    """Touching a cache record for a query should update access count, last accessed, & counter."""
```
maxalbert:
```suggestion
    """
    Touching a cache record for a query should update access count, last accessed, & counter.
    """
```

flowmachine/tests/test_cache_utils.py
```diff
@@ -0,0 +1,350 @@
...
+    assert (
+        accessed_at
+        < flowmachine_connect.fetch(
+            f"SELECT last_accessed FROM cache.cached WHERE query_id='{table.md5}'"
+        )[0][0]
+    )
+
+
+def test_touch_cache_record_for_table(flowmachine_connect):
+    """Touching a cache record for a table should update access count and last accessed but not touch score, or counter."""
```
maxalbert:
```suggestion
    """
    Touching a cache record for a table should update access count and last accessed but not touch score, or counter.
    """
```

flowmachine/tests/test_cache_utils.py
```diff
@@ -0,0 +1,350 @@
...
+
+def test_compute_time():
+    """Compute time should take value returned in ms and turn it into seconds."""
+    connection_mock = Mock()
+    connection_mock.fetch.return_value = [[10]]
+    assert 10 / 1000 == compute_time(connection_mock, "DUMMY_ID")
+
+
+def test_get_cached_queries_by_score(flowmachine_connect):
+    """Test that all records which are queries are returned in correct order."""
```
maxalbert:
```suggestion
    """
    Test that all records which are queries are returned in correct order.
    """
```

flowmachine/tests/test_cache_utils.py
```diff
@@ -0,0 +1,350 @@
...
+    # Should prefer the larger, but slower to calculate and more used dl over the aggregation
+    cached_queries = get_cached_queries_by_score(flowmachine_connect)
+    assert 2 == len(cached_queries)
+    assert dl_agg.md5 == cached_queries[0][0].md5
+    assert dl.md5 == cached_queries[1][0].md5
+    assert 2 == len(cached_queries[0])
+
+
+def test_shrink_one(flowmachine_connect):
+    """Test that shrink_one removes a cache record."""
```
maxalbert:
```suggestion
    """
    Test that shrink_one removes a cache record.
    """
```

flowmachine/tests/test_cache_utils.py
```diff
@@ -0,0 +1,350 @@
...
+        f"UPDATE cache.cached SET cache_score_multiplier = 0.5 WHERE query_id='{dl.md5}'"
+    )
+    removed_query, table_size = shrink_one(flowmachine_connect)
+    assert dl.md5 == removed_query.md5
+    assert not dl.is_stored
+    assert dl_aggregate.is_stored
+
+
+def test_shrink_to_size_does_nothing_when_cache_ok(flowmachine_connect):
+    """Test that shrink_below_size doesn't remove anything if cache size is within limit."""
```
maxalbert:
```suggestion
    """
    Test that shrink_below_size doesn't remove anything if cache size is within limit.
    """
```

flowmachine/tests/test_cache_utils.py
maxalbert: Actually, I stopped halfway through because the Github interface is a little clumsy for this. 😂

Let me know if you're ok with this change and I'll reformat the remaining docstrings in a single commit.

flowmachine/tests/test_cache_utils.py
```diff
@@ -0,0 +1,350 @@
...
+def test_scoring(flowmachine_connect):
+    """Test that score updating algorithm is correct by comparing to cachey as reference implementation"""
+    dl = daily_location("2016-01-01").store().result()
+    dl_time = compute_time(flowmachine_connect, dl.md5)
+    dl_size = size_of_table(flowmachine_connect, *dl.table_name.split(".")[::-1])
+    initial_score = score(flowmachine_connect, dl.md5)
+    cachey_scorer = Scorer(1000)
+    cache_score = cachey_scorer.touch("dl", dl_time / dl_size)
+    assert cache_score == pytest.approx(initial_score)
+    # Touch again
```
maxalbert: Add blank lines before comments to improve readability?
```suggestion
    # Touch again

```

flowmachine/tests/test_cache_utils.py
```diff
@@ -0,0 +1,350 @@
...
+    new_score = touch_cache(flowmachine_connect, dl.md5)
+    updated_cache_score = cachey_scorer.touch("dl")
+    assert updated_cache_score == pytest.approx(new_score)
+    # Add another unrelated cache record, which should have a higher initial score
```
maxalbert:
```suggestion

    # Add another unrelated cache record, which should have a higher initial score
```

flowmachine/tests/test_cache_utils.py
```diff
@@ -0,0 +1,350 @@
...
+    set_cache_half_life,
+    invalidate_cache_by_id,
+)
+from flowmachine.features import daily_location
+
+
+def test_scoring(flowmachine_connect):
+    """Test that score updating algorithm is correct by comparing to cachey as reference implementation"""
+    dl = daily_location("2016-01-01").store().result()
+    dl_time = compute_time(flowmachine_connect, dl.md5)
```
maxalbert: A little nitpicky, but I suggest renaming `compute_time()` to `get_ compute_time()`. The reason is that otherwise it sounds like it's an instruction: "compute the time [... of what?]", which made me trip over this line a few times.

(Just for consistency we could also rename `size_of_table()` to `get_ size_of_table()`, although there is less opportunity for misundestanding there.)

flowmachine/flowmachine/core/query.py
maxalbert: Maybe `query_as_table`?

maxalbert: (Or even `this_query_as_table`?)

flowmachine/flowmachine/core/cache.py
maxalbert: I see, thanks for the explanation. Related to my comment a bit further down, do you think in the medium run it is worth storing Table queries elsewhere so that we don't need to treat them as exceptions all over the place?

greenape: Definitely worth investigating that.

flowmachine/tests/test_cache.py
```diff
@@ -70,11 +71,7 @@ def test_store_cache_simple(flowmachine_connect):
     dl1.store().result()
     # Should be stored
     assert dl1.is_stored
-    in_cache = bool(
-        flowmachine_connect.fetch(
-            f"SELECT * FROM cache.cached WHERE query_id='{dl1.md5}'"
-        )
-    )
+    in_cache = cache_table_exists(flowmachine_connect, dl1.md5)
```
maxalbert: Thanks for adding the helper function `cache_table_exists()`, it reads a lot cleaner now!

The `in_cache` variable isn't really needed any more though, so maybe change this to `assert cache_table_exists(flowmachine_connect, dl1.md5)`? (Also applies to the other tests.)

flowmachine/tests/test_cache.py
```diff
@@ -88,11 +85,7 @@ def test_store_cache_multi(flowmachine_connect):
...
     # Should be stored
     assert hl1.is_stored
-    in_cache = bool(
-        flowmachine_connect.fetch(
-            f"SELECT * FROM cache.cached WHERE query_id='{hl1.md5}'"
-        )
-    )
+    in_cache = cache_table_exists(flowmachine_connect, hl1.md5)
 
     assert in_cache
```
maxalbert: Should this also check that the underlying daily_location now exists in the cache? Apologies if this is actually checked in a different test further down which I haven't seen yet.

flowmachine/tests/test_cache.py
```diff
@@ -149,18 +132,8 @@ def test_invalidate_cache_multi(flowmachine_connect):
...
-        )
-    )
-    assert not in_cache
-    in_cache = bool(
-        flowmachine_connect.fetch(
-            f"SELECT * FROM cache.cached WHERE query_id='{hl1.md5}'"
-        )
-    )
-    assert not in_cache
+    assert not cache_table_exists(flowmachine_connect, dl1.md5)
```
maxalbert: It just occurred to me that for this test to work properly we assume that the query _was_ previously stored. I know we check this in a different test, just wondering whether it is best to make all tests work independently (in case someone later removes the other test and/or changes something about the functionality)?

If so, we could add
```
assert dl1.is_stored
assert hl1.is_stored
```
above, before `dl1.invalidate_db_cache()`.

maxalbert: (A similar observation applies to the other `test_invalidate_...` tests.)

flowmachine/tests/test_cache_utils.py
```diff
@@ -0,0 +1,396 @@
...
+
+def test_scoring(flowmachine_connect):
+    """
+    Test that score updating algorithm is correct by comparing to cachey as reference implementation
+    """
+    dl = daily_location("2016-01-01").store().result()
+    dl_time = get_compute_time(flowmachine_connect, dl.md5)
+    dl_size = get_size_of_table(flowmachine_connect, *dl.table_name.split(".")[::-1])
+    initial_score = get_score(flowmachine_connect, dl.md5)
+    cachey_scorer = Scorer(1000)
```
maxalbert:
```suggestion
    cachey_scorer = Scorer(halflife=1000)
```

flowmachine/tests/test_cache_utils.py
```diff
@@ -0,0 +1,396 @@
...
+def test_touch_cache_record_for_query(flowmachine_connect):
+    """
+    Touching a cache record for a query should update access count, last accessed, & counter.
+    """
+    table = daily_location("2016-01-01").store().result()
+
+    assert (
+        1
+        == flowmachine_connect.fetch(
+            f"SELECT access_count FROM cache.cached WHERE query_id='{table.md5}'"
```
maxalbert: It feels to me these SQL statements could also be useful helper functions? E.g. `get_access_count(table.md5)`, `get_last_access_time(table.md5)`?

flowmachine/tests/test_cache_utils.py
```diff
@@ -0,0 +1,396 @@
...
+    touch_cache(flowmachine_connect, table.md5)
+    assert (
+        2
+        == flowmachine_connect.fetch(
+            f"SELECT access_count FROM cache.cached WHERE query_id='{table.md5}'"
+        )[0][0]
+    )
+    # Two cache touches should have been recorded
+    assert (
+        4 == flowmachine_connect.fetch("SELECT nextval('cache.cache_touches');")[0][0]
```
maxalbert: This bit is not clear to me. Why are we manually selecting the next value of the `cache.cache_touches` counter?

flowmachine/tests/test_cache_utils.py
```diff
@@ -0,0 +1,396 @@
...
+    )
+    assert (
+        accessed_at
+        < flowmachine_connect.fetch(
+            f"SELECT last_accessed FROM cache.cached WHERE query_id='{table.md5}'"
+        )[0][0]
+    )
+
+
+def test_compute_time():
```
maxalbert:
```suggestion
def test_get_compute_time():
```

flowmachine/tests/test_cache_utils.py
```diff
@@ -0,0 +1,396 @@
...
+    Test that shrink_below_size removes cache records in ascending score order."""
+    dl = daily_location("2016-01-01").store().result()
+    dl_aggregate = dl.aggregate().store().result()
+    flowmachine_connect.engine.execute(
+        f"UPDATE cache.cached SET cache_score_multiplier = 100 WHERE query_id='{dl_aggregate.md5}'"
+    )
+    flowmachine_connect.engine.execute(
+        f"UPDATE cache.cached SET cache_score_multiplier = 0.5 WHERE query_id='{dl.md5}'"
+    )
+    table_size = get_size_of_table(flowmachine_connect, *dl.table_name.split(".")[::-1])
```
maxalbert: This pattern of passing the split table_name/schema as `*args` occurs a few times but is a little cryptic to read at first. I was wondering if it's worth to optionally lett `get_size_of_table()` accept a fully qualified table name so that we don't need to do this splitting every time?

So the following would then be both allowed:
```
get_size_of_table(connection, table_name="cells", table_schema="infrastructure")
get_size_of_table(connection, table_name="infrastructure.cells")
```
(In the second case the `table_schema` must be None, of course.)

flowmachine/tests/test_cache_utils.py
greenape: You can't get a global current, or last value for a Postgres sequence - those are only available within a connection. What you _can_ do is get the _next_ value, and infer from that what the last one is.

flowmachine/tests/test_to_sql.py
```diff
@@ -28,29 +28,20 @@ def test_stores_table(flowmachine_connect):
...
-    assert "test_view" in flowmachine_connect.inspector.get_view_names(schema="tests")
-
-
 def test_can_force_rewrite(flowmachine_connect, get_length):
     """
     Test that we can force the rewrite of a test to the database.
     """
     query = EventTableSubset("2016-01-01", "2016-01-01 01:00:00")
-    query.to_sql(schema="tests", name="test_rewrite")
+    query.to_sql(name="test_rewrite", schema="tests").result()
```
maxalbert: I was wondering why we need the extra `.result()`, but then I realised it's because `to_sql()` now returns a future.

When I looked up what `to_sql()` actually does it says: "Store the result of the calculation back into the database." To be honest, I find this a little confusing. Either the name is wrong (`to_sql()` sounds to me like it should simply return the SQL string of a Query object) or its primary purpose is something different.

Either way (and also looking at the implementation in `query.py`), it feels like this needs some clarification and tidying up. Not for this PR, but I wanted to note it here so I don't forget.

greenape: I believe it was because of this: https://pandas.pydata.org/pandas-docs/stable/generated/pandas.DataFrame.to_sql.html

flowmachine/tests/test_cache_utils.py
maxalbert: Yes, I like that idea!

Btw, happy to leave this to a separate PR in case it is too much hassle to include it in this one.

flowmachine/tests/test_cache_utils.py
maxalbert: Ah, I see. Could be useful to add this info to the comment.

flowmachine/tests/test_to_sql.py
maxalbert: Oh I see, yes that could well be the reason. Do you feel the name is adequate? My personal impression is that our interface & use case is sufficiently different from pandas now that the name is more confusing than it helps.

maxalbert:
Oh, and there is a file called `cache_scores.csv` which was added accidentally and needs to be stripped out before merging. I suggest to strip out the commit where it was added and force-push the branch again to avoid it being in the history (because it's > 1MB).

greenape:
The bubbling - I've added an explicit cast python-side to these, so you'll get a ValueError if you give it something silly.

PR #252, opened 2019-01-16
Clone using https in FlowDB Dockerfile

### LINKED ISSUE #251: Use git:// instead of https:// to clone in FlowDB Dockerfile
Currently, postgresql pldebugger is cloned using the `git://` protocol in the FlowDB Dockerfile. This isn't a problem, but could cause complications if running on a server on which the port used by `git://` is blocked by a firewall. Since the clone also works using `https://`, let's use that instead.

jch1g10:
Closes #251

### I have:

### Description

Two changes to the command to clone `git.postgresql.org/git/pldebugger.git` in the FlowDB Dockerfile:
- Use `https://` protocol instead of `git://`
- Remove `-c http.sslVerify=false` option, since it appears to be unnecessary now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,6 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Fixed being unable to build if the port used by `git://` is not open
 
 ### Removed
-- Removed `-c http.sslVerify=false` option from the command to clone postgres pldebugger repository in flowdb Dockerfile
 
 ## [0.2.0]
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,11 +7,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Added
 
 ### Changed
-Use `https://` protocol instead of `git://` to clone postgres pldebugger repository in flowdb Dockerfile
+- Use `https://` protocol instead of `git://` to clone postgres pldebugger repository in flowdb Dockerfile
 
 ### Fixed
 
 ### Removed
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -7,10 +7,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Added
 
 ### Changed
+- Use `https://` protocol instead of `git://` to clone postgres pldebugger repository in flowdb Dockerfile
```
greenape: Maybe make the changelog entry for this a fixed one instead - the intent of this is to fix being unable to build if the port git:// isn't open, right?

CHANGELOG.md
```diff
@@ -9,8 +9,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Changed
 
 ### Fixed
+- Fixed being unable to build if the port used by `git://` is not open
 
 ### Removed
+- Removed `-c http.sslVerify=false` option from the command to clone postgres pldebugger repository in flowdb Dockerfile
```
greenape: Drop this line?

jch1g10: The fact that `-c http.sslVerify=false` was there in the first place implies that somebody needed it to get the clone to work. It works fine without this, in my experience, but I wonder whether there may be any cases where it's still necessary. I put this line in so that we have an explicit record in the changelog in case anybody sees a problem as a result of this in the future. I could be convinced otherwise, though, if you feel differently.

CHANGELOG.md
maxalbert: Yup, correct.

PR #255, opened 2019-01-17
Support running FlowDB as arbitrary user via --user flag to docker run

### LINKED ISSUE #247: Remove uid/gid build arguments to FlowDB
FlowDB currently takes two build arguments `POSTGRES_UID` and `POSTGRES_GID` which _were_ necessary to enable one to work around conflicts between the user Postgres ran as, and the uids on the host system. Per 
https://github.com/docker-library/postgres/pull/253 and 
https://github.com/docker-library/postgres/pull/448 the upstream image now supports setting these via `--user` to match an existing user on the host system. Using this method also means docker won't clobber permissions when spinning up the Postgres instance if you don't have a user matching the internal uid & gid.

We should remove the old workaround, and explain how to do this in the docs.

greenape:
Closes #247

### I have:

### Description

This deprecates setting the uid and gid used by FlowDB during the image build, in favour of setting it by passing `--user uid:gid` to `docker run` or the `user` key in a compose file.

Removes a bunch of legacy workarounds in the Dockerfiles, and updates the upstream Postgres image to the most current one.

This change is primarily important in scenarios where you are persisting FlowDB via a bind mount, or want to run ETL by bind mounting a directory on the host. Previously, one had to ensure that you'd either built the image with the Postgres uid and gid matching an existing user on the host, or that you could create a new user matching the uid and gid expected. You can now use any user that exists on the host, and specify that at runtime.

(Unrelatedly, I've also fixed the GitHub readme, which had diverged from the docs one)

### DIFF
--- a/flowdb/testdata/Dockerfile
+++ b/flowdb/testdata/Dockerfile
@@ -18,7 +18,7 @@ FROM flowminder/flowdb:${CODE_VERSION}
 #
 
 RUN mkdir -p /docker-entrypoint-initdb.d/sql/testdata/
-COPY --chown postgres ./bin/z_ingest_test_data.sh /docker-entrypoint-initdb.d/
-ADD --chown postgres ./test_data/sql/* /docker-entrypoint-initdb.d/sql/testdata/
-ADD --chown postgres ./test_data/data/ /docker-entrypoint-initdb.d/data/
-COPY  --chown postgres ./test_data/data/*.csv /docker-entrypoint-initdb.d/data/csv/
+COPY --chown=postgres ./bin/z_ingest_test_data.sh /docker-entrypoint-initdb.d/
+ADD --chown=postgres ./test_data/sql/* /docker-entrypoint-initdb.d/sql/testdata/
+ADD --chown=postgres ./test_data/data/ /docker-entrypoint-initdb.d/data/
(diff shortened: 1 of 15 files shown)


flowdb/testdata/Dockerfile
```diff
@@ -18,7 +18,9 @@ FROM flowminder/flowdb:${CODE_VERSION}
...
 RUN mkdir -p /docker-entrypoint-initdb.d/sql/testdata/
-COPY ./bin/z_ingest_test_data.sh /docker-entrypoint-initdb.d/
-ADD ./test_data/sql/* /docker-entrypoint-initdb.d/sql/testdata/
-ADD ./test_data/data/ /docker-entrypoint-initdb.d/data/
-COPY ./test_data/data/*.csv /docker-entrypoint-initdb.d/data/csv/
+COPY --chown=postgres ./bin/z_ingest_test_data.sh /docker-entrypoint-initdb.d/
+ADD --chown=postgres ./test_data/sql/* /docker-entrypoint-initdb.d/sql/testdata/
+ADD --chown=postgres ./test_data/data/ /docker-entrypoint-initdb.d/data/
+COPY  --chown=postgres ./test_data/data/*.csv /docker-entrypoint-initdb.d/data/csv/
+# Need to make postgres is owner of any subdirectrories
```
maxalbert:
```suggestion
# Need to make postgres owner of any subdirectories
```

Issue #265, opened 2019-01-21
Better development setup/workflow needed


maxalbert:
While FlowKit's production environment is fairly smooth to set up due to the dockerisation of our stack, setting up a development environment is currently very fiddly as the steps are different for each component and not well documented. We could also do with putting more scaffolding in place to make it easier to "jump in" at different levels of our stack if debugging is needed.

One thing that would help is to dockerise our development environment, too. The docker containers should have the source directories mounted inside them and the components should be started in such a way to enable live reload in case of any code changes. Tools like [watchdog](https://pythonhosted.org/watchdog/) can help with this.

Useful references:
- https://hackernoon.com/efficient-development-with-docker-and-docker-compose-e354b4d24831
- https://testdriven.io/courses/microservices-with-docker-flask-and-react/)

As a second step we can also consider dockerising our tests so that they can run continuously during development whenever the code changes (for which it would help to bring the test runtimes down a bit, esp. for flowmachine).

PR #269, opened 2019-01-22
Dev cleanup

maxalbert:
This addresses most of #265, but does not yet allow live reload of Flowauth (this will be addressed in a separate PR).

### I have:

### Description

This PR tidies up the `docker-compose-dev.yml` file so that it actually starts up a proper development environment. This has the following goals:

- Provide a "one-button-press" way of spinning up a dev environment (as opposed to having to spin up each component separately with a custom command)
- Enable "live reload" of `flowmachine` and `flowapi` during development so that source code changes are picked up immediately without a need to restart them (or even re-build docker containers).

This also moves our dev setup closer to the production one because both now use docker containers (with only small variations in how the packages are installed inside the containers, which can be minimised further in the future).

The main change was to change the `docker-compose-dev.yml` file so that the source directories for `flowmachine` and `flowapi` are mounted into the containers, and to start the processes via [watchdog](https://pythonhosted.org/watchdog/) so that any changes to `*.py` files inside the source folders trigger a restart of the process (inside the running docker container). To achieve th elatter, I added separate `Dockerfile-dev` files for `flowmachine` and `flowapi` (which I aimed to keep as similar to the production `Dockerfile`s as possible).

Other changes:

- Previously `flowdb` and `flowapi` used additional volumes to store the data and logs, which I have removed. I'm not entirely sure whether we will _want_ volumes for some services in the future also during development (for example, I can potentially see it being useful for Flowauth and perhaps for FlowDB) but for now I though it would be better to remove them until we see a clear need.

- Added a section to the docs with some basic explanations on how to spin up a dev environment and test that it works. This will need further cleanup but should be sufficient to get a new developer started.

- I added a small helper function (plus test) in `flowmachine/flowmachine/core/server/utils.py` called `send_message_and_receive_reply()` which allows to send a JSON message to the flowmachine via `zmq`, which is helpful for interactive testing. Note that this is very similar to an existing helper function in the flowmachine tests, so the latter should probably be removed in a subsequent PR and we should use the new one for the tests also.

Still to do:

- Enable live reload of Flowauth (this shouldn't require much work but I didn't want to delay this PR further)
- Enable proper configurations for Flowapi (for example "production" vs. "development"; this is set a as `CONFIG` env var in the Dockerfiles but I don't think it currently has any effect).


flowapi/Dockerfile-dev
```diff
@@ -0,0 +1,14 @@
...
+
+FROM python:3.7
+
+WORKDIR /flowapi
+COPY Pipfile* /
+RUN pip install pipenv && pipenv install --dev
+COPY app /flowapi/app
+RUN mkdir /var/log/flowapi
+VOLUME /var/log/flowapi
+CMD pipenv run watchmedo auto-restart --recursive --patterns="*.py" --directory="." pipenv run ./app/start.sh
```
greenape: Not objecting, but what's the rationale for using this as opposed to the autoreload that's built into flask & quart?

maxalbert: Good point. For the time being it was just expediency to get it working quickly (and I was going to address this in a subsequent PR when adding proper dev/prod configurations for flowapi) but there is no reason to not already use this here. I'll switch it over.

maxalbert: On that note, is the `CONFIG` variable (which is set in the `Dockerfile` and also in `flowapi/tests/unit/conftest.py`) actually used anywhere, or can I remove this and only set `FLASK_ENV=development`?

maxalbert: Also, the `app/start.sh` script runs the app via `hypercorn`. Does this actuallly pick up the `FLASK_ENV` variable and trigger a restart of the app upon source code changes?

maxalbert: Hmm, a quick experiment seems to suggest that setting `QUART_ENV=development` in the `Dockerfile-dev` (before calling `pipenv run app/start.sh`) is not sufficient to enable a live reload. Not sure what I'm missing, but perhaps worth investigating later and addressing in a separate PR?

flowmachine/Pipfile
```diff
@@ -32,8 +32,8 @@ pytest-asyncio = "*"
 pytest-cov = "*"
 pytest-testmon = "*"
 asynctest = "*"
-flowmachine = {editable = true,path = "."}
```
greenape: Why's this removed?

maxalbert: Because otherwise we can't run `pipenv install --dev` (which implicitly installs flowmachine in case this line is present) before `COPY . /app/` in the Dockerfile. However, we want to run `pipenv install --dev` beforehand to take advantage of docker layer caching (so that we don't need to reinstall all dependencies if the source code changes).

maxalbert: (Otherwise I would have preferred to leave it in but I didn't see a way of having both.)

docker-compose-dev.yml
```diff
@@ -22,28 +26,33 @@ services:
...
       - FM_PASSWORD=${FM_PASSWORD:-foo}
       - API_PASSWORD=${API_PASSWORD:-foo}
+      - CACHE_SIZE=${CACHE_SIZE:-""}  # if empty, defaults to using 10% of the available disk space
+      - CACHE_HALF_LIFE=${CACHE_HALF_LIFE:-1000}
     shm_size: 1G
     tty: true
     stdin_open: true
     restart: always
-    volumes:
-      - data_volume_flowdb:/var/lib/postgresql/data
```
greenape: I'd put this back I think, purely because the upstream Postgres image creates a volume that maps to this dir _anyway_, and this at least makes it explicit.

maxalbert: The reason I took the flowdb volume out is because if it is present then for flowdb to pick up source code changes it is _not_ sufficient to simply re-build the docker image and restart the container. In addition, you also need to remember to delete the volume before restarting the container, which is easy to forget and has led to lots of confusion and long debugging sessions in the past. Therefore I deemed it safer to remove the volume.

maxalbert:
@greenape I added a few commits to refactor the logging according to your suggestions (mainly with the intention to make the tests pass - this probably needs some further cleaning up). Would you mind taking a quick look whether you are happy with the new changes?

flowmachine/flowmachine/core/init.py
```diff
@@ -95,11 +98,6 @@ def connect(
         if log_level is None
         else log_level
     )
-    log_file = (
```
greenape: Don't we want write_log_file to be settable via env var as well?

Makefile
```diff
@@ -9,58 +9,92 @@
...
-	docker-compose -f $(DOCKER_COMPOSE_FILE_DEV) up -d $(DOCKER_SERVICES)
+	docker-compose -f $(DOCKER_COMPOSE_FILE_DEV) up -d --build $(DOCKER_SERVICES)
 
 down:
 	docker-compose -f $(DOCKER_COMPOSE_FILE_DEV) down
 
-flowdb_testdata-up:
-	docker-compose -f $(DOCKER_COMPOSE_FILE_DEV) up -d flowdb_testdata
+
+flowdb-up: flowdb-build
```
jch1g10: Is there any reason we can't use pattern rules to shorten this? e.g.
```
%-up: %-build
	docker-compose -f $(DOCKER_COMPOSE_FILE_DEV) up -d --build $*
```

maxalbert: Yeah, I've been going back and forth on this one.

Your solution of using patterns was my first instinct, but the reasons I went for the longer version are:
- discoverability / ease of understanding for someone not familiar with Makefile syntax
- you can't tab-complete patterns, so typing something like `make flowmachine_testdata-up` becomes tedious.

No strong opinion though, so I'm happy to switch to patterns if you think it makes it more readable.

jch1g10: Good points. Yes, that makes sense.

maxalbert: It's a good point, though. I've added a comment in the Makefile to explain the repetition.

flowmachine/flowmachine/core/init.py
```diff
@@ -185,15 +191,18 @@ def connect(
...
+def _init_logging(log_level, write_log_file):
     """
 
     Parameters
     ----------
     log_level : str
         Level to emit logs at
-    log_file : str
-        Path to file to write logs to
+    write_log_file : bool, default False
```
jch1g10: `write_log_file` has no default in `_init_logging`

integration_tests/tests/flowmachine_server_tests/conftest.py
```diff
@@ -11,17 +11,30 @@
...
+    the environment variable FLOWMACHINE_ZMQ_PORT or else 5555.
+    """
+    return os.getenv("FLOWMACHINE_ZMQ_PORT", "5555")
+
+
+@pytest.fixture(scope="session")
+def zmq_url(zmq_host, zmq_port):
+    """
+    Return the URL where to connect to zeromq when running the tests.
+    This is currently hard-coded to "tcp://localhost:5555".
```
jch1g10: Is this hard-coded? Looks like setting `FLOWMACHINE_ZMQ_HOST` and `FLOWMACHINE_ZMQ_PORT` should override this.

flowmachine/flowmachine/core/init.py
maxalbert: Good catch. The default value is set in `connect()` and then passed in here, so we don't need a default here. I'll remove it from the docstring.

integration_tests/tests/flowmachine_server_tests/conftest.py
maxalbert: Indeed 👍

Issue #138, opened 2018-12-05
Remove GeoDataMixin


greenape:
After providing #135 and #136 we should be able to remove the GeoDataMixin, since this obviates the need to return query results formatted as geojson, in favour of that being done by the end user combining data frames with geo as they see fit.

greenape:
I'm no longer as convinced about doing this as I was, because we've moved away from deprecating flowmachine as a standalone library.

PR #277, opened 2019-01-23
Geography API route

### LINKED ISSUE #135: Add API routes for obtaining geography
The API should provide a route for downloading geographies as GeoJSON, and FlowClient should provide a function to facilitate this.

jch1g10:
Closes #135

### I have:

### Description

Adds an API route for getting geography data from FlowDB, and a corresponding FlowClient function to return this GeoJSON object as a dict. Permissions checked are `"geography"` - `"get_result"` - `<aggregation>` (`"geography"` - `"run"`/`"poll"` are not used at present).
Currently this does interact with FlowMachine, since the code to generate the relevant SQL query exists in `GeoDataMixin`, but this could change when we address #138.

#### Other changes
- Changes zmq messages to send `"error"` status, as FlowAPI expects, instead of `"rejected"`.
- Fixes FlowClient test `test_get_result_by_id_error`, which was catching an error from `query_is_ready` instead of `get_result_by_query_id`.


flowapi/app/check_claims.py
```diff
@@ -0,0 +1,156 @@
...
+                "AUTHENTICATED",
+                request_id=request.request_id,
+                route=request.path,
+                user=get_jwt_identity(),
+                src_ip=request.headers.get("Remote-Addr"),
+                json_payload=json_payload,
+            )
+
+            # Get query kind
+            if request.path.split("/")[3] == "geography":
```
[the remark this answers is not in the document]
greenape: (Like that this is split off into a separate file now tho!)

flowapi/tests/unit/test_geography.py
```diff
@@ -0,0 +1,76 @@
...
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import pytest
+from json import loads
+
+from asynctest import return_once
+
+
+@pytest.mark.asyncio
+async def test_get_geography(app, dummy_zmq_server, access_token_builder):
```
greenape: Can we test that this is going to construct valid Geojson as well?

flowapi/app/geography.py
```diff
@@ -0,0 +1,81 @@
...
+    #  Get the reply.
+    message = await request.socket.recv_json()
+    current_app.logger.debug(f"Got message: {message}")
+    if message["status"] == "done":
+        results_streamer = stream_with_context(assemble_geojson_feature_collection)(
+            message["sql"], message["crs"]
+        )
+        mimetype = "application/geo+json"
+
+        current_app.logger.debug("Returning.")
```
maxalbert: Maybe make this message slightly more informative (e.g. to say what we're returning)?

flowmachine/flowmachine/core/mixins/geodata_mixin.py
```diff
@@ -243,16 +244,32 @@ def _get_geojson(self, proj4):
...
-        ]
+        features = [x[0] for x in self.connection.fetch(self.geojson_query(crs=proj4))]
         js = {
             "properties": {"crs": proj4},
             "type": "FeatureCollection",
             "features": features,
         }
         return js
 
+    def get_proj4_string(self, crs=None):
```
greenape: Not immediately clear why this is necessary, given it just wraps the util method?

integration_tests/tests/test_queries.py
```diff
@@ -47,3 +47,22 @@ def test_run_query(query_kind, params, access_token_builder, api_host):
...
+                "geography": {
+                    "permissions": permissions_types,
+                    "spatial_aggregation": aggregation_types,
+                }
+            }
+        ),
+    )
+    result_geojson = flowclient.get_geography(con, "admin3")
+    assert "FeatureCollection" == result_geojson["type"]
+    assert "crs" in result_geojson["properties"]
```
greenape: The crs property shouldn't actually be there - I'd like this to actually validate the geojson that comes back instead.

flowapi/app/geography.py
```diff
@@ -0,0 +1,81 @@
...
+            {
+                "Transfer-Encoding": "chunked",
+                "Content-Disposition": f"attachment;filename={aggregation_unit}.geojson",
+                "Content-type": mimetype,
+            },
+        )
+    elif message["status"] == "error":
+        return jsonify({"status": "Error", "msg": message["error"]}), 403
+    else:
+        return jsonify({}), 404
```
maxalbert: I don't know what best practice is here, since 404 already indicates that this route doesn't exist, but is it worth adding an explicit human-readable error message too, to make it slightly easier to debug the failure?

flowapi/app/geography.py
```diff
@@ -0,0 +1,81 @@
...
+
+    Yields
+    ------
+    bytes
+        Encoded lines of JSON
+
+    """
+    logger = current_app.logger
+    pool = current_app.pool
+    yield f'{{"properties":{{"crs":"{crs}"}}, "type":"FeatureCollection", "features":['.encode()
```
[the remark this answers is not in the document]
jch1g10: Yes, I was wondering about this. Is it safe to assume all data in FlowDB will be wsg84, or are there any situations where we would need to project from a different crs when constructing the geojson?

flowapi/tests/unit/test_geography.py
```diff
@@ -0,0 +1,76 @@
...
+
+    token = access_token_builder(
+        {
+            "geography": {
+                "permissions": {"get_result": True},
+                "spatial_aggregation": ["DUMMY_AGGREGATION"],
+            }
+        }
+    )
+    dummy_zmq_server.side_effect = ({"status": status, "error": "Some error., "},)
```
maxalbert: Is the comma and blank space at the end intentional?

flowapi/app/geography.py
```diff
@@ -0,0 +1,81 @@
...
+    prepend = ""
+    logger.debug("Starting generator.")
+    async with pool.acquire() as connection:
```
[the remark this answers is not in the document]
jch1g10: Almost, but not quite. In the regular sql->json streamer we yield `json.dumps(dict(row.items()))` for each row, but here we yield `row[0]` since we just want the json object from the (single) column.

flowmachine/flowmachine/core/mixins/geodata_mixin.py
```diff
@@ -243,16 +244,32 @@ def _get_geojson(self, proj4):
...
+        crs : int or str
+            Optionally give an integer srid, or valid proj4 string
+        
+        Returns
+        -------
+        str
+            proj4 string
+        """
+        # TODO: Some refactoring would be beneficial here.
+        proj4_string = proj4string(self.connection, crs)
```
maxalbert: Minor thing, but here we're creating a local variable with the same name as a pre-existing function, and then immediately return it. I suggest to simply change this to
```
return proj4string(self.connection, crs)
```

flowapi/app/geography.py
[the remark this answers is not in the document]
greenape: The FM-side code does do crs conversion, so I'd explicitly convert it to wsg84 (in general, will be that by default anyway, but best to be safe). (This is really an error in my original implementation.)

flowapi/app/geography.py
```diff
@@ -0,0 +1,81 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from quart import Blueprint, current_app, request, stream_with_context, jsonify
+from .check_claims import check_claims
+
+blueprint = Blueprint(__name__, __name__)
```
maxalbert: I suggest to pass an explicit name for the first argument here:
```suggestion
blueprint = Blueprint("geography", __name__)
```

We should also do this for the blueprint in `run_query.py` and tweak the call to `url_for(...)` accordingly.

flowmachine/flowmachine/core/mixins/geodata_mixin.py
jch1g10: It was necessary to get the proj4 string in `server.py`, so that it could be returned to FlowAPI to include in the geojson. But since that shouldn't be there, this isn't necessary any more.

flowapi/app/stream_results.py
```diff
@@ -0,0 +1,52 @@
...
+
+
+async def stream_result_as_json(sql_query, headers={}, result_name="query_result"):
+    """
+    Generate a JSON representation of a query result.
+    Parameters
+    ----------
+    sql_query : str
+        SQL query to stream output of
+    headers : dict
```
maxalbert: Why the name "headers" here? From the implementation it seems that these are included in the output before the actual streaming output, but this isn't apparent from the description. Can we either change the parameter name or explain this a bit more?

flowapi/app/stream_results.py
```diff
@@ -0,0 +1,52 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import ujson as json
+from quart import current_app
+
+
+async def stream_result_as_json(sql_query, headers={}, result_name="query_result"):
+    """
+    Generate a JSON representation of a query result.
```
maxalbert: Missing blank line:
```suggestion
    Generate a JSON representation of a query result.

```

flowmachine/flowmachine/core/server/query_proxy.py
```diff
@@ -72,6 +72,13 @@ class QueryProxyError(Exception):
     """
 
 
+class InvalidGeographyError(Exception):
+    """
+    Custom exception to indicate that the aggregation unit for a
+    geography query is invalid (corresponds to a 404 error in the API).
```
maxalbert: Nice! I like the clear explanation in the docstring, including mentioning the 404 error.

maxalbert:
Somehow the status is not successfully updating (it says: _"Expected — Waiting for status to be reported"_).

Seems like a failure of CircleCI to report the successful CI run, or does it have anything to do with the changes requested by @greenape that still need to be approved?

PR #279, opened 2019-01-24
Hartigan api

### LINKED ISSUE #139: Enable Hartigan via API
Hartigan clustering should be available via the API.

greenape:
Closes #139 

### I have:

### Description

Exposes aggregates of, and od matrices between, meaningful locations produced using the methodology developed by @gzagatti and @mgonzalezcanudas based on Isaacman et al for the World Bank's urbanisation review.

# Additions
- Adds `MeaningfulLocations`, which bundles Hartigan clustering with cluster scoring and labelling
- Adds `MeaningfulLocationAggregate` and `MeaningfulLocationsOD`
- Exposes `MeaningfulLocationAggregate` and `MeaningfulLocationsOD` via FlowAPI as `meaningful_locations_aggregate` and `meaningful_locations_od_matrix` query kinds
- Adds a function for obtaining aggregates to FlowClient
- Adds two functions for obtaining od matrices to FlowClient (between two labels, and between the same label on different date brackets)

# Misc
- Cleans up `CallDays`, `HartiganCluster`, `LabelEventScore` and `EventScore` classes and constructors
- Splits `LabelEventScore` and `EventScore` to separate files
- Switches all the compose files to use 'latest' as default instead of now defunct 'master' tag
- Skips date checking in tests by default, and adds a mark `check_available_dates` to turn it on
- Makes md5 hashing of queries a little more robust, by dumping dictionaries to json strings, which properly handles nested dicts
- Adds a bunch of smoke tests for `construct_query_object`

(Haven't included a worked example here, because I'm having a degree of difficulty figuring out how to produce one which will work with the default test data)


flowclient/flowclient/client.py
```diff
@@ -455,6 +455,320 @@ def daily_location(
...
+    labels: Dict[str, Dict[str, dict]],
+    tower_cluster_radius: float = 1.0,
+    tower_cluster_buffer: float = 0.0,
+    tower_cluster_call_threshold: int = 0,
+    tower_dow_scores: Dict[str, float] = {
+        "monday": 1,
+        "tuesday": 1,
+        "wednesday": 1,
+        "thursday": 0,
+        "friday": -1,
```
maxalbert: These values seem very context-dependent. Should we even provide a default here? Might be better to force the user to supply these values expicitly?

flowclient/flowclient/client.py
```diff
@@ -455,6 +455,320 @@ def daily_location(
...
+        "saturday": -1,
+        "sunday": -1,
+    },
+    tower_hour_of_day_scores: List[float] = [
```
maxalbert: Similar to `tower_dow_scores`, not sure if we can provide a generically useful default value here.

flowclient/flowclient/client.py
```diff
@@ -455,6 +455,320 @@ def daily_location(
...
 
+def _meaningful_locations(
+    start_date: str,
+    stop_date: str,
+    label: str,
+    labels: Dict[str, Dict[str, dict]],
+    tower_cluster_radius: float = 1.0,
+    tower_cluster_buffer: float = 0.0,
+    tower_cluster_call_threshold: int = 0,
+    tower_dow_scores: Dict[str, float] = {
```
maxalbert: Personally, I always struggle to interpret variable names that have a mix of full words and abbreviations in them.

I think I'd be in favour of a longer but more explicit name such as `tower_day_of_week_scores` (or perhaps `tower_weekday_scores`), but I realise this is personal preference so I'm also ok to leave it as it is if you prefer.

flowclient/flowclient/client.py
```diff
@@ -455,6 +455,320 @@ def daily_location(
     }
 
 
+def _meaningful_locations(
+    start_date: str,
+    stop_date: str,
+    label: str,
+    labels: Dict[str, Dict[str, dict]],
```
maxalbert: What do `label` and `labels` refer to? It's somewhat confusing that these variables have such similar names, can we make them more explicit (or provide a docstring to explain them)?

flowclient/flowclient/client.py
```diff
@@ -455,6 +455,320 @@ def daily_location(
...
+                },
+            },
+            "scores": {
+                "query_kind": "event_score",
+                "params": {
+                    "score_hour": tower_hour_of_day_scores,
+                    "score_dow": tower_dow_scores,
+                    "start": start_date,
+                    "stop": stop_date,
+                    "level": "versioned-site",
```
maxalbert: Should this still be called `level`, or can we rename it to `aggregation_unit`?

flowclient/flowclient/client.py
maxalbert: (At first I wasn't even sure if these are alternative ways of providing a value for the same thing.)

flowclient/flowclient/client.py
```diff
@@ -455,6 +455,320 @@ def daily_location(
...
+        0,
+        0,
+        0,
+        0,
+        -1,
+        -1,
+        -1,
+    ],
+    subscriber_subset: Union[dict, None] = None,
+) -> dict:
```
maxalbert: Worth adding a docstring to explain the meaning of parameters (or alternatively to refer to some other piece of documentation that explains them)? Esp. in light of the meaning of `label` vs. `labels` as mentioned above.

flowclient/flowclient/client.py
```diff
@@ -455,6 +455,320 @@ def daily_location(
...
+                tower_cluster_call_threshold=tower_cluster_call_threshold,
+                tower_cluster_radius=tower_cluster_radius,
+                tower_dow_scores=tower_dow_scores,
+                tower_hour_of_day_scores=tower_hour_of_day_scores,
+            ),
+        },
+    }
+
+
+def meaningful_locations_between_label_od_matrix(
```
maxalbert: See above: add docstring?

flowclient/flowclient/client.py
```diff
@@ -455,6 +455,320 @@ def daily_location(
...
+                tower_cluster_call_threshold=tower_cluster_call_threshold,
+                tower_cluster_radius=tower_cluster_radius,
+                tower_dow_scores=tower_dow_scores,
+                tower_hour_of_day_scores=tower_hour_of_day_scores,
+            ),
+        },
+    }
+
+
+def meaningful_locations_between_dates_od_matrix(
```
maxalbert: See above: add docstring?

flowmachine/flowmachine/features/subscriber/label_event_score.py
```diff
@@ -0,0 +1,192 @@
...
+    Parameters
+    ----------
+    scores : flowmachine.Query
+        A flowmachine.Query object.
+        This represents a table that contains scores which are used to label
+        a given location. This table must have a subscriber column (called subscriber).
+    labels : dict
+        A dictionary whose keys are the label names and the values geojson shapes,
+        specified hour of day, and day of week score, with hour of day score on the x axis
+        and day of week score on the y axis, where all values are in the range [-1, 1]
```
maxalbert:
```suggestion
        and day of week score on the y axis, where all scores are real numbers in the range [-1.0, +1.0]
```

maxalbert: To be honest, an example of a valid dict for `label` would really help here. I'm not sure I could write one down just from this description.

flowmachine/flowmachine/features/subscriber/label_event_score.py
```diff
@@ -0,0 +1,192 @@
...
+from .scores import EventScore
+from .subscriber_location_cluster import _JoinedHartiganCluster
+from ...core import Query
+
+
+class LabelEventScore(Query):
+    """
+    Represents a label event score class.
+
+    This class will label a table containing scores based on a labelling
```
maxalbert: What's the "output" of this labelling? Is it a new table, or will there be an added column to the existing table (or are the scores replaced with the labels)?

flowmachine/flowmachine/features/subscriber/label_event_score.py
```diff
@@ -0,0 +1,192 @@
...
+
+        super().__init__()
+
+    @staticmethod
+    def _make_bounds_dict(
+        geojson_labels: Dict[str, Dict[str, Any]]
+    ) -> Dict[str, BaseGeometry]:
+        """
+        Takes a dictionary of labels and bounds expressed as lists of geojson shapes
+        and returns a dictionary of labels and bounds expressed as Shapely polygons.
```
maxalbert: Do I understand correctly that the purpose of this helper method is to convert geojson objects into shapely polygons? From the name `_make_bounds_dict()` I would have expected that it constructs a new dictionary containing bounds. But AFAIU it basically transforms an existing dictionary? So maybe rename the method to `_convert_bounds_to_shapely_polygons` or similar?

flowmachine/flowmachine/features/subscriber/label_event_score.py
```diff
@@ -0,0 +1,192 @@
...
+    ):
+
+        self.scores = scores
+        if not isinstance(scores, Query):
+            raise TypeError(
+                "Scores must be of type Query, e.g. EventScores, Table, CustomQuery"
+            )
+
+        labels = LabelEventScore._make_bounds_dict(labels)
+        LabelEventScore.bounds_dict_has_overlaps(labels)
```
maxalbert: This reads a little confusingly, because the method name implies that it returns a boolean but the return value is ignored here. Can we rename it to `LabelEventScore.verify_bounds_dict_has_no_overlaps(labels)` or similar?

flowmachine/flowmachine/features/subscriber/label_event_score.py
```diff
@@ -0,0 +1,192 @@
...
+
+    def _make_query(self):
+
+        scores_cols = self.scores.column_names
+        scores = f"({self.scores.get_query()}) AS scores"
+
+        sql = f"""
+        SELECT COALESCE(score_bounds.label, 'unknown') as label, scores.* FROM {scores}
+        LEFT JOIN {LabelEventScore._get_sql_bound(self.labels)}
+        ON st_contains(score_bounds.geom, st_point(scores.score_hour, scores.score_dow))
```
maxalbert: Nitpick: I think in other places we are using capitalised versions of PostGIS commants. Since we're also uppercasing SQL commands, maybe best to stick to the versions used in the official PostGIS docs (e.g. `ST_Contains`, `ST_Point`)?

flowmachine/flowmachine/features/subscriber/label_event_score.py
```diff
@@ -0,0 +1,192 @@
...
+    ----------
+    scores : flowmachine.Query
+        A flowmachine.Query object.
+        This represents a table that contains scores which are used to label
+        a given location. This table must have a subscriber column (called subscriber).
+    labels : dict
+        A dictionary whose keys are the label names and the values geojson shapes,
+        specified hour of day, and day of week score, with hour of day score on the x axis
+        and day of week score on the y axis, where all values are in the range [-1, 1]
+    required :
```
maxalbert: Missing type (str or None?)

flowmachine/flowmachine/features/subscriber/label_event_score.py
```diff
@@ -0,0 +1,192 @@
...
+        Specifies a label which every subscriber must possess independently of
```
maxalbert: From reading this it's not clear to me whether `required` must be supplied for `LabelEventScore` to function correctly. Also, what is the difference in logic depending on whether or not it is supplied?

flowmachine/flowmachine/features/subscriber/label_event_score.py
maxalbert: Not sure if this is easy to do, but it would be great to have an example of the input and output in this docstring.

flowmachine/flowmachine/features/subscriber/label_event_score.py
```diff
@@ -0,0 +1,192 @@
...
+            """
+
+        return sql
+
+    @property
+    def column_names(self) -> List[str]:
+        return ["label"] + self.scores.column_names
+
+    @staticmethod
+    def _get_sql_bound(bounds: Dict[str, List[Dict[str, float]]]) -> str:
```
maxalbert:
```suggestion
    def _get_bound_as_sql(bounds: Dict[str, List[Dict[str, float]]]) -> str:
```

flowmachine/flowmachine/features/subscriber/label_event_score.py
```diff
@@ -0,0 +1,192 @@
...
+            SQL case statement
+
+        """
+        table_rows = []
+        for label, label_bounds in bounds.items():
+            table_rows.append(f"('{label}', '{label_bounds.wkt}'::geometry)")
+        return f"(VALUES {', '.join(table_rows)}) as score_bounds(label, geom)"
+
+    @staticmethod
+    def bounds_dict_has_overlaps(bounds: Dict[str, BaseGeometry]) -> bool:
```
maxalbert: See comment above about renaming this, e.g. to `verify_bounds_dict_has_no_overlaps`?

flowmachine/flowmachine/features/subscriber/meaningful_locations.py
```diff
@@ -0,0 +1,259 @@
...
+
+    This is an extension of work by Isaacman et al.[1]_ by Flowminder in collaboration with the World Bank[2]_.
+
+    Parameters
+    ----------
+    clusters : HartiganCluster
+        Per subscriber clusters of towers
+    scores : EventScore
+        Per user, per tower scores based on hour of day and day of week of interactions with the tower
+    labels : LabelEventScore
```
maxalbert: In this class the difference between the `labels` and `label` argument is a little clearer, but probably still worth expanding on it (or giving an example in the docstring).

flowmachine/flowmachine/features/subscriber/meaningful_locations.py
```diff
@@ -0,0 +1,259 @@
...
+        SELECT subscriber, label, cluster, (sum(1) OVER (PARTITION BY subscriber)) as n_clusters FROM 
+            ({self.labelled_subset.get_query()}) clus 
+        GROUP BY subscriber, label, cluster
+        ORDER BY subscriber
+        """
+
+
+class MeaningfulLocationsAggregate(Query):
+    """
+    Aggregates an individual level meaningful location to a spatial unit by assigning
```
maxalbert:
```suggestion
    Aggregates an individual-level meaningful location to a spatial unit by assigning
```

flowmachine/flowmachine/features/subscriber/meaningful_locations.py
```diff
@@ -0,0 +1,259 @@
...
+    subscribers with clusters in that unit to it. For subscribers with more than one cluster,
+    assigns `1/n_clusters` to each spatial unit that the cluster lies in.
+
+    Parameters
+    ----------
+    meaningful_locations : MeaningfulLocations
+        A per-subscriber meaningful locations object to aggregate
+    level : {"admin3", "admin2", "admin1", "grid", "polygon"}, default "admin3"
```
maxalbert: Rename `level` to `aggregation_unit`?

flowmachine/flowmachine/features/subscriber/meaningful_locations.py
```diff
@@ -0,0 +1,259 @@
...
+        Spatial unit to aggregate to
+    column_name : str or list of str, default None
+        Optionally specify a non-default column name or names from the spatial unit table
+    polygon_table : str, default None
+        When using the "polygon" level, you must specify the fully qualified name of a table
```
maxalbert: Substitute "level" with "aggregation unit"?

flowmachine/flowmachine/features/subscriber/meaningful_locations.py
```diff
@@ -0,0 +1,259 @@
...
+        containing polygons.
+    geom_column : str, default "geom"
+        When using the "polygon" level, you must specify the name of column containing geometry
```
maxalbert: Substitute "level" with "aggregation unit"?

flowmachine/flowmachine/features/subscriber/meaningful_locations.py
```diff
@@ -0,0 +1,259 @@
...
+    size : int, default None
+        When using the "grid" level, you must specify the size of the grid to use in KM
```
maxalbert: Substitute "level" with "aggregation unit"?

maxalbert: (Similar for other classes where appropriate)

flowmachine/flowmachine/features/subscriber/meaningful_locations.py
```diff
@@ -0,0 +1,259 @@
...
+
+    @property
+    def column_names(self) -> List[str]:
+        return (
+            ["label"] + get_columns_for_level(self.level, self.column_name) + ["total"]
+        )
+
+    def _make_query(self):
+        agg_query, agg_cols = self.aggregator._geo_augmented_query()
+        level_cols = get_columns_for_level(self.level, self.column_name)
```
maxalbert: Note to self: a _lot_ of this "levels" logic is duplicated and spread all over the place. Will be good to make this the focus of one of the next refactorings to encapsulate it (and esp. allow treating all aggregation units uniformly) so that no other code needs to know about and deal with this logic.

flowmachine/flowmachine/features/subscriber/scores.py
```diff
@@ -112,20 +102,73 @@ class EventScore(Query):
...
+            "friday",
+            "saturday",
+            "sunday",
+        }:
+            raise ValueError(
+                f"Day of week score dictionary must have values for all days. Only got {set(score_dow.keys())}"
+            )
+        if len(score_hour) != 24:
+            raise ValueError(
+                f"Hour of day score dictionary must have values for all hours. Only got {len(score_hour)}"
```
maxalbert: Maybe remove "only" because implies that the user always provides fewer than 24 hours?

flowmachine/tests/conftest.py
```diff
@@ -62,13 +63,15 @@ def pytest_itemcollected(item):
...
-@pytest.fixture
-def skip_datecheck(monkeypatch):
+@pytest.fixture(autouse=True)
+def skip_datecheck(request, monkeypatch):
     """Temporarily patches EventTableSubset so that it thinks any date is
     available, _without_ needing to touch the database. This shaves a little
     time off every `daily_location` creation.
     """
-    monkeypatch.setattr(EventTableSubset, "_check_dates", lambda x: True)
+    run_date_checks = request.node.get_closest_marker("check_available_dates", False)
```
maxalbert: What does this line do? From the docstring it sounds like the purpose of `skip_datecheck` is to unconditionally patch `EventTableSubset`, so what is the purpose of this?

maxalbert: Oh I see, this is needed for the `@pytest.mark.check_available_dates` decorator below. This should be clarified in the docstring then.

flowmachine/flowmachine/features/subscriber/label_event_score.py
maxalbert: One thing that only occurred to me when I looked at the tests (and only because we discussed the algorithm a couple of days a ago) is that these geojson shapes are _not_ your typical geographic locations but that they refer to polygons in the "phase space" of valid scores. This really needs to be clarified somewhere, otherwise it's impossible for users to guess the meaning.

flowmachine/tests/test_label_event_score.py
```diff
@@ -0,0 +1,233 @@
...
+                        ]
+                    ],
+                }
+            },
+        )
+
+
+def test_whether_required_label_relabels(get_dataframe):
+    """
+    Test whether required label relabel the location of subscribers who did not originally have the required label.
```
maxalbert: Hmm, I'm still not sure whether I understand what the meaning/purpose of the required label actually is. Does this mean it unconditionally relabels locations of subscribers to this? Or only previously unassigned locations?

flowmachine/tests/test_label_event_score.py
```diff
@@ -0,0 +1,233 @@
...
+        },
+        required="evening",
+    )
+    df = get_dataframe(ls)
+    assert list(df["label"].unique()) == ["evening"]
+
+
+def test_overlaps_bounds_dict():
+    """
+    Should return False if there are no overlaps in the labels dict.
```
maxalbert: I think I'd prefer for `bounds_dict_has_overlaps` to be renamed to e.g. `verify_...` or `check_...` and either pass silently or raise an exception on failure.

flowmachine/tests/test_label_event_score.py
```diff
@@ -0,0 +1,233 @@
...
+        },
+        ValueError,
+    ),
+]
+
+
+@pytest.mark.parametrize("bad_bound", next(zip(*bad_bounds)))
+def test_constructor_raises_value_error(bad_bound):
+    """
+    Constructor should only raise value errors.
```
maxalbert: Not opposed to this, but out of curiosity, what's the reason we (a) want this and (b) explicitly need to test for it?

flowmachine/tests/test_meaningful_locations.py
```diff
@@ -0,0 +1,299 @@
...
+                label="evening",
+            ),
+            level="NOT_A_LEVEL",
+        )
+
+
+def test_column_names_meaningful_locations_od(
+    exemplar_level_param, get_column_names_from_run
+):
+    """ Test that column_names property matches head(0) for an od matrix between meaningful locations"""
```
maxalbert: Again just a note to self: we have a bunch of tests that test this property for each query. We can probably encapsulate this better and ensure it is automatically valid?

flowmachine/tests/test_meaningful_locations.py
```diff
@@ -0,0 +1,299 @@
...
+        mfl_od = MeaningfulLocationsOD(
+            meaningful_locations_a=mfl, meaningful_locations_b=mfl, level="NOT_A_LEVEL"
+        )
+
+
+def test_meaningful_locations_od_results(get_dataframe):
+    """
+    Test that od on MeaningfulLocations returns expected results and counts clusters per subscriber correctly.
+    """
+    # Because of the nature of the test data, we can't actually test much for most admin levels because
```
maxalbert:
```suggestion
    # FIXME: Because of the nature of the test data, we can't actually test much for most admin levels because
```

flowmachine/tests/test_meaningful_locations.py
```diff
@@ -0,0 +1,299 @@
...
+
+    with pytest.raises(BadLevelError):
+        mfl_od = MeaningfulLocationsOD(
+            meaningful_locations_a=mfl, meaningful_locations_b=mfl, level="NOT_A_LEVEL"
+        )
+
+
+def test_meaningful_locations_od_results(get_dataframe):
+    """
+    Test that od on MeaningfulLocations returns expected results and counts clusters per subscriber correctly.
```
maxalbert:
```suggestion
    Test that OD on MeaningfulLocations returns expected results and counts clusters per subscriber correctly.
```

josh-gree:
Just a suggestion here - do feel free to ignore. 

It may make sense to completely ignore the day of the week score in the api!? It really makes no sense in terms of classifying things as day/night. Only having one dimension along which to do the classification would also remove the need to think about using geometries of any kind to do the classification. 

How it currently stands the decision to use geometries to do the classification in the dow/hod axis is really quite confusing. I see that perhaps it makes sense if you would like to have multiple disjoint areas that are assigned to different classes but I can't really see a case where you would want this!?

Really you want 3 classes night, no idea, possibly day and the best way to do that is just split the hod score into three sections. I would suggest that just providing the ability to specify two numbers (day_bound, night_bound) is the best way to do the classification. Everything less than day_bound is a day location, everything greater than night_bound is a night location and everything in the middle is unknown...my inequalities may be the wrong way round but depends on how you do the hod score anyway...

flowmachine/flowmachine/features/subscriber/meaningful_locations.py
greenape: Gonna skip that for now, because all of the test harness stuff assumes level is the name of the arg.

greenape:
@josh-gree Definitely be worth providing a super (relatively) simple API for this that works as described, but it would sacrifice a lot of flexibility. Basically, feels like taking one decision too many for the end user to me - day/night isn't the only thing this might be used for.

flowmachine/flowmachine/features/subscriber/meaningful_locations.py
maxalbert: Ok 👍

josh-gree:
It may sacrifice flexibility but the question remains whether the flexibility present is what we need!? I have spent many hours staring at the dow/hod "Phase space" the extra dimension adds nothing as far as I am concerned...it just complicates the situation! There is no clear separation into home / work areas and that would be the only reason for having the dow score. 

You could of course make the argument that this should be much more general and that we should allow arbitrary scoring functions applied to the users interactions with each cluster and that then a much more complicated method for cleaving the resulting space into the various classes would be needed...of course I am not suggesting this...

So yeah only having hod score removes flexibility but it simplifies the use of this function greatly. 

Anyway I would ask that you think of a different way of doing the class areas that does not involve using geometries! The original way seemed much more user friendly to me...

```
{
'evening': 
'(score_hour > 0) AND (score_dow > 0.5 OR score_dow < -0.5)' , 
'daytime': 
'(score_hour < 0) AND (score_dow < 0.5 AND score_dow > -0.5)'
}
```

Perhaps allowing for a list of constraints if needed!?

greenape:
I'm averse to forcing people to write SQL, which we can't check is even going to be syntactically correct before execution _and_ requires you to have a bunch of mitigation stuff in place for injection attacks.

I did initially implement the labelling as just accepting upper & lower bounds, but that's obviously waaaaay less flexible than the original SQL were clause approach. So, I've gone with geometries because they're really what you'd be expressing anyway, there's a decent standard for expressing them, checking they're valid, checking that your bounds are exclusive etc.

josh-gree:
Ok then perhaps not raw SQL then I can see that's probably a bad idea via an API. Then perhaps just top left, bottom right? Possibly a list of? In what situation do you want an arbitrary non rectangular region?

josh-gree:
Yeah but I am not suggesting we add anything to the algorithm I am suggesting we make it less complex and thus more usable. Day of the week is not needed - if anybody can come up with a way of splitting the phase space that makes sense to classify home and work then I am all for it. However I don't think it's possible...

flowclient/flowclient/client.py
```diff
@@ -455,6 +455,434 @@ def daily_location(
...
+    """
+    Helper function which constructs a spec for a meaningful locations query
+    to be aggregated.
+    
+    Parameters
+    ----------
+    start_date : str
+        ISO format date that begins the period, e.g. "2016-01-01"
+    stop_date : str
+        ISO format date that begins the period, e.g. "2016-01-07"
```
jch1g10:
```suggestion
        ISO format date that ends the period, e.g. "2016-01-07"
```

flowclient/flowclient/client.py
```diff
@@ -455,6 +455,434 @@ def daily_location(
...
+        tower.
+    tower_hour_of_day_scores : list of float
+        A length 24 list containing numerical scores in the range [-1.0, +1.0], where the first entry is midnight.
+        Each of a subscriber's interactions with a tower is given a score for the hour of the day it took place in. For
+        example, if the first entry of this list was 1, and all others were zero, each interaction the subscriber had
+        that used a tower at midnight would receive a score of 1. If the subscriber used a particular tower twice, once
+        at midnight, and once at noon, the final hour score for that tower would be 0.5.
+    tower_cluster_radius : float
+        When constructing clusters, towers will be considered for inclusion in a cluster only if they are within this
+        number of KM from the current cluster centroid. Hence, large values here will tend to produce clusters containing
```
jch1g10:
```suggestion
        number of km from the current cluster centroid. Hence, large values here will tend to produce clusters containing
```

flowclient/flowclient/client.py
```diff
@@ -455,6 +455,434 @@ def daily_location(
...
+) -> dict:
+    """
+    Return an origin-destination matrix between two meaningful locations at some unit of spatial aggregation.
+    Generates clusters of towers used by subscribers' over the given time period, scores the clusters based on the
+    subscribers' usage patterns over hours of the day and days of the week. Each subscriber then has a number of
+    clusters, each of which has a score for hourly usage, and day of week usage. These clusters are then labelled
+    based on whether they overlap with the regions of that space defined in the `labels` parameter.
+
+    Once the clusters are labelled, those clusters which have either `label_a` or `label_b` are extracted, and then
+    a count of of number of subscribers who move between the labels is returned, after aggregating spatially.
```
jch1g10:
```suggestion
    a count of number of subscribers who move between the labels is returned, after aggregating spatially.
```

PR #282, opened 2019-01-24
More threads

### LINKED ISSUE #280: Allow more threads than db connections
At present, the thread pool that FlowMachine uses for running queries is limited to 1 thread per db connection. This was historically because we had issues with timeouts when trying to get a connection from the db connection pool, and to avoid clobbering FlowDB with millions of simultaneous queries.

This isn't an issue anymore, so we can have _way_ more threads than db connections. Why would we want more threads? Because the number of threads is going to limit how many queries-composed-of-other-queries we can manage trying to cache at once. For example, while working on #279 I was investigating whether it would be doable to call `store` inside `construct_query_object`. Turns out, that if you do that for a deeply nested query you'll hit a scenario where the outermost `store` won't manage to start running in a thread, and hence acquire a lock in Redis, so it'll inaccurately respond immediately to a call to `get_query` with the output of `_make_query`.

greenape:
Closes #280

### I have:

### Description

Small tweak to FlowMachine, which allows it to manage considerably more queries at the same time. Stops the size of the thread pool used for managing queries being the same as the number of connections allowed to FlowDB, and makes it to default to the threadpoolexecutor default, optionally specifiable via param to `connect`, or the `THREAD_POOL_SIZE` env var.


greenape:
Way back when, we a. set a timeout for getting a connection from the pool, and b. checked the connection out before successfully acquiring a lock, which made it pretty trivial to exhaust the pool and start getting errors.

This is alleviating - the Redis lock issue is likely to still be an issue under much heavier load than we’ve encountered though I think.

flowmachine/tests/conftest.py
```diff
@@ -119,8 +119,8 @@ def clean_env(monkeypatch):
     monkeypatch.delenv("DB_PW", raising=False)
     monkeypatch.delenv("DB_HOST", raising=False)
     monkeypatch.delenv("DB_NAME", raising=False)
-    monkeypatch.delenv("POOL_SIZE", raising=False)
-    monkeypatch.delenv("POOL_OVERFLOW", raising=False)
+    monkeypatch.delenv("DB_CONNECTION_POOL_SIZE", raising=False)
+    monkeypatch.delenv("DB_CONNECTION_POOL_OVERFLOW", raising=False)
```
maxalbert: Actually, the env var `THREAD_POOL_SIZE` needs testing too.

PR #284, opened 2019-01-25
Event count

gzagatti:
### I have:

### Description

This PR adds the `SubscriberEventCount` method that allows one to count the number of events per subscriber. It allows user to select the desired direction of the events, the desired tables and all the options generally available to `EventsTableUnion`.

If a count of directed event is chosen, the query will only look at events which have directed information skiping the rest. A warning is raised in those cases to tell the user that only a partial number of tables was selected.


flowmachine/flowmachine/features/subscriber/subscriber_event_count.py
```diff
@@ -0,0 +1,136 @@
...
+import warnings
+
+from ...core import Table
+from ..utilities.sets import EventsTablesUnion
+from .metaclasses import SubscriberFeature
+
+
+class SubscriberEventCount(SubscriberFeature):
+    """
+    This class returns the event count pr subscriber within the period,
```
maxalbert:
```suggestion
    This class returns the event count per subscriber within the period,
```

flowmachine/flowmachine/features/subscriber/subscriber_event_count.py
```diff
@@ -0,0 +1,136 @@
...
+    2  0Ze1l70j0LNgyY4w           57
+    3  Nnlqka1oevEMvVrm           63
+    4  4dqenN2oQZExwEK2           59
+                    ...          ...
+    """
+
+    def __init__(
+        self,
+        start,
+        stop,
```
maxalbert: Just for safety, I suggest to make the names of kwargs non-optional:
```suggestion
        stop,
        *,
```

flowmachine/flowmachine/features/subscriber/subscriber_event_count.py
```diff
@@ -0,0 +1,136 @@
...
+        subscriber_identifier="msisdn",
+        direction="both",
+        hours="all",
+        subscriber_subset=None,
+        tables="all",
+        **kwargs,
```
maxalbert: Please no `**kwargs` any more. We used to do this pretty much everywhere in the past but have started to make all kwargs explicit (see #257), so that we can soon bundle them up and encapsulate them in more meaningful and generally smarter domain objects.

Given that `**kwargs` is only passed on to `EventsTablesUnion` and all arguments for this are already provided below, we can simply remove `**kwargs` in both places.

flowmachine/flowmachine/features/subscriber/subscriber_event_count.py
```diff
@@ -0,0 +1,136 @@
...
+    ):
+        self.start = start
+        self.stop = stop
+        self.subscriber_identifier = subscriber_identifier
+        self.direction = direction
+        if direction not in {"in", "out", "both"}:
+            raise ValueError("{} is not a valid direction.".format(self.direction))
+
+        try:
```
maxalbert: I think this try...except block can simply be changed to `self.hours = hours` (since `hours` is passed in explicitly).

flowmachine/flowmachine/features/subscriber/subscriber_event_count.py
```diff
@@ -0,0 +1,136 @@
...
+        super().__init__()
+
+    def _has_outgoing(self, table):
+
+        return "outgoing" in Table(table).column_names
+
+    def _make_query(self):
+        where_clause = ""
+        if self.direction != "both":
+            where_clause = "WHERE {}outgoing".format(
```
maxalbert: How about the following?
```python
where_clause = f"WHERE outgoing IS {'TRUE' if self.direction == 'out' else 'FALSE}"
```

flowmachine/flowmachine/features/subscriber/subscriber_event_count.py
```diff
@@ -0,0 +1,136 @@
...
+            hours=hours,
+            subscriber_identifier=subscriber_identifier,
+            subscriber_subset=subscriber_subset,
+            **kwargs,
+        )
+        super().__init__()
+
+    def _has_outgoing(self, table):
+
+        return "outgoing" in Table(table).column_names
```
maxalbert: Hmm, I'm not strictly opposed to having this in a helper method but it's quite succinct so might actually be more readable if it is inlined in L96 above (esp. since it doesn't actually use `self`, which confused me a little at first when I looked up the definition).

flowmachine/tests/test_subscriber_event_count.py
```diff
@@ -0,0 +1,103 @@
...
+    df = get_dataframe(query).set_index("subscriber")
+    assert df.loc["E0LZAa7AyNd34Djq"].event_count == 8
+
+    query = SubscriberEventCount(
+        "2016-01-01", "2016-01-08", direction="both", tables="events.mds"
+    )
+    df = get_dataframe(query).set_index("subscriber")
+    assert df.loc["E0LZAa7AyNd34Djq"].event_count == 8
+
+    query = SubscriberEventCount("2016-01-01", "2016-01-08", direction="out")
```
maxalbert: Just for completeness maybe also add a case for `direction="in"`?

flowmachine/flowmachine/features/subscriber/subscriber_event_count.py
```diff
@@ -0,0 +1,136 @@
...
+            self.hours = "all"
+
+        if self.direction == "both":
+            column_list = [self.subscriber_identifier]
+            self.tables = tables
+        else:
+
+            column_list = [self.subscriber_identifier, "outgoing"]
+
+            if isinstance(tables, str) and tables.lower() == "all":
```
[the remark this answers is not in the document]
maxalbert: Also, what happens if _no_ tables have an `outgoing` column? Will this still raise only a warning? Imho this case should definitely be an error, and then it feels a little inconsistent if other cases only raise a warning.

Is this the case that's covered by the second case in `test_directed_count_undirected_tables_raises` below? If so, it also seems strange that it would raise `MissingDateError` rather than, say, `MissingDirectionColumnError`.

So all in all it seems cleaner to always raise an error, which also keeps the internal logic simpler.

flowmachine/flowmachine/features/subscriber/subscriber_event_count.py
gzagatti: And I was wondering what this `*` was for which is everywhere in the codebase. Thanks for illuminating my day.

flowmachine/flowmachine/features/subscriber/subscriber_event_count.py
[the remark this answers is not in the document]
gzagatti: I implemented the suggestions, now a `MissingDirectionColumnError` will be raised when such column is missing from any of the tables requested.

flowmachine/flowmachine/features/subscriber/subscriber_event_count.py
```diff
@@ -0,0 +1,126 @@
...
+
+        if direction not in {"in", "out", "both"}:
+            raise ValueError("{} is not a valid direction.".format(self.direction))
+
+        if self.direction == "both":
+            column_list = [self.subscriber_identifier]
+            self.tables = tables
+        else:
+            column_list = [self.subscriber_identifier, "outgoing"]
+            self.tables = self._parse_tables_with_direction(tables)
```
maxalbert: Very nice, this reads much clearer now that you've moved the parsing code into a helper method.

Tiny suggestion: the name `_parse_tables_with_direction` seems to imply that it only parses the tables that have a direction and omits the others, when in fact it raises an error in this case (as it should). How about renaming it to `_parse_tables_ensuring_direction_is_present` or something similar?

maxalbert:
@gzagatti Thanks for fixing the recent CI failures. Nearly there! Only the file `/home/circleci/project/flowmachine/tests/test_subscriber_event_count.py` needs to be reformatted with black to make the linting test pass.

May I ask why you decided to rename the class from `SubscriberEventCount` to `EventCount`? We may need to re-think the naming conventions in general, but for the time being the former sounds a little more explicit to me, so why not keep it?

(Also, if we do decide that renaming it is better then we should also tweak the filename of the .py and test file; and there is still one instance of the old class name in a docstring.)

gzagatti:
I changed the name of the class because the majority of subscriber features do not have the `Subscriber` prefix. For instance, we have `RadiusOfGyration` and not `SubscriberRadiusOfGyration`. Connected to that, I found it redundant the following `from features.subscriber import SubscriberEventCount`. `from features.subscriber import EventCount` reads better.

But I agree with you we need a re-think of naming conventions.

PR #303, opened 2019-01-29
Entropy

### LINKED ISSUE #350: Entropy covariates
Entropy is a useful metric, would be good to be able to calculate entropy e.g. of the period/time when events take place, of locations visited and of counterparties contacted.

Entropy can be calculated as: `-1 * SUM( relative_freq * LN( relative_freq ) )`

gzagatti:
Closes #350.

### I have:

### Description

This PR introduces a number of entropy calculations including periodic, location and contact entropy. It is based on a number of pre-existing queries, adding a new layer of aggregation which is the entropy.

This PR also fixes some issues with `ContactBalance` which arose from using this class in this PR. The issues include fixing where clauses and removing the filtering for event ID since this cannot be guaranteed in the case of SMS events.


flowmachine/flowmachine/features/subscriber/contact_balance.py
```diff
@@ -69,59 +67,67 @@ def __init__(
         stop,
         *,
         hours="all",
-        table="all",
+        tables="all",
```
maxalbert: Hmm, I'm in two minds about renaming this argument from `table` to `tables`. I think I've concluded that it's fine to change it (because we'll need to do cleanup on this soon anyway), but below are a few thoughts just for the general discussion & awareness.

In principle I agree that `tables` is a better name (although we also still allow a single string, in which case using the plural in `tables="<some_table_name>"` is also slightly confusing).

I'm not super happy with changing the name just in some clcasses because it introduces inconsistencies across the codebase that we should aim to eliminate. Before this PR almost every other class calls it `table`, while afterwards there is roughly a 50-50 mix of `table` and `tables`.

But as I said, I'm fine with it for now because the real underlying problems are that (i) we shouldn't really need to specify this in every instantiation of every class (I think?), and also (ii) this is one of the many examples (with other being aggregation units, subsets, etc.) where [primitive obsession](http://wiki.c2.com/?PrimitiveObsession) makes our lives hard and duplicates logic and arguments all over the codebase because we're using a string/list/... to capture a concept that needs richer logic and should really be captured in a separate class that can deal with that logic in a single place.

Anyway, long rambling but the bottom line is: no need to change it in this PR but let's ensure that subsequent refactorings ensure this is consistent again across the codebase and we have a more cohesive way of dealing with the underlying logic (the latter is mostly a note to self).

maxalbert: However, if you're renaming the argument then it needs to be `tables` in the docstring above, too. (I can't seem to add a comment to that line directly?)

flowmachine/flowmachine/features/subscriber/contact_balance.py
```diff
@@ -69,59 +67,67 @@ def __init__(
...
         self.direction = direction
+        self.subscriber_identifier = subscriber_identifier
         self.exclude_self_calls = exclude_self_calls
 
-        if self.direction not in ("both", "in", "out"):
-            raise ValueError("Unidentified direction: {}".format(self.direction))
+        if self.direction == "both":
+            column_list = [self.subscriber_identifier, "msisdn_counterpart"]
+            self.tables = tables
+        else:
```
maxalbert: Why was the check removed that `direction` needs to be one of `("both", "in", "out")`? With the new logic I could be passing arbitrary values for `direction`, no?

flowmachine/flowmachine/features/subscriber/contact_balance.py
```diff
@@ -69,59 +67,67 @@ def __init__(
...
-                """
-            AND outgoing = FALSE
-            """
-            )
-        elif self.direction == "out":
-            filters.append(
-                """
-            AND outgoing = TRUE
-            """
+                f"outgoing IS {'TRUE' if self.direction == 'out' else 'FALSE'}"
```
maxalbert: See above, I'm not super keen on unconditional `else` clauses that act as a catch-all when we are expecting a particular value (here: `self.direction == 'in'`). If the code in `__init__` already did the check that `direction` was only one of the three expected values than this would be kind-of ok but here it could be an arbitrary value and would land in the `else` branch.

Really what this is hinting at is again that `direction` should perhaps not be a string but a slightly smarter object (or an enum). But I'd leave that for another PR. For the time being it should be ok to just check valid values in `__init__` (see comment above).

flowmachine/flowmachine/features/subscriber/contact_balance.py
```diff
@@ -138,21 +144,17 @@ def _make_query(self):
             SELECT
                 subscriber,
                 count(*) AS events
-            FROM (
-              SELECT DISTINCT U.subscriber,
```
maxalbert: Just to check my understanding: how does removing the `DISTINCT` here (and below) change the logic what's in the result?

flowmachine/flowmachine/features/subscriber/entropy.py
```diff
@@ -0,0 +1,391 @@
...
+
+        return f"""
+        SELECT
+            subscriber,
+            -1 * SUM( relative_freq * LN( relative_freq ) ) AS entropy
+        FROM ({self._relative_freq_query}) u
+        GROUP BY subscriber
+        """
+
+    @property
```
maxalbert: Suggest to add an `@abstractproperty` here, which forces any subclass to implement it (and which is detected during the parsing stage, otherwise you'll actually have to run the query to hit the NotImplementedError).
```suggestion
    @property
    @abstractmethod
```

flowmachine/flowmachine/features/subscriber/entropy.py
```diff
@@ -0,0 +1,391 @@
...
+    events regularly occur at a certain time of day, say at 9:00 and 18:00 then
+    this user will have a low period entropy.
+
+    Parameters
+    ----------
+    start, stop : str
+         iso-format start and stop datetimes
+    phase : {"century", "day", "decade", "dow", "doy", "epoch", "hour",
+            "isodow", "isoyear", "microseconds", "millennium", "milliseconds",
+            "minute", "month", "quarter", "second", "timezone", "timezone_hour",
```
maxalbert: It wasn't clear to me what `timezone_hour` was until I looked it up (and my initial guess was wrong). It actually refers to "the hour component of the time zone offset". This is always constant, so the entropy will always be zero. How about removing `timezone_hour`, `timezone_minute` and `timezone` from the allowed values?

maxalbert: Might also be useful to provide a link to the Postgres manual so that people have a reference for the supported values?

flowmachine/flowmachine/features/subscriber/entropy.py
```diff
@@ -0,0 +1,391 @@
...
+        FROM ({self._absolute_freq_query}) u
+        """
+
+
+class PeriodicEntropy(BaseEntropy):
+    """
+    Calculates the recurrence period entropy for events, that is the entropy
+    associated with the period in which events take place. For instance, if
+    events regularly occur at a certain time of day, say at 9:00 and 18:00 then
+    this user will have a low period entropy.
```
maxalbert: Can we please include a definition/equation in the docstring, or alternatively a reference to a definition (or both)? Entropy is such an overloaded concept with many slightly varying definitions in different contexts that it will be good for the user to know how we define it here.

flowmachine/flowmachine/features/subscriber/entropy.py
```diff
@@ -0,0 +1,391 @@
...
+        GROUP BY subscriber, EXTRACT( {self.phase} FROM datetime )
+        HAVING COUNT(*) > 0
+        """
+
+
+class LocationEntropy(BaseEntropy):
+    """
+    Calculates the entropy of locations visited. For instance, if an individual
+    regularly makes her/his calls from certain location then this user will
+    have a low location entropy.
```
maxalbert: See above: please add definition/reference.

flowmachine/flowmachine/features/subscriber/entropy.py
```diff
@@ -0,0 +1,391 @@
...
+        SELECT subscriber, COUNT(*) AS absolute_freq FROM
+        ({self.subscriber_locations.get_query()}) u
+        GROUP BY subscriber, {self.location_cols}
+        HAVING COUNT(*) > 0
+        """
+
+
+class ContactEntropy(BaseEntropy):
+    """
+    Calculates the entropy of locations visited. For instance, if an individual
```
maxalbert: Copy & paste error ("locations visited"). What exactly does ContactEntropy measure?

As above, please also add a definition/reference.

flowmachine/tests/test_subscriber_entropy.py
```diff
@@ -0,0 +1,93 @@
...
+
+def test_subscriber_location_entropy(get_dataframe):
+    """
+    Test some hand picked periods and tables.
+    """
+    query = LocationEntropy("2016-01-01", "2016-01-08")
+    df = get_dataframe(query).set_index("subscriber")
+    assert df.loc["0DB8zw67E9mZAPK2"].entropy == pytest.approx(2.996_587)
+
+    query = LocationEntropy("2016-01-01", "2016-01-08", level="admin1")
```
maxalbert: Not a big deal, but could be useful to change the period in the second calculation (as you did above), just to have a bit more variety. (Also in the test below).

flowmachine/flowmachine/features/subscriber/contact_balance.py
gzagatti: I agree with all you mentioned above. A common pattern I have observed when creating new covariates is to basically copy the  `__init__` method from other covariates over and over again with small modifications between them (eg. some methods allow direction, others not). This creates inconsistencies and is rather tedious.  Also a lot of unit testing has to do with testing different combinations of parameters which are passed to `EventTablesUnion` the same way, which is a bit time consuming and redundant.

gzagatti: I changed the docstring to `tables`, btw.

flowmachine/flowmachine/features/subscriber/contact_balance.py
gzagatti: The `DISTINCT` was here to ensure we did not double count events. We used distinct on subscriber, counterpart and id, since we only want to count this triplet once. Since we removed the id from this query, if we use `DISTINCT` we will basically select distinct pairs of subscriber and counterparts which will reduce all the events between both to a single line precluding us from counting the events which we want to count in the first place.

gzagatti:
@maxalbert, thanks for the review. I have addressed all your comments. It is ready to merge if you approve all the modifications.

flowmachine/flowmachine/features/subscriber/entropy.py
```diff
@@ -30,12 +30,12 @@ def _make_query(self):
         GROUP BY subscriber
         """
 
-    @property
+    @abstractproperty
```
maxalbert: Minor thing, but `abstractproperty` is deprecated since Python 3.3. They recommend to use
```
    @property
    @abstractmethod
```
instead (see https://docs.python.org/3/library/abc.html#abc.abstractproperty)

flowmachine/flowmachine/features/subscriber/contact_balance.py
```diff
@@ -81,10 +81,13 @@ def __init__(
         self.subscriber_identifier = subscriber_identifier
         self.exclude_self_calls = exclude_self_calls
 
-        if self.direction == "both":
+        if self.direction not in ("both", "in", "out"):
+            raise ValueError("Unidentified direction: {}".format(self.direction))
```
maxalbert: Thanks for adding the check for valid values of `direction` back in. Just as a minor suggestion, I'd add an explicit `else` clause below and raise the ValueError in there.

in this particular case the logic is pretty obvious, but we have/had other places in the code where explicit `else` clauses were missing and there is no way to tell whether this is intentional or the logic is flawed (and in some cases it was actually a bug). So I think we should make it a habit to always add an `else` clause - even if it just contains `pass` and a comment of why the "else" case doesn't need to be explicitly treated. This makes it much easier for the reader to verify the logic at a glance.

Again, not really an issue here, just as a general comment for our general coding habits.

maxalbert:
Awesome, thanks for the changed @gzagatti. Looks good now! 👍 I added a couple of minor tweaks related to my inline comments (just for faster turnaround), will merge once the the tests pass.

maxalbert:
Oh, and I also added an issue related to this added feature (#350). Could you add new issues for your other PRs too? They can be brief, but it is good to be able to associate each pull request with a Github issue so that the work and progress is documented.

PR #320, opened 2019-01-31
Distance counterparts

### LINKED ISSUE #359: Implement metrics for stats about distance to counterparts
Feature request for a covariate which calculates metrics (average, standard deviation, etc) of the distance between event initiator and counterparts.

gzagatti:
Closes #359

### I have:

### Description

This PR implements another covariate. This covariate calculates metrics (average, standard deviation, etc) of the distance between event initiator and counterparts.

This PR also adds a utility for checking whether certain columns are available in event type tables before processing. This was required because we do not know which event types the user will request and event types do not contain the same column in all cases.


flowmachine/flowmachine/features/subscriber/distance_counterparts.py
```diff
@@ -0,0 +1,151 @@
...
+
+    Parameters
+    ----------
+    start, stop : str
+         iso-format start and stop datetimes
+    hours : 2-tuple of floats, default 'all'
+        Restrict the analysis to only a certain set
+        of hours within each day.
+    tables: str, default 'all'.
+        The table must have a `msisdn_counterpart` column.
```
maxalbert: I just realised that we allow the `subscriber_identifier` to be either `msisdn` or `imei`, but from this comment it seems that the counterpart is always identified via `msisdn`? There doesn't seem to be an easy fix given the current table structure, so this needs to be addressed separately. It probably applies to other metrics as well which are calculated from two parties.

flowmachine/flowmachine/features/subscriber/distance_counterparts.py
```diff
@@ -0,0 +1,151 @@
...
+        ).get_query()
+
+        self.distance_matrix = DistanceMatrix()
+
+        super().__init__()
+
+    def _make_query(self):
+
+        filters = []
+        if self.direction != "both":
```
maxalbert: A similar comment applies here as in the other PR: would be good to check that direction actually has a valid value. However, this really should be solved more properly so I've opened #357 for this.

flowmachine/flowmachine/features/subscriber/distance_counterparts.py
```diff
@@ -0,0 +1,151 @@
...
+            filters.append(
+                f"A.outgoing IS {'TRUE' if self.direction == 'out' else 'FALSE'}"
+            )
+        if self.exclude_self_calls:
+            filters.append("A.subscriber != A.msisdn_counterpart")
```
maxalbert: See above, this logic doesn't really make sense if the subscriber_identifier is `imei`, but we can't solve that in this PR.

flowmachine/flowmachine/features/subscriber/distance_counterparts.py
```diff
@@ -0,0 +1,151 @@
...
+                )
+            )
+
+        column_list = ["msisdn", "msisdn_counterpart", "id", "location_id", "outgoing"]
+        self.tables = parse_tables_ensuring_columns(
+            self.connection, tables, column_list
+        )
+
+        # EventsTablesUnion will only subset on the subscriber identifier,
+        # which means that we need to query for a unioned table twice.
```
maxalbert: I haven't had time to dig into this in detail, but my hunch is that the double join between the two unioned tables (in `_make_query` below) is likely to be very slow. The test currently takes ~12 seconds to run, and this is for the small test data! Have you had a chance to test this on larger data, or do you see a way to improve this? It's not a blocker if not, but we should leave a note somewhere to indicate the performance, and revisit it later.

flowmachine/tests/test_subscriber_distance_counterparts.py
```diff
@@ -0,0 +1,118 @@
...
+
+
+@pytest.fixture()
+def distance_matrix(get_dataframe):
+    """ Calculates the distance matrix between cells. """
+    return get_dataframe(DistanceMatrix())
+
+
+@pytest.fixture()
+def distance_counterparts(get_dataframe, distance_matrix):
```
maxalbert: This is a reference implementation using pandas, right? Can we make this clear in the fixture name, and also explain it in the docstring? E.g. name the fixture `distance_counterparts_expected` or `distance_counterparts_reference` or similar?

flowmachine/flowmachine/features/subscriber/distance_counterparts.py
gzagatti: The main problem here is that the MNO can only really tell the MSISDN for which the call/SMS is being directed to. Unless the subscriber is also on its network, it cannot tell anything about the IMEI of the receiving phone (at least that's my hunch). This becomes even more problematic with SMS, where the MNO sometimes can't even tell the counterpart details of an SMS transaction because they are not synchronous.

gzagatti: In any case, do you want me to address this issue in this PR in any particular way? For instance, by changing the docstring, etc.

flowmachine/flowmachine/features/subscriber/distance_counterparts.py
gzagatti: I have at least an extra condition that `self.subscriber_identifier in {"msisdn"}` to the if clause.

flowmachine/flowmachine/features/subscriber/distance_counterparts.py
gzagatti: Haven't done this test, but I agree that it should slow things down given that it is filtering the `events` table twice. Do you think it would perhaps more sense to simple join to the unfiltered `event` table using `msisdn_counterpart` as the key?

Ideally, I would like to see the `subscriber_subset` also collecting events for which the `msisdn_counterpart` is present. At the moment, it only allows for one column to be used for filtering.

flowmachine/flowmachine/features/subscriber/distance_counterparts.py
maxalbert: No, I think this is fine because it clearly states the expectation on the input table. (My comment above was more of a general reflection that this needs to be addressed in flowmachine, but #358 tracks this now).

flowmachine/flowmachine/features/subscriber/contact_balance.py
```diff
@@ -102,36 +106,14 @@ def __init__(
...
-
-        return parsed_tables
-
     def _make_query(self):
 
         filters = []
         if self.direction != "both":
             filters.append(
-                f"outgoing IS {'TRUE' if self.direction == 'out' else 'FALSE'}"
+                f"outgoing = {'TRUE' if self.direction == 'out' else 'FALSE'}"
```
maxalbert: FYI, the issue with the failing boolean clause due to the `WHERE ... IS TRUE` statement has been fixed in pglast (see [here](https://github.com/lelit/pglast/issues/11)). So we could revert this to your original version, but either is fine, it doesn't really matter.

flowmachine/flowmachine/features/subscriber/contact_balance.py
```diff
@@ -81,15 +81,19 @@ def __init__(
         self.subscriber_identifier = subscriber_identifier
         self.exclude_self_calls = exclude_self_calls
 
-        if self.direction in {"both"}:
+        if self.direction == "both":
             column_list = [self.subscriber_identifier, "msisdn_counterpart"]
-            self.tables = tables
+            self.tables = verify_columns_exist_in_all_tables(
```
maxalbert: This assignment to `self.tables` (and the verification of the columns) is duplicate, because it also happens in lines 94-95 below. I'd remove this statement here and only set the `column_list` (analogous to the `elif` branch in line 90 below).

Then we can simplify the helper function `verify_columns_exist_in_all_tables` by removing the return statement as well, which isn't needed for the verification that the columns are present.

flowmachine/flowmachine/utils/utils.py
```diff
@@ -240,6 +242,54 @@ def proj4string(conn, crs=None):
...
+    ----------
+    conn : Connection
+        FlowMachine db connection to use to get the subscriber_tables.
+    tables : str or list of strings, default 'all'
+        Can be a sting of a single table (with the schema) or a list of these.
+        The keyword all is to select all subscriber tables
+    columns : str or list
+        A string or list of strings with the column names that must be present.
+
+    Returns
```
maxalbert: I would suggest to not return anything from this function, because the return value will be exactly equal to the input `tables` (unless one of the columns is missing, in which case an error will be thrown).

flowmachine/flowmachine/utils/utils.py
```diff
@@ -240,6 +242,54 @@ def proj4string(conn, crs=None):
...
+        tables = [f"events.{t}" for t in conn.subscriber_tables]
+    elif type(tables) is str:
+        tables = [tables]
+    else:
+        tables = tables
+
+    if isinstance(columns, str):
+        columns = [columns]
+
+    parsed_tables = []
```
maxalbert: If we remove the return value then the logic in lines 275-290 can be significantly simplified because you don't need to keep track of parsed tables:
```
for t in tables:
    for c in columns:
        if c not in flowmachine.core.Table(t).column_names:
            raise MissingColumnsError(tables_lacking_columns, columns)
```

flowmachine/flowmachine/features/subscriber/distance_counterparts.py
maxalbert: Yeah, might make sense to test alternative methods to calculate the result, to see if any speed-ups can be achieved. I'd say it's not needed for this PR, but we should look into it.

maxalbert:
Great, nearly ready. :) Just a small change to simplify `verify_columns_exist_in_all_tables` (see inline comment), then this is ready to go.

PR #325, opened 2019-02-01
Hide token generation in docs notebook

### LINKED ISSUE #308: Example usage docs page contains misleading details of token generation
The docs contain [example FlowClient usage](https://flowminder.github.io/FlowKit/analyst/example_usage/) which is generated from a Jupyter notebook.  This notebook needs to automatically generate a token, which it does using `integration_tests` utils, and the code to do this appears in the ["Create an access token"](https://flowminder.github.io/FlowKit/analyst/example_usage/#create-an-access-token) section of the corresponding docs page.

This is misleading for a user, who should use FlowAuth to generate a token, so it would be helpful to hide these cells in the markdown output from the notebook.

[This link](https://stackoverflow.com/questions/31517194/how-to-hide-one-specific-cell-input-or-output-in-ipython-notebook) contains information which might help.

jch1g10:
Closes #308

### I have:

### Description

This moves the token generation from `example_usage.ipynb` into the preamble python script, so that it doesn't appear in the markdown output in the docs. The preamble script is renamed from `pandas_to_markdown.py` to `notebook_preamble.py` to reflect that it is doing more now.


docs/notebook_preamble.py
```diff
@@ -0,0 +1,68 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
```
maxalbert: Can we add a brief comment here what the purpose of this file is, and how/where it gets included in the docs? I know this is obvious in this PR, but if someone looks at it separately it may not be clear why it contains various disparate functions and how it gets integrated with the notebook(s) (i.e. via a setting in `mkdocs.yml`).

docs/source/analyst/example_usage.ipynb
```diff
@@ -21,87 +21,15 @@
...
-   "metadata": {},
-   "outputs": [],
-   "source": [
-    "TOKEN = make_token(\n",
-    "    username=\"testuser\",\n",
-    "    secret_key=\"secret\",\n",
-    "    lifetime=timedelta(days=1),\n",
-    "    claims=claims,\n",
-    ")"
+    "Generate an API access token using [FlowAuth](../../flowauth)."
```
maxalbert: This is likely correct, but I had issues with relative links before where sometimes one more level was needed than I thought. Have you been able to check if it works locally? If not we should double-check once the docs are built via CI.

docs/notebook_preamble.py
```diff
@@ -0,0 +1,68 @@
...
+get_ipython().display_formatter.formatters["text/html"].for_type(pd.DataFrame, to_md)
+get_ipython().display_formatter.formatters["text/markdown"].for_type(dict, format_dict)
+get_ipython().display_formatter.formatters["text/markdown"].for_type(
+    str, lambda x: f">`{x}`"
+)
+
+
+# Create an API access token
+
+import os
```
maxalbert: I think I'd be in favour of keeping all imports at the top (following PEP-8).

docs/source/analyst/example_usage.ipynb
maxalbert: That's a good point. It looks like the file `docs/source/flowauth.md` is still present but not actually used in the deployed docs, and at a first glance the info is indeed all in the "Installing FlowAuth" section so this file should probably be removed. I assume eventually we will want separate docs pages for each individual component again (beyond just installation) but for now probably best to link to the page you mentioned.

jch1g10: Looks like we have a similar situation for the README in each component directory - they all link to pages which are still present in the docs, but unreachable from the main docs pages.

PR #326, opened 2019-02-01
Interevent period

gzagatti:
### I have:

### Description

Implements interevent period statistics for subscribers as another covariates. Things such as average duration between calls, etc.


gzagatti:
`column_names` added.

flowmachine/flowmachine/features/subscriber/interevent_period.py
```diff
@@ -0,0 +1,146 @@
...
+
+# -*- coding: utf-8 -*-
+"""
+Intervent period statistics, such as the average and standard deviation of the
+duration between calls.
+"""
+import warnings
+from typing import List
+
+from ...utils.utils import parse_tables_ensuring_columns
```
greenape: This doesn't seem to exist?

You should get an error bubbled up from `Table` if one of the events tables doesn't have the columns you're expecting.

maxalbert: I think this was renamed to (or merged with) `verify_columns_exist_in_all_tables` in a previous PR.

gzagatti: Just FIY, in a previous PR we talked about the benefits of being explicit about the nonexistence of columns rather than relying on the SQL command failing. 

That is the reason why we are using `verify_columns_exist_in_all_tables`.

I have corrected the wrong function call as well.

greenape: Right, but it won't _be_ an SQL fail. You effectively ask for a set of tables, and each table must have columns thus, and `Table` will throw you an exception if one of the columns doesn't exist in the table - https://github.com/Flowminder/FlowKit/blob/a5ee3b15d554c97f5f089a542b65e429c44d7289/flowmachine/flowmachine/core/table.py#L108

Appreciate that this func isn't something introduced in this PR, but I don't get the duplication of functionality.

maxalbert: @greenape What exactly is your suggestion to do here?

flowmachine/flowmachine/features/subscriber/interevent_period.py
```diff
@@ -0,0 +1,146 @@
...
+            statistic_clause = (
+                f"MAKE_INTERVAL(secs => {self.statistic}(EXTRACT(EPOCH FROM delta)))"
+            )
+        else:
+            statistic_clause = f"{self.statistic}(delta)"
+
+        sql = f"""
+        SELECT
+            subscriber,
+            {statistic_clause} AS interevent_period_{self.statistic}
```
greenape: Per #421 can we drop the suffix on this column?

gzagatti: Ok, what would be the best to call it simply `interevent_period` or `value`? I didn't quite get from #421 what's our consensus for column names.

flowmachine/flowmachine/features/subscriber/interevent_period.py
gzagatti: I'm happy to ditch the extra function as I agree it is not a good idea to duplicate the functionality. But it is easier to do in another PR since other classes are already using this extra functionality.

flowmachine/flowmachine/features/subscriber/interevent_period.py
maxalbert: I think eventually we want to have column names that are the same across all queries of a particular type (e.g. the two column names `subscriber` and `value` for queries that return a subscriber-based metric). This would eliminate case distinctions (and the need to pass column names as parameters) from any code that consumes a subscriber-based metric because it doesn't need to worry which column names the query returns. So personally I'd prefer using `value` here.

gzagatti: @greenape, do you agree with max? Can I switch the column name to `value`?

flowmachine/flowmachine/features/subscriber/interevent_period.py
```diff
@@ -0,0 +1,143 @@
...
+
+        if self.direction in {"both"}:
+            column_list = [self.subscriber_identifier, "datetime"]
+        elif self.direction in {"in", "out"}:
+            column_list = [self.subscriber_identifier, "datetime", "outgoing"]
+        else:
+            raise ValueError("{} is not a valid direction.".format(self.direction))
+
+        self.statistic = statistic.lower()
+        if self.statistic not in valid_stats:
```
greenape: Need a test for this pls :)

flowmachine/flowmachine/features/subscriber/interevent_period.py
```diff
@@ -0,0 +1,143 @@
...
+
+        if direction not in {"in", "out", "both"}:
+            raise ValueError("{} is not a valid direction.".format(self.direction))
+        self.direction = direction
+
+        if self.direction in {"both"}:
+            column_list = [self.subscriber_identifier, "datetime"]
+        elif self.direction in {"in", "out"}:
+            column_list = [self.subscriber_identifier, "datetime", "outgoing"]
+        else:
```
greenape: Need a test for this pls - check the codecov report.

flowmachine/tests/test_subscriber_interevent_period.py
```diff
@@ -0,0 +1,84 @@
...
+
+    query = IntereventPeriod("2016-01-01", "2016-01-08", "stddev")
+    df = get_dataframe(query).set_index("subscriber")
+    sample = df.sample(n=5)
+    want = intervent_period("2016-01-01", "2016-01-08", "both", sample, "std")
+    assert query.column_names == ["subscriber", "value"]
+    assert pd.to_numeric(sample["value"]).to_dict() == pytest.approx(want)
+
+
+def test_interevent_period_errors():
```
greenape: Split to two test funcs, or parametrise?

gzagatti:
Got the test parametrized.

PR #339, opened 2019-02-05
Reciprocal percent

gzagatti:
### I have:

### Description

This PR implements `ReciprocalContact` which outputs a `ContactBalance` list with an additional column `reciprocal` which shows whether the contact is reciprocal or not.

It also implements `ProportionReciprocal` which calculates the proportion of events which are between reciprocal contacts.

Unfortunately, the test data does not have any reciprocal contact except for `self_caller` which means unit testing is a bit lacking.


flowmachine/flowmachine/features/subscriber/contact_reciprocal.py
```diff
@@ -0,0 +1,388 @@
...
+        exclude_self_calls=True,
+        subscriber_subset=None,
+    ):
+        self.start = start
+        self.stop = stop
+        self.hours = hours
+        self.exclude_self_calls = exclude_self_calls
+        self.tables = tables
+
+        self.contact_reciprocal_query = ContactReciprocal(
```
greenape: Couldn't the class just take a `ContactReciprocal` object as the argument?

flowmachine/flowmachine/features/subscriber/contact_reciprocal.py
```diff
@@ -0,0 +1,388 @@
...
+
+        self.start = start
+        self.stop = stop
+        self.subscriber_identifier = subscriber_identifier
+        self.hours = hours
+        self.exclude_self_calls = exclude_self_calls
+        self.direction = direction
+        self.tables = tables
+
+        if self.direction == "both":
```
greenape: ValueError on bad vals?

flowmachine/flowmachine/features/subscriber/contact_reciprocal.py
```diff
@@ -0,0 +1,388 @@
...
+        return ["subscriber", "value"]
+
+    def _make_query(self):
+
+        filters = []
+        if self.direction != "both":
+            filters.append(
+                f"outgoing IS {'TRUE' if self.direction == 'out' else 'FALSE'}"
+            )
+        if self.exclude_self_calls:
```
greenape: This is the non-testable bit, because the only reciprocal caller is the self-caller, right? Can we have a test for it _anyway_, showing that person is excluded and everybody else has a 0?

flowmachine/flowmachine/features/subscriber/proportion_outgoing.py
```diff
@@ -16,7 +16,7 @@
 from .metaclasses import SubscriberFeature
 
 
-class ProportionOutgoing(SubscriberFeature):
+class ProportionEventOutgoing(SubscriberFeature):
```
greenape: Why's this been renamed? And if it needs to be renamed,  please rename file to correspond.

flowmachine/flowmachine/features/subscriber/contact_reciprocal.py
gzagatti: Good point. But my intention was to homogenize the call for feature extraction such that all of them have roughly the same parameters, that is, `start`, `stop`, `hours`, `tables`, etc. I find this approach is somewhat cleaner than having too many variations across feature classes. In fact, I would ideally like to see the init parameters for query classes defined in a parent class.

flowmachine/flowmachine/features/subscriber/proportion_outgoing.py
gzagatti: I have actually deleted this class in PR#348 as it becomes redundant with the introduction of `ProportionEventType`. If you think we will want to keep this class I can make the requested changes.

In any case, I renamed this class to align with `ProportionEventReciprocal` which was introduced here.

flowmachine/flowmachine/features/subscriber/contact_reciprocal.py
greenape: My rationale here is that bubbling up the subquery is slightly better from an API call & caching perspective, and there's not really any point in storing the params and passing them straight on down.

gzagatti: Does it mean, that you rather using `ContactReciprocal` as an argument or that I do not save pointers to the initialized params and rather just save a pointer to `ContactReciprocal`?

If we do decide to have `ContactReciprocal` as one of the arguments, could we perhaps allow the user to either pass all the usual ones or `ContactReciprocal`? Or is this bad coding?

I generally find convenient to be able to reference to attributes of the query class such as start, stop, tables, etc. But I don't know if that's the design choice we would like to make.

greenape: `ContactReciprocal` as only argument. I hate cluttered signatures.

flowmachine/flowmachine/features/subscriber/proportion_outgoing.py
gzagatti: I revert it back for now since we'll delete in another PR.

PR #355, opened 2019-02-07
Fix reversed hours logic

### LINKED ISSUE #352: EventTableSubset gives wrong result if `hours` is used with "reversed" values
**Describe the bug**

In `EventTableSubset`, if the second value in `hours` is smaller than the first it is interpreted as "spanning midnight" (see [L185](https://github.com/Flowminder/FlowKit/blob/master/flowmachine/flowmachine/features/utilities/event_table_subset.py#L185)). However, the generated SQL is currently wrong so that the result includes events that lie outside the `start` and `stop` dates.

**Product**
Flowmachine

**Version**
Current master

**To Reproduce**

First we use flowmachine to calculate a daily location limited to the date `2016-01-05`, a single subscriber and hours between 10pm-midnight and midnight-6am on that date:
```python
subset_query = CustomQuery("SELECT * FROM (VALUES ('dr9xNYK006wykgXj')) as tmp (subscriber)")
dl = daily_location(
    "2016-01-05",
    table="events.calls",
    hours=(22, 6),
    subscriber_subset=subset_query,
)
```
Result:
```
subscriber,pcod
dr9xNYK006wykgXj,524 4 12 63
```

However, checking directly in FlowDB it turns out this subscriber has no events on that day:
```
flowdb=# SELECT * FROM events.calls WHERE datetime::date = '2016-01-05' AND msisdn = 'dr9xNYK006wykgXj';
 id | outgoing | datetime | duration | network | msisdn | msisdn_counterpart | location_id | imsi | imei | tac | operator_code | country_code 
----+----------+----------+----------+---------+--------+--------------------+-------------+------+------+-----+---------------+--------------
(0 rows)
```

**Expected behaviour**

The daily_location calculation should return an empty result.

**Debugging**

The reason is that the generated SQL contains the following `WHERE` clause, with the `OR` outside of the last `AND` statement.
```
WHERE ((    (datetime >= '2016-01-05'::timestamptz)
        AND (datetime <= '2016-01-06'::timestamptz)
        AND (EXTRACT(HOUR FROM datetime) >= 22))
   OR (EXTRACT(HOUR FROM datetime) < 6))
```
As a consequence, the result includes events from _any_ day where calls were made before 6am.

Instead, the sql should read as follows:
```
WHERE (datetime >= '2016-01-05'::timestamptz)
  AND (datetime <= '2016-01-06'::timestamptz)
  AND ((   (EXTRACT(HOUR FROM datetime) >= 22)
        OR (EXTRACT(HOUR FROM datetime) < 6)))
```

While this is a fairly minor bug, it hints at some underlying problems:
- The manual SQL construction is error-prone, especially with large nested queries as they are being generated under the hood (and the resulting SQL strings are difficult to inspect by hand). Instead, we should be relying more / exclusively on `sqlalchemy.core` to construct SQL queries.
- The `hours` functionality is poorly designed, especially as far as the UI and construction is concerned. It uses tuples of integers in an attempt to capture a much richer concept and doesn't do a good job with it. This should be extracted out and put into its own class so that we can capture the concept of "limiting by time periods" in all its richness, add dedicated validation and error checks, and generate the necessary SQL in a more consistent way.

maxalbert:
Closes #352 

### I have:

### Description

The immediate bug fix for #352 is in `event_table_subset.py` (added missing parentheses to fix the grouping logic).

I also added a couple of integration tests which exhibit the bug / verify it is fixed. These are in `integration_tests/tests/flowmachine_tests/test_daily_location_results.py`. They run a `daily_location` query (with reversed hours to exhibit the bug) and check the resulting SQL, as well as the contents of the returned dataframe. I use [approvaltests](http://approvaltests.com/) for this because I wanted a convenient way to check large SQL strings and/or (potentially large) dataframes. We could probably use a simple method here, but I'm planning to add a few more approval tests during upcoming refactorings so it seems useful to introduce it here already.

I also moved the contents of `integration_tests/tests/flowmachine_server_tests/conftest.py` into `integration_tests/tests/conftest.py` so that the fixtures can be re-used in other tests.

As mentioned in #352, this is just a quick fix. Upcoming refactorings will help tidy up the underlying logic.


integration_tests/tests/flowmachine_tests/conftest.py
```diff
@@ -0,0 +1,18 @@
...
+@pytest.fixture
+def get_dataframe(fm_conn):
+    yield lambda query: pd.read_sql_query(query.get_query(), con=fm_conn.engine)
+
+
+@pytest.fixture(scope="session")
+def diff_reporter():
+    diff_reporter_factory = GenericDiffReporterFactory()
+    return diff_reporter_factory.get("opendiff")
+    # return diff_reporter_factory.get_first_working()
```
jch1g10: Has this comment been left intentionally?

maxalbert: Hmm, initially yes but it's probably better to remove it for now.

PR #360, opened 2019-02-07
Contact home distance

gzagatti:
### I have:

### Description

This PR implements a new covariate which calculates statistics for the distance between subscriber's own modal location and its contacts' modal location.

Calculations using `versioned-site` work but are really slow. We might want to consider only allowing the user to pass `ModalLocation` instance calculated with `versioned-cell`. Let me know what you think.


CHANGELOG.md
```diff
@@ -8,7 +8,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 - Dockerised development setup, with support for live reload of `flowmachine` and `flowapi` after source code changes.
 - Pre-commit hook for Python formatting with black.
-- Added new `IntereventPeriod`, `ContactReciprocal`, `ProportionContactReciprocal`, `ProportionEventReciprocal` query classes to FlowMachine.
+<<<<<<< HEAD
```
greenape: Merge conflict has snuck in here.

flowmachine/flowmachine/features/subscriber/contact_balance.py
```diff
@@ -136,14 +142,86 @@ def _make_query(self):
...
+        include_subscribers: bool, default True
+            Wether to include the list of subscribers in the subset as well.
+        """
+
+        return _ContactBalanceSubset(
+            contact_balance=self, include_subscribers=include_subscribers
+        )
+
+
+class _ContactBalanceSubset(SubscriberFeature):
```
greenape: I'm not quite getting this one I think. Doesn't it just give you the set of people who had an inbound call/text?

greenape: Ahh. I guess it would also allow you to get the set of subscribers who had an inbound call/text from some other subset of subscribers?

Why not make it a regular class rather than a hidden one?

flowmachine/flowmachine/features/subscriber/contact_modal_location_distance.py
```diff
@@ -0,0 +1,189 @@
...
+                self.stop,
+                hours=self.hours,
+                tables=self.tables,
+                subscriber_identifier="msisdn",
+                direction=self.direction,
+                exclude_self_calls=True,
+                subscriber_subset=subscriber_subset,
+            )
+
+        print(self.contact_balance_query.get_query())
```
greenape: Remove this.

flowmachine/flowmachine/features/subscriber/contact_modal_location_distance.py
```diff
@@ -0,0 +1,189 @@
...
+    hours : 2-tuple of floats, default 'all'
+        Restrict the analysis to only a certain set
+        of hours within each day.
+    subscriber_subset : str, list, flowmachine.core.Query, flowmachine.core.Table, default None
+        If provided, string or list of string which are msisdn or imeis to limit
+        results to; or, a query or table which has a column with a name matching
+        subscriber_identifier (typically, msisdn), to limit results to.
+    direction : {'in', 'out', 'both'}, default 'out'
+        Whether to consider calls made, received, or both. Defaults to 'out'.
+    column_name : str
```
greenape: This isn't in the init signature.

flowmachine/flowmachine/features/subscriber/contact_modal_location_distance.py
```diff
@@ -0,0 +1,189 @@
...
+        stop,
+        statistic="avg",
+        method="last",
+        *,
+        contact_balance=None,
+        tables="all",
+        direction="both",
+        hours="all",
+        subscriber_subset=None,
+        size=None,
```
greenape: This doesn't get passed on anywhere?

flowmachine/flowmachine/features/subscriber/contact_modal_location_distance.py
```diff
@@ -0,0 +1,189 @@
...
+    """
+
+    def __init__(
+        self,
+        start,
+        stop,
+        statistic="avg",
+        method="last",
+        *,
+        contact_balance=None,
```
greenape: It looks, further down, as though this needs to have specific params for this to work, so I'd be inclined to say don't allow passing one in.

flowmachine/flowmachine/features/subscriber/contact_modal_location_distance.py
```diff
@@ -0,0 +1,189 @@
...
+        subscriber           value
+    gwAynWXp4eWvxGP7    298.721500
+    GnyZMedmKQ4X78Wa    290.397556
+    BKMy1nYEZpnoEA7G     78.919136
+    m4L326vrwE6elJxQ    249.033988
+    NG1km5NzBg5JD8nj    188.679378
+               ...           ...
+    """
+
+    def __init__(
```
greenape: I'm wondering if it wouldn't be preferable to have this take either or both of the ML and CB as args and either create one from the other, or check they're compatible. I think this way kind of gives the illusion of being more flexible, but actually isn't because you've got lots of args you can set but they're the same across the two subqueries. Would also allow you to calculate this for things other than a ML e.g. a Hartigan clustered location, or a daily location.

flowmachine/flowmachine/features/subscriber/contact_modal_location_distance.py
```diff
@@ -0,0 +1,189 @@
...
+
+        self.statistic = statistic.lower()
+        if self.statistic not in valid_stats:
+            raise ValueError(
+                "{} is not a valid statistic. Use one of {}".format(
+                    self.statistic, valid_stats
+                )
+            )
+
+        self.distance_matrix_query = DistanceMatrix(level="versioned-cell")
```
greenape: You mentioned that this was super slow in some scenarios, I'm wondering if this might be why - you don't really need the full n*n matrix of distances here, so it might actually be faster to use `ST_Distance` directly on that reduced set.

I reckon that while in theory the query planner should skip creating the whole matrix, and instead only do it for things which will be included in the join, it probably doesn't because it'll need to do the calculation to work out if it _should_ be in the join.

flowmachine/flowmachine/features/subscriber/contact_balance.py
gzagatti: This class is meant to get the subset of subscriber contacts, given that not all contacts are reciprocals.

`Flowmachine` does not allow one to subset by the `msisdn` and `msisdn_counterpart` we can only subset by one or the other column. However, as in this case we want to calculate home location for the subset of our target subscribers and for all his contacts as well. This is basically what you mentioned in your previous comment.

I am indifferent between making this a regular or hidden class. My reasoning was that one can only get the subset of contacts by instantiating a `ContactBalance` class. I feel like getting the subset of subscribers would be a method of this class. So we would never want to instantiate a `ContactBalanceSubset` class on its own except when invoking this particular method.

I let you make the final decision on this one.

flowmachine/flowmachine/features/subscriber/contact_modal_location_distance.py
gzagatti: I had initially created the class in the way you described above. But then I reverted to what I have now because I wanted the signature of the feature classes to be somewhat consistent.

Requesting the ML and CB classes actually make the whole thing much easier to define. In which case, it would also make sense to allow for other classes instead of ML such as Hartigan and/or daily location. 

One reason why I wanted to exclude Hartigan was because of what I discuss in the next comment below.

I would then also rename the class to something more broad like `ContactHomeDistance`?

flowmachine/flowmachine/features/subscriber/contact_modal_location_distance.py
gzagatti: The thinking here was that in this reduced subset calculating `ST_Distance` for each and every pair would be quicker. But in a production scenario the cost of the `DistanceMatrix` would be negligible compared to all distance pairs. The matrix would also probably already be cached because other queries already use it.

Because I wanted to avoid the cost of calculating `ST_Distance` for every pair that I exclude Hartigan from this class. Alternatively, I would consider first getting all distinct pairs, calculating the distances and then joining back to the list of all pairs. And in the case that `DistanceMatrix` is available using it instead for the ML case. Do you think this would be more efficient?

gzagatti:
@greenape, I will let you comment on my comments before proceeding with the modifications.

flowmachine/flowmachine/features/subscriber/contact_modal_location_distance.py
greenape: I think the latter (get the set of distinct pairs, calculate distances for those, join back) is what I'm suggesting here. There's not a good mechanism to skip doing that iff a distance matrix is applicable and pre-calculated, so my thinking is that not using the distance matrix is the better path, and we can introduce it if this untenable in production.

flowmachine/tests/test_contact_reference_locations_stats.py
```diff
@@ -0,0 +1,114 @@
...
+    ml = CustomQuery(
+        f"SELECT subscriber, ST_POINT(lon, lat) AS loc FROM ({ml.get_query()}) _",
+        ["subscriber", "loc"],
+    )
+    query = ContactReferenceLocationStats(cb, ml, statistic="avg", geom_column="loc")
+    df = get_dataframe(query).set_index("subscriber")
+    assert df.value["gwAynWXp4eWvxGP7"] == pytest.approx(298.7215)
+
+
+@pytest.mark.parametrize("kwarg", ["statistic", "contact_locations"])
```
greenape: I'd prefer this split into three tests I think, because there's too much logic required to deal with the parameterisation.

greenape:
Looks like a couple of the `ValueError` aren't tested, hence codecov report blocks from merge. I'm happy otherwise I think.

PR #365, opened 2019-02-08
Docs cleanup

### LINKED ISSUE #364: Invisible docs pages
The recent docs restructure has made the important information clearer and easier to find, but by explicitly defining the order of pages in the navbar several pages have become invisible. Also, several of these invisible pages contain information which is duplicated in the visible pages, and are therefore redundant.

We can't explicitly define the full nav order without explicitly including all of the auto-generated pages (which would defeat the point of auto-generating them), and `mkdocs` currently has no facility for partially defining the nav, so we will either have to work with the automatic alphanumeric ordering or use a plugin to define the page order.

jch1g10:
Closes #364

### I have:

### Description

I have removed the explicit `nav` definition in `mkdocs.yml`, so that all the docs pages appear in the navbar. I have numbered the top-level filenames to fix the order without defining the `nav`, but `mkdocs` orders the files and directories separately by default, so I have flattened the hierarchy of the main docs pages so that we can fix their order.

To do this, I have moved the auto-generated docs for FlowMachine and FlowClient into the flowmachine/ and flowclient/ directories (we will probably want to add similar directories for FlowAPI, FlowDB and FlowAuth at some point). I have moved the example notebook into flowclient/, but to avoid it getting mixed up with the flowclient docs I have put the docs in flowclient/flowclient/ (and similar for flowmachine in case we want to add other files to the flowmachine/ directory). This isn't an optimal layout, and I hope we will come up with a better structure.

I have also:
- Put back `docs/README.md`, which was removed by mistake.
- Moved some of the FlowAuth user instructions from the installation page to the analyst page.
- Fixed some formatting in the license.
- Removed files whose contents are duplicated elsewhere.
- Cleared outputs in `example-usage.ipynb`
- Added some details from the FlowAuth README which were missing from the docs.


docs/mkdocs.yml
```diff
@@ -33,7 +26,8 @@ theme:
   # Default values, taken from mkdocs_theme.yml
   language: en
   feature:
-    tabs: true
+    # True to use tabs at the top, in addition to the side navbar
```
maxalbert: Thanks for this comment! 👍

maxalbert:
Nice work! 👍 

I agree that there is more cleanup to be done here. One obvious issue is that the section "Setting up a development environment" should be a sub-page of the "Developer" section, but currently it is in its own top-level page, at the same level as "Developer".

However, this is clearly an improvement over before. Given that we need to do more restructuring anyway I'm happy to merge this and address further issues in subsequent PRs.

Issue #373, opened 2019-02-12
Refactor flowmachine internals


maxalbert:
Some of the classes in flowmachine suffer from entangled code with mutual dependencies, logic being in the wrong place and/or repeated in multiple places, and the fact that they are mostly procedural code wrapped up in classes. We should extract out coherent bits of logic and encapsulate it so that each class has a single, well-defined responsibility and can be used without knowledge of the internals leaking into other places.

Obvious low hanging fruit:
- [ ] Subsetting logic
- [ ] Spatial aggregation logic
- [ ] Dealing with time periods / temporal filtering
- [ ] ...

We'll track these in individual issues to break down the refactoring work into manageable chunks.

PR #375, opened 2019-02-12
Refactor subsetting

### LINKED ISSUE #374: Refactor subsetting logic
The subsetting logic currently lives in `EventTableSubset`. The `._make_query()` method manually tweaks the SQL code to apply a subset, but this is done in a fairly ad-hoc way for each of the supported kinds of subsets, and it is mixed with other logic (e.g. temporal filtering). This should be broken out.

maxalbert:
Closes #374 

### I have:

### Description

This PR is a first stab to address #373 and refactor the flowmachine internals.

It starts with the subsetting machinery (#374). The key changes are in `flowmachine/flowmachine/features/utilities/event_table_subset.py` and `flowmachine/flowmachine/core/subscriber_subset.py`.

Previously the class `EventTableSubset` would manually tweak the SQL string in order to subset a given query. The logic for this (in `_make_query`) is not very clear, and even slightly buggy (see #352). One of the things that makes the existing logic awkward is that it needs to deal with different kinds of subsets (e.g. subsetting by another flowmachine query, or using an explicit list), and the required modifications to the existing SQL string are applied in a relatively ad-hoc manner and don't allow for a uniform treatment.

The main purpose of this PR is to extract the subsetting logic from `EventTableSubset` and to allow applying it uniformly, independently of the kind of subset we're dealing with. (Future refactorings will extract other pieces of logic in a similar way, such as spatial aggregation, temporal filtering, etc.)

The logic was moved into `subscriber_subset.py`, which contains different classes (derived from `SubscriberSubsetBase`) for the different kinds of subsets:
- `AllSubscribers` (= no subsetting)
- `SubsetFromFlowmachineQuery` (= subsetting by a flowmachine query)
- `ExplicitSubset` (= subsetting by an explicit list of subscribers)

In order to allow a uniform treatment the implementation uses `sqlalchemy` to apply the `WHERE` clause (this is also a first step towards moving away from the manual SQL generation).

The new code is applied in the method `EventTableSubset._make_query_with_sqlalchemy()` in the following line:
```
select_stmt = self.subscriber_subset.apply_subset(
    select_stmt, subscriber_identifier=self.subscriber_identifier
)
```
The rest of this method is still mostly identical to the original version (kept for now in `_make_query_ORIG()`), but construts a sqlalchemy select statement (`select_stmt`) which gets fed into the subsetting routine (this mimics the original manual SQL statement as closely as possible - see [this commit](https://github.com/Flowminder/FlowKit/commit/94a22470e490d8ce97fa8432be34737892c7693e) for the changes which the switch to sqlalchemy introduced).

During the refactoring I wanted to ensure that I didn't accidentally change the functionality and stayed as close to the existing SQL strings as possible, so I added a few functional tests in `flowmachine/functional_tests/test_daily_location.py` which use `approvaltests` to verify both the generated SQL query string, as well as the resulting pandas dataframe, of a few daily location queries with different parameters.

Note that currently the base class `SubscriberSubsetBase` inherits from `flowmachine.core.Query`, but this is more out of necessity because otherwise subsets would not be taken into account for the md5 generation (or it would have required a much more invasive change to enable this). As a consequence, I had to provide stub implementations of `_make_query` and `_get_query_attrs_for_dependency_graph` for the subset classes. In the long run this should not be needed, however, and we should get rid of these again.

It would be great to get a few pairs of eyes on this, any comments are welcome.


flowmachine/flowmachine/core/subscriber_subset.py
```diff
@@ -0,0 +1,186 @@
...
+
+
+class SubsetFromFlowmachineQuery(SubscriberSubsetBase):
+    """
+    Represents a subset given by a flowmachine query.
+    """
+
+    is_proper_subset = True
+
+    def __init__(self, flowmachine_query):
```
gzagatti: Can we expand the docstring to say what this query should contain in terms of columns, etc.

flowmachine/flowmachine/core/subscriber_subset.py
```diff
@@ -0,0 +1,186 @@
...
+        Returns
+        ----------
+        sqlalchemy.sql.ClauseElement
+        """
+        assert isinstance(sql, ClauseElement)
+
+        tbl = sql.alias("tbl")
+
+        try:
+            subset = self.subset_query.distinct().alias("subset")
```
gzagatti: Didn't you use `text(flowmachine.get_query())` to transform the query into an sqlalchemy query? So when would we get an error here?

flowmachine/flowmachine/core/subscriber_subset.py
```diff
@@ -0,0 +1,186 @@
...
+        ----------
+        sqlalchemy.sql.ClauseElement
+        """
+        assert isinstance(sql, ClauseElement)
+        assert len(sql.froms) == 1
+        parent_table = sql.froms[0]
+        return sql.where(parent_table.c[subscriber_identifier].in_(self.subscribers))
+
+
+def make_subscriber_subset(subset):
```
gzagatti: I find this name confusing. I would go for something like `make_subscriber_subset_factory` or `make_subscriber_subset_generator`

flowmachine/flowmachine/features/utilities/event_table_subset.py
```diff
@@ -164,7 +173,47 @@ def _check_dates(self):
                 stacklevel=2,
             )
 
-    def _make_query(self):
+    def _make_query_with_sqlalchemy(self):
+        table = get_sqlalchemy_table_definition(
```
gzagatti: Should we call `sqlalchemy_table` to match with `sqlalchemy_columns`?

flowmachine/flowmachine/core/subscriber_subset.py
```diff
@@ -0,0 +1,186 @@
...
+        # enough to ensure Query.dependency_graph() doesn't break.
+        attrs = {}
+        attrs["name"] = self.__class__.__name__
+        attrs["stored"] = "N/A"
+        attrs["cost"] = "N/A"
+        attrs["runtime"] = "N/A"
+        return attrs
+
+
+class AllSubscribers(SubscriberSubsetBase):
```
gzagatti: I think the existence of this class slightly confusing, I think we should not be calling down below
   
```
select_stmt = self.subscriber_subset.apply_subset(
            select_stmt, subscriber_identifier=self.subscriber_identifier
        ) 
```

when we are not extracting a subset. I would rather go for an if-statement like `if subset: ...`

flowmachine/flowmachine/features/utilities/event_table_subset.py
```diff
@@ -164,7 +173,47 @@ def _check_dates(self):
...
+            else:
+                # If dates are backwards, then this will be interpreted as spanning midnight
+                select_stmt = select_stmt.where(
+                    or_(
+                        extract("hour", table.c.datetime) >= hour_start,
+                        extract("hour", table.c.datetime) < hour_end,
+                    )
+                )
+
+        select_stmt = self.subscriber_subset.apply_subset(
```
gzagatti: See my comment above. I would consider using an if-statement and only calling when we want to subset by a subscriber.

flowmachine/flowmachine/features/utilities/event_table_subset.py
```diff
@@ -164,7 +173,47 @@ def _check_dates(self):
                 stacklevel=2,
             )
 
-    def _make_query(self):
+    def _make_query_with_sqlalchemy(self):
+        table = get_sqlalchemy_table_definition(
+            self.table.fully_qualified_table_name, engine=Query.connection.engine
+        )
+        sqlalchemy_columns = [
```
gzagatti: In the same vein as with `self.subscriber_subset_ORIG`, should we not also be initializing `self.column` and `self.table` directly as `sqlalchemy` objects in `__init__` and removing those calls here.

flowmachine/flowmachine/core/sqlalchemy_utils.py
```diff
@@ -0,0 +1,133 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import pandas as pd
+from sqlalchemy import Table, MetaData
+from sqlalchemy.dialects import postgresql
+from sqlalchemy.sql import Selectable
+
+
+def get_sqlalchemy_table_definition(fully_qualified_table_name, *, engine):
```
gzagatti: I am not sure the name is really explanatory here. I first thought you were getting a generic table when you are only after `events` table. Second, I would rather move this to `event_table_subset.py` unless you consider using it in other places.

flowmachine/flowmachine/core/sqlalchemy_utils.py
```diff
@@ -0,0 +1,133 @@
...
+        Fully qualified table name. Example: "events.calls"
+    engine : sqlalchemy.engine.Engine
+        SQLAlchemy engine to use for reading the table information.
+
+    Returns
+    -------
+    sqlalchemy.Table
+    """
+    metadata = MetaData()
+    if fully_qualified_table_name == "events.calls":
```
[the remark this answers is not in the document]
maxalbert: Sure, I was considering this. Originally I wanted to be very strict about which tables are allowed in order to catch any errors / unexpected situations during refactoring (e.g. if there are different columns present in the different tables), but this seems less of an issue now. I'll change it to a more generic version.

flowmachine/flowmachine/core/subscriber_subset.py
```diff
@@ -0,0 +1,186 @@
...
+            f"Class {self.__class__.__name__} does not implement 'is_proper_subset'"
+        )
+
+    @abstractmethod
+    def apply_subset(self, sql, *, subscriber_identifier):
+        raise NotImplementedError(
+            f"Class {self.__class__.__name__} does not implement 'apply_subset'"
+        )
+
+    def _get_query_attrs_for_dependency_graph(self, analyse=False):
```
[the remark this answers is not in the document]
maxalbert: Yes, this is a subtle one. The only reason I had to factor this out into a helper function is because the new subset classes currently derive from `flowmachine.Query` (which I'd like to change again soon-ish but which was the easiest way of avoiding ripples through the rest of the code base). As a consequence of deriving from `Query`, the `dependency_graph()` function will try to call `.explain()` and `.is_stored` on each sub-query that contributes to the given flowmachine query. In particular, this will be called for the subset classes, too, but they don't represent full calleable queries so these don't make sense. The easiest way I could see to work around this was to factor out the code which retrieved information about this query into a helper function (`._get_query_attrs_for_dependency_graph()`) and the implement a stub for this for the subset classes which just returned a dictionary of the correct shape (but without having to call `.explain()` and `.is_stored` internally.

It's not an ideal solution, but I also don't see any harm in it because the helper function `Query._get_query_attrs_for_dependency_graph()` isn't a bad idea in any case (it separates the logic which extracts info about the query from the rest of the logic in `Query.dependency_graph()` which only applies attributes for drawing the graph).

Does this make things clearer?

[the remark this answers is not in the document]
maxalbert: To answer your question: _"Don't we still need to be able to know whether the query has been stored, the cost, etc? Is it something we cannot derive anymore?"_

Yes, we still need to be able to do it. And we're still doing it - except that it now happens in the helper function `Query._get_query_attrs_for_dependency_graph()` rather than directly in ` Query.dependency_graph()`.

flowmachine/flowmachine/core/sqlalchemy_utils.py
```diff
@@ -0,0 +1,133 @@
...
+
+    with engine.connect() as con:
+        result = con.execute(query)
+
+    columns = [c.name for c in query.columns]
+    df = pd.DataFrame(result.fetchall(), columns=columns)
+    return df
+
+
+def get_sqlalchemy_column(table, column_str):
```
josh-gree: Why does this not just have two arguments? column and label?

flowmachine/flowmachine/core/subscriber_subset.py
```diff
@@ -0,0 +1,186 @@
...
+
+
+class AllSubscribers(SubscriberSubsetBase):
+    """
+    Represents the subset of all subscribers (i.e., no subsetting at all).
+    """
+
+    is_proper_subset = False
+
+    def _make_query(self):
```
josh-gree: So what happens if another Query tries to make use of one of these subscriber objects? The string from `get_query` will not be valid sql?

flowmachine/flowmachine/core/subscriber_subset.py
```diff
@@ -0,0 +1,186 @@
...
+    Represents a subset given by a flowmachine query.
+    """
+
+    is_proper_subset = True
+
+    def __init__(self, flowmachine_query):
+        assert isinstance(flowmachine_query, Query)
+
+        self.subset_query = text(flowmachine_query.get_query()).columns(
+            column("subscriber")
```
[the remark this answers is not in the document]
josh-gree: X2

flowmachine/Pipfile
```diff
@@ -34,6 +34,7 @@ pytest-cov = "*"
 pytest-testmon = "*"
 asynctest = "*"
 cachey = "*"
+approvaltests = "*"
```
josh-gree: Could you explain the idea behind this and how it differs from other approaches to testing?

maxalbert: The main idea is that you produce some output and ask approvaltests to verify it. The first time the test runs you are asked to approve the output, and approaltests then stores it in a file. On subsequent test runs it simply compares the new output to the previous one. The test passes if the output hasn't changed, otherwise it fails.

This is basically the same idea as with usual tests and assert statements. The main advantage is in situations where the output is large (e.g. a full dataframe or a large SQL string, or any other large object). This is where approvaltests takes a lot of the scaffolding and boilerplate away from you because it presents you with a diff that you can quickly inspect visually and then either approve or reject.

A lot of our existing tests in flowmachine only check a few hand-selected values. This is usually ok, but makes it easy to miss subtle differences during refactoring. I wanted to ensure that the full output is _exactly_ as it was before, or if there are any changes to see exactly what they are (both in terms of the generated SQL and the resulting dataframe), which approvaltests makes very easy.

flowmachine/flowmachine/core/sqlalchemy_utils.py
maxalbert: Mainly because of the way it's being used in `EventTableSubset`, where the `column_str` comes from flowmachine. Because `column_str` can be either of the form `<column_name>` or `<column_name> AS <alias>` this string needs to be potentially broken apart, and the main point of this helper function is to do this here rather than outside. So it sounds like I need to find a better name for this function, to make it clearer. Any suggestions?

flowmachine/flowmachine/core/subscriber_subset.py
maxalbert: Yes. This is clearly not great but it should only be transitional until we can avoid the subsets being inherited from `Query`, and no other query should need to make use of this string other than in generating the md5 hash. Now that you mention it I do wonder, though, how this will affect the caching and whether we need to include the arguments in the string returned from here...

flowmachine/flowmachine/core/subscriber_subset.py
maxalbert: I can see where you're coming from, but the existence of this class is actually crucial to reduce / eliminate coupling with other parts of the code. Let me try to explain why.

One of the reasons why the current code base is so entangled is because certain classes apply logic or behaviour _on behalf of_ other classes when they shouldn't need to know anything about how things are implemented. A lot of the `if`/`elif`/`else` statements in the code currently have this purpose (especially where they have the form`if isinstance(...)`).

If we omit the `AllSubscribers` case here then we suddenly have a special case that needs to be treated _elsewhere_ depending on whether we do or do not have a proper subset. This check might only happen once (for example in `EventTableSubset`), but once other classes are also interested in subsetting, all of a sudden this if/else statement is duplicated in multiple different places. If you later change anything about the implementation of the subset classes which affects how this check works you end up having to do "shotgun surgery" and change it in all those different places.

By contrast, if you _do_ have an `AllSubscribers` class then there are no ripple effects because all types of subsets are treated exactly the same. So the only thing that any external code needs to know is how to apply a subset (by calling the `.apply_subset()` method of the subset class, no matter which of the multiple different cases it is).

I agree that it feels strange when you first see it, but it's actually a powerful technique to eliminate dependencies. It's also in the spirit of true object-oriented programming (rather than the procedural-style programming inside classes that we have in most places) where instead of _asking_ for information from an object (here: a subset) and then applying different actions depending on the answer you simply _tell_ it what to do (here: apply a subset, whatever it is) and trust that it knows how to do it.

I'm happy to dig up more resources and examples of why this is useful if you're interested, but all I can say at the moment is that using this technique in more places will allow us to massively reduce coupling and dependencies in the code.

maxalbert: Thinking about it, maybe we can reduce the confusion by renaming the method `.apply_subset()` to `.apply_subset_if_needed()`? This would make it clearer that the "no subsetting needed" case is intentionally included.

flowmachine/flowmachine/features/utilities/event_table_subset.py
maxalbert: See reply above, I'd very much like to not do this because it's one of the main sources of coupling in the code base at the moment.

flowmachine/flowmachine/core/subscriber_subset.py
```diff
@@ -0,0 +1,186 @@
...
+    """
+    Represents a subset given by a flowmachine query.
+    """
+
+    is_proper_subset = True
+
+    def __init__(self, flowmachine_query):
+        assert isinstance(flowmachine_query, Query)
+
+        self.subset_query = text(flowmachine_query.get_query()).columns(
```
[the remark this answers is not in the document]
maxalbert: Excellent point, I'll move it there. 👍

flowmachine/flowmachine/core/subscriber_subset.py
maxalbert: 👍

flowmachine/flowmachine/core/subscriber_subset.py
maxalbert: Yes, you are right. Or rather, I think this will _always_ raise an error because the textual query doesn't have a `.distinct()` attribute. I think I was confused when I wrote this. 😂Should be clearer once I've reworked it following your suggestion above to move `flowmachine.get_query()` down here.

flowmachine/flowmachine/core/subscriber_subset.py
maxalbert: Agreed it is quite confusing, especially because we have multiple different types of subsets at the moment (which should be cleaner again once the refactoring is complete). I'd like to avoid the word "factory" if possible because it sounds like tech lingo and some people (not me) have an aversion against it. Also, we're not really making a subscriber subset factory (rather, this _is_ the subscriber subset factory).

How about `make_subscriber_subset_object` or similar? Does this make it clearer that we're creating an object from "raw" input (such as a string or a list)?

flowmachine/flowmachine/core/sqlalchemy_utils.py
maxalbert: Yes, good point. The name was much more general than the implementation. I have changed the implementation so that it will work for any existing table. I do expect to reuse this in other places, so I'd rather keep it here.

flowmachine/flowmachine/core/subscriber_subset.py
maxalbert: I actually think it's worth making it explicit in the names that this object plays an "active" role rather than just representing a "passive" subset. I'll rename the classes from "...Subset" to "...Subsetter" and call this function `make_subscriber_subsetter`. Hopefully this will make things more obvious.

flowmachine/flowmachine/features/utilities/event_table_subset.py
maxalbert: 👍

flowmachine/flowmachine/features/utilities/event_table_subset.py
maxalbert: Yes, that's a good suggestion. At first I wanted to keep the changes in this PR as confined as possible (to allow for gradual refactoring) but this seems to make more sense.

flowmachine/flowmachine/core/sqlalchemy_utils.py
maxalbert: I have renamed this function to make it a bit clearer that it has a pretty narrow purpose at the moment. I expect this to go away in the medium run once the sqlalchemy table and column definitions are created a bit more centrally (and simply passed into the classes that need them).

maxalbert:
Many thanks for the useful comments @gzagatti and @josh-gree, very helpful! I have revised the PR according to your suggestions (with the only exception of removing the `AllSubscribers` subset class - see my comment for justification). Hopefully this will make things a lot clearer. Please let me know if you have further comments or questions.

flowmachine/flowmachine/core/subscriber_subset.py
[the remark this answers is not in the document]
gzagatti: Ok, got it. So basically you are saying that `sqlalchemy` queries do not have the `explain` and `is_stored` which would break the subset class otherwise.

My point was that the Subset `_get_query_attrs_for_dependency_graph` method does not contain information about whether it has been stored, costs, etc because they are just stubs. So I was just wondering if this is a temporary state of affairs.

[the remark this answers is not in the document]
maxalbert: Ah, I see. Yes, you are right. And I can see that the comment doesn't really clarify this. 😂 I'll tweak it.

flowmachine/flowmachine/core/subscriber_subset.py
gzagatti: Got it. Very clear explanation. I am indifferent about renaming, but feel more strongly about `apply_subset` just because it is shorter. At the end of the day, selecting all is a special subset case.

flowmachine/flowmachine/core/subscriber_subsetter.py
```diff
@@ -0,0 +1,249 @@
...
+            The flowmachine query to be used for subsetting. The only requirement
+            on it is that the result has a column called "subscriber" (it is fine
+            for other columns to be present, too).
+        """
+        assert isinstance(flowmachine_query, Query)
+
+        self._verify_that_subscriber_column_is_present(flowmachine_query)
+        self.flowmachine_query = flowmachine_query
+
+    def _verify_that_subscriber_column_is_present(self, flowmachine_query):
```
gzagatti: Not sure if this might be a replication of [this](https://github.com/Flowminder/FlowKit/blob/1334dcec5a237b45c5078790ea3a5fae0b67eab8/flowmachine/flowmachine/utils/utils.py#L245) which will be soon merged.

flowmachine/flowmachine/core/subscriber_subsetter.py
```diff
@@ -0,0 +1,249 @@
...
+        ]
+
+        # Explicitly inform the textual query about the output columns we expect
+        # and provide an alias. This allows the generated SQL string to be embedded
+        # as a subquery in other queries.
+        sqlalchemy_subset_query = textual_sql.columns(*sqlalchemy_columns).alias(
+            "subset_query"
+        )
+
+        # SQL string. Also explicitly mark the expected columns o, and ensure
```
gzagatti: Comment incomplete?

flowmachine/flowmachine/features/utilities/event_table_subset.py
```diff
@@ -94,6 +103,10 @@ def __init__(
                 )
         self.columns = sorted(self.columns)
```
gzagatti: Does it make sense to have `self.sqalchemy_columns` here instead of line 183 like you do with `self.sqlachemy_table` or is it overkill?

flowmachine/flowmachine/core/subscriber_subsetter.py
maxalbert: Yes, I think it is. Good spot. I'll change it once #320 has been merged.

flowmachine/flowmachine/features/utilities/event_table_subset.py
maxalbert: I actually moved it here temporarily but in the end decided to move it back. This doesn't feel ideal yet but it's probably better to refactor it further in a separate PR.

PR #420, opened 2019-02-25
column_names

### LINKED ISSUE #419: All queries should declare a column_names prop, and this should be enforced
Many queries now declare the columns they are expected to return via the `column_names` property, but this should be universally the case, and should be enforced, rather than relying on running `head(0)` to get Postgres to infer the column names.

Helpful in a small way for #390 and #391.

greenape:
Closes #419 

### I have:

### Description

This adds a `column_names` property to those query types which had been missing it up until now, and mandates that classes inheriting from `Query` implement one. I've also turned the two `FeatureCollection` classes into functions, because there was nothing in them to warrant being classes really.

### DIFF
--- a/flowmachine/flowmachine/core/custom_query.py
+++ b/flowmachine/flowmachine/core/custom_query.py
@@ -55,5 +55,8 @@ class CustomQuery(Query):
 
     def __getstate__(self):
         state = super().__getstate__()
-        del state["_column_names"]
+        try:
+            del state["_column_names"]
+        except KeyError:
+            pass
         return state
--- a/flowmachine/flowmachine/core/custom_query.py
+++ b/flowmachine/flowmachine/core/custom_query.py
@@ -7,6 +7,8 @@
 Simple utility class that allows the user to define their
 own custom query via a python string.
 """
+from typing import List
+
 from .utils import pretty_sql
 from .query import Query
 
@@ -20,11 +22,13 @@ class CustomQuery(Query):
     ----------
     sql : str
         An sql query string
+    column_names : list of str
+        The column names to return
     
     Examples
     --------
 
-    >>> CQ = CustomQuery('SELECT * FROM events.calls')
+    >>> CQ = CustomQuery('SELECT * FROM events.calls', ["msisdn"])
     >>> CQ.head()
 
 
@@ -33,13 +37,18 @@ class CustomQuery(Query):
     .table.Table for an equivalent that deals with simple table access
     """
 
-    def __init__(self, sql):
+    def __init__(self, sql: str, column_names: List[str]):
         """
 
         """
 
         self.sql = pretty_sql(sql)
+        self._column_names = column_names
         super().__init__()
+
+    @property
+    def column_names(self) -> List[str]:
+        return self._column_names
 
     def _make_query(self):
         return self.sql
--- a/flowmachine/flowmachine/core/custom_query.py
+++ b/flowmachine/flowmachine/core/custom_query.py
@@ -7,7 +7,7 @@
 Simple utility class that allows the user to define their
 own custom query via a python string.
 """
-from typing import List
+from typing import List, Set, Union
 
 from .utils import pretty_sql
 from .query import Query
@@ -22,7 +22,7 @@ class CustomQuery(Query):
     ----------
     sql : str
         An sql query string
-    column_names : list of str
+    column_names : list of str or set of str
         The column names to return
     
     Examples
@@ -37,13 +37,12 @@ class CustomQuery(Query):
     .table.Table for an equivalent that deals with simple table access
     """
 
-    def __init__(self, sql: str, column_names: List[str]):
-        """
-
-        """
-
+    def __init__(self, sql: str, column_names: Union[List[str], Set[str]]):
         self.sql = pretty_sql(sql)
-        self._column_names = column_names
+        seen = {}  # Dedupe the column names but preserve order
+        self._column_names = [
+            seen.setdefault(x, x) for x in column_names if x not in seen
+        ]
         super().__init__()
 
     @property
@@ -52,11 +51,3 @@ class CustomQuery(Query):
 
     def _make_query(self):
         return self.sql
-
-    def __getstate__(self):
-        state = super().__getstate__()
-        try:
-            del state["_column_names"]
-        except KeyError:
-            pass
-        return state
--- a/flowmachine/flowmachine/core/custom_query.py
+++ b/flowmachine/flowmachine/core/custom_query.py
@@ -52,3 +52,8 @@ class CustomQuery(Query):
 
     def _make_query(self):
         return self.sql
+
+    def __getstate__(self):
+        state = super().__getstate__()
+        del state["_column_names"]
+        return state
--- a/flowmachine/flowmachine/core/query.py
+++ b/flowmachine/flowmachine/core/query.py
@@ -449,7 +449,7 @@ class Query(metaclass=ABCMeta):
             return []
 
         Q = f"""EXPLAIN (ANALYZE TRUE, TIMING FALSE, FORMAT JSON) CREATE TABLE {full_name} AS 
-            ({self._make_query() if force else self.get_query()})"""
+            (SELECT {", ".join(self.column_names)} FROM ({self._make_query() if force else self.get_query()}) _)"""
         queries.append(Q)
         for ix in self.index_cols:
             queries.append(
--- a/flowmachine/flowmachine/core/query.py
+++ b/flowmachine/flowmachine/core/query.py
@@ -255,6 +255,7 @@ class Query(metaclass=ABCMeta):
         return self.get_dataframe_async().result()
 
     @property
+    @abstractmethod
     def column_names(self) -> List[str]:
         """
         Returns the column names.
--- a/flowmachine/flowmachine/core/query.py
+++ b/flowmachine/flowmachine/core/query.py
@@ -226,15 +226,13 @@ class Query(metaclass=ABCMeta):
                 try:
                     return self._df.copy()
                 except AttributeError:
-                    qur = f"SELECT {', '.join(self.column_names)} FROM ({self.get_query()}) _"
+                    qur = f"SELECT {self.column_names_as_string_list} FROM ({self.get_query()}) _"
                     with self.connection.engine.begin():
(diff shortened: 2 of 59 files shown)


flowmachine/flowmachine/core/custom_query.py
```diff
@@ -33,13 +37,26 @@ class CustomQuery(Query):
...
         super().__init__()
 
+    @property
+    def column_names(self) -> List[str]:
+        return self._column_names
+
     def _make_query(self):
         return self.sql
+
+    def __getstate__(self):
```
maxalbert: I'm not sure I understand what's going on in this method. First of all, why would it try to delete the attribute `_column_names`? Is this not essential to determine this query? Secondly, why would the `del` line ever fail, since this query does have an attribute called `self._column_names`? Would be good to explain these things in a docstring or comment.

flowmachine/flowmachine/features/utilities/feature_collection.py
```diff
@@ -3,13 +3,13 @@
...
-Class definition for FeatureCollection, this is a group of
+Class definition for feature_collection, this is a group of
 joined features.
 """
-from ...core import Query
+from flowmachine.core.join import Join
 
 
-class FeatureCollection(Query):
+def feature_collection(metrics, dropna=True) -> Join:
```
maxalbert: Mainly out of interest (certainly not against this): what made it necessary / desirable to turn `feature_collection` into a function rather than leaving it as a class?

flowmachine/flowmachine/core/query.py
```diff
@@ -227,13 +226,15 @@ def do_get():
                 try:
                     return self._df.copy()
                 except AttributeError:
-                    qur = self.get_query()
+                    qur = f"SELECT {', '.join(self.column_names)} FROM ({self.get_query()}) _"
```
maxalbert: Looks like we're doing `', '.join(self.column_names)` a lot, typically inside f-strings. How about adding a convenience property `self.column_names_text` or `self.column_names_str` to wrap this up and make it more readable?

flowmachine/flowmachine/features/spatial/distance_matrix.py
```diff
@@ -71,6 +73,25 @@ def __init__(self, level="versioned-cell", date=None, return_geometry=False):
...
+
+        try:
+            cols.remove("lat")
+            cols.remove("lon")
+        except ValueError:
+            pass  # Nothing to remove
+
+        col_names = [f"{c}_from" for c in cols]
+        col_names += [f"{c}_to" for c in cols]
+        col_names += [f"{c}_from" for c in ("lon", "lat")]
```
maxalbert: Just to check: the lon/lat columns may be missing from the columns (depending on the value of `self.level`), but here we are adding them unconditionally. Is this intentional?

flowmachine/flowmachine/features/spatial/location_cluster.py
```diff
@@ -187,6 +188,13 @@ def __init__(
 
         super().__init__()
 
+    @property
+    def column_names(self) -> List[str]:
+        if self.aggregate:
```
maxalbert: Not relevant to this PR, rather a note to self: the name `self.aggregate` is confusing because it doesn't indicate whether it's a boolean or some sort of object representing an aggregate. Would be useful to rename this for clarity (or rather, we probably shouldn't be doing the aggregation inside this class in the first place.)

flowmachine/flowmachine/features/spatial/versioned_infrastructure.py
```diff
@@ -58,25 +62,25 @@ def __init__(self, table="sites", date=None):
...
 
-        self.table = table
+        self.table = Table(schema="infrastructure", name=table)
         self.date = date
 
         super().__init__()
 
+    @property
+    def column_names(self) -> List[str]:
+        return self.table_name.column_names
```
maxalbert: Should this be `self.table.column_names`?

flowmachine/flowmachine/features/subscriber/distance_counterparts.py
```diff
@@ -120,6 +121,10 @@ def __init__(
 
         super().__init__()
 
+    @property
+    def column_names(self) -> List[str]:
+        return ["subscriber", f"distance_{self.statistic}"]
```
maxalbert: Another note to self: we should work towards removing these types of column names that depend on the parameter name. It would simplify a lot of code if the second column here (and in other classes) was simply named `value`. It would also make the life of the caller much easier because they don't need a case distinction in their code to consume the resulting table. (After all, they should _know_ which statistic they requested.)

flowmachine/flowmachine/features/utilities/feature_collection.py
```diff
@@ -82,53 +82,50 @@ class FeatureCollection(Query):
...
+    See Also
+    --------
+    feature_collection
+    """
+
+    metrics = [c(*args, **kwargs) for c in classes]
+    return feature_collection(metrics, dropna=dropna)
+
+
+# Private method that joins multiple queries together
```
maxalbert: s/method/function/`?

Also, not sure if name mangling is any issue here but maybe replace the leading double-underscore with a single one?

flowmachine/flowmachine/features/utilities/subscriber_locations.py
```diff
@@ -79,7 +80,7 @@ def _make_query(self):
         return sql
 
 
-class BaseLocation(Query):
+class BaseLocation(Query, metaclass=ABCMeta):
```
maxalbert: What's the point in making this an abstract base class if it doesn't contain any abstract methods?

flowmachine/tests/functional_tests/conftest.py
```diff
@@ -1,31 +1,9 @@
...
-    con = flowmachine.connect()
-    yield con
-    reset_cache(con)
-    con.engine.dispose()  # Close the connection
-    Query.redis.flushdb()  # Empty the redis
-    del Query.connection  # Ensure we recreate everything at next use
-
-
-@pytest.fixture
-def get_dataframe(flowmachine_connect):
```
maxalbert: I assume these can be removed because they already exist in `conftest.py` in the parent directory? Good catch!

flowmachine/flowmachine/features/utilities/feature_collection.py
greenape: It really just wraps a bunch of calls to Join, and doesn’t really have any relevant state of its own that means it _needs_ to be a class, is my thinking here.

flowmachine/flowmachine/features/spatial/location_cluster.py
greenape: This class needs a refactor, and probably splitting into at least three I think.

flowmachine/flowmachine/features/utilities/feature_collection.py
maxalbert: Makes sense. I just saw that because of the conversion the `Returns` section is missing in the docstring. Can we add this please?

flowmachine/flowmachine/features/utilities/subscriber_locations.py
greenape: Because it doesn’t define a column_names prop. But, looking at it again there’s no real reason for it to inherit from Query - could just be used as a mixin.

flowmachine/flowmachine/features/spatial/distance_matrix.py
greenape: lon & lat will always be in the cols returned by `get_columns_for_level` because the only permitted levels include them. This slightly odd dance is to make sure the list returned is in the right order.

This is another class that could use a refactor really - aside from anything else, much of the logic could be replaced by passing in a `VersionedInfrastructure` query.

flowmachine/flowmachine/core/custom_query.py
```diff
@@ -33,13 +37,17 @@ class CustomQuery(Query):
...
-    def __init__(self, sql):
-        """
-
-        """
-
+    def __init__(self, sql: str, column_names: Union[List[str], Set[str]]):
         self.sql = pretty_sql(sql)
+        seen = {}  # Dedupe the column names but preserve order
+        self._column_names = [
+            seen.setdefault(x, x) for x in column_names if x not in seen
```
maxalbert: Crikey, that's a lot of cheeky abuse of functionality in a single line... 😂

maxalbert: No strong opinion, but I feel this sort of thing is often useful in a helper function - partly for reuse but mostly for readability. I like how marshmallow collects a few useful ones in their [utils module](), for example `is_iterable_but_not_string ()`. Naming this `dedupe_list_while_preserving_order()` would both eliminate the comment and allow the reader to not have to worry about the implementation. But again, no strong opinion if you want to keep this inline here.

flowmachine/flowmachine/core/query.py
```diff
@@ -265,11 +265,18 @@ def column_names(self) -> List[str]:
...
-        except AttributeError:
-            self._cols = self.head(0).columns.tolist()
-            return list(self._cols)
+        pass
+
+    @property
+    def column_names_as_string_list(self) -> str:
+        """
+        Get the column names as a comma separated list
+        Returns
```
maxalbert: Blank line before Returns (but not after `str` below)?

flowmachine/flowmachine/features/spatial/distance_matrix.py
maxalbert: So why do we need the try/except statement if `cols` (obtained from `get_columns_for_level`) will always contain `lat` and `lon`? Under what conditions can the ValueError actually occur?

greenape: Hah. Good point.

flowmachine/tests/conftest.py
```diff
@@ -162,7 +162,8 @@ def clean_env(monkeypatch):
 @pytest.fixture
 def get_dataframe(flowmachine_connect):
     yield lambda query: pd.read_sql_query(
-        query.get_query(), con=flowmachine_connect.engine
+        f"SELECT {', '.join(query.column_names)} FROM ({query.get_query()}) _",
```
maxalbert:
```suggestion
        f"SELECT {query.column_names_as_string_list} FROM ({query.get_query()}) _",
```

PR #426, opened 2019-02-27
Black pre commit hook

### LINKED ISSUE #425: Add pre-commit hook for black
FlowKit's code is formatted with black. There is editor support for this, but it would be good to add a [pre-commit hook](https://black.readthedocs.io/en/stable/version_control_integration.html?highlight=pre-commit) so that (a) this is enforced before the CI runs and (b) we don't rely on editor support, which can be a little flaky sometimes.

maxalbert:
Closes #425 

### I have:

### DIFF
--- a/docs/source/dev_environment_setup.md
+++ b/docs/source/dev_environment_setup.md
@@ -1,5 +1,51 @@
 # Setting up a development environment
 
+## Prerequisites
+
+### Pre-commit hook for Python formatting with black
+
+FlowKit's Python code is formatted with [black](https://black.readthedocs.io/en/stable/) (which is included in the
+FlowKit pipenv environments). There is also a [pre-commit](https://pre-commit.com/) hook which runs black on all Python
+files before each commit.
+
+To install the pre-commit hook run:
+```bash
+$ cd flowmachine && pipenv run pre-commit install
+```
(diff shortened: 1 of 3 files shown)


docs/source/dev_environment_setup.md
```diff
@@ -1,5 +1,51 @@
...
+
+### Pre-commit hook for Python formatting with black
+
+FlowKit's Python code is formatted with [black](https://black.readthedocs.io/en/stable/) (which is included in the
+FlowKit pipenv environments). There is also a [pre-commit](https://pre-commit.com/) hook which runs black on all Python
+files before each commit.
+
+To install the pre-commit hook run:
+```bash
+$ cd flowmachine && pipenv run pre-commit install
```
greenape: Don't we want pre-commit to be available/used in the other bits as well?

maxalbert: It will be installed and available globally, but the `pre-commit` package is only available in the flowmachine Pipfile, so in order to run the `pre-commit install` command you have to switch to that pipenv environment (since we don't have a global one). :-/

I should add an explanation for this though, good point! 👍

maxalbert: Or do you have an idea for a better solution that avoids using a specific pipenv environment?

PR #433, opened 2019-02-28
Combine utils files

### LINKED ISSUE #422: Too many utils
FlowMachine currently has _two_ utils modules in different places:

flowmachine.utils and flowmachine.core.utils, which should be merged.

(There also flowmachine.features.utilities, which is basically a 'misc' submodule for query classes with no more obvious home, but that's a slightly different issue).

greenape:
Closes #422 

### I have:

### Description

Merges flowmachine.core.utils and flowmachine.utils.utils into just flowmachine.utils.


flowmachine/flowmachine/features/spatial/distance_matrix.py
```diff
@@ -9,12 +9,10 @@
 """
 from typing import List
 
-from flowmachine.utils.utils import get_columns_for_level
+from flowmachine.utils import get_columns_for_level
```
maxalbert: Just to check: do we generally recommend absolute imports over relative ones? Would be good to document this in the CONTRIBUTING docs because relative imports are all over the place, so if we have a clear recommendation we can gradually change these whenever we touch some code that changes imports (as is the case here).

Would also be good to have a reference for this. A quick google only seems to bring up pages which discourage from the old-style (Py2) relative imports.

greenape: Yeah, relative imports in py3 are 'ok', and the pep that used to say never use them no longer says that. Guido still doesn't like them though.

Obviously the sweeping change here is the result of an automated refactor, and pycharm doesn't like relative imports either.

I've previously felt that relative is easier to read, but I'm less sure about that now, and would be inclined to gradually switch to absolute.

maxalbert: Sounds good, especially if editors do this automatically. I've previously switched imports back to relative that were auto-changed by PyCharm and it is a pain. 😂 Do you happen to know how other editors handle this, in particular Visual Studio?

greenape:
The conf.py is deliberate - this dates from when we used sphinx to build the API docs. Not sure what it is still doing there. The empty utils.py, I forgot to delete :)

PR #434, opened 2019-02-28
Replace Redis locking with a query state machine

### LINKED ISSUE #283: Redis lock racing
In some scenarios, it is possible to exhaust the number of available threads used for managing query running. Once this happens, any calls to `store` will not actually start their internal thread running, and as a result, won't acquire a lock for that query. 

This means that any methods which should block while a store is running won't (e.g. `get_query` will return the result of `_make_query`, when it should block and eventually return a select from the cache table). In addition, server.py will incorrectly report the query as awol, because it hasn't finished running, but also doesn't have a lock which is used to indicate that the query is currently being run.

(#282 will mitigate this, but doesn't actually solve it.)

greenape:
Closes #283

### I have:

### Description

Previously we used a quasi-reentrant lock to prevent executing a query on FlowDB if it was already in the process of being executed. In the ancient times, this was implemented as a thread based rlock, but in living memory this has used the redis locking algorithm.

This doesn't quite suffice, because the lock is only acquired once a thread is available to start running the query, which, combined with the 1:1 relationship between the number of threads and the number of database connections means this would happen fairly quickly.

In FlowMachine library terms, this as shown in #283, but could equally result in a query composing a subquery re-executing the subquery when it should actually wait for the subquery to be written to cache. In FlowMachine _server_ terms, this means that with n_threads queries currently running, any additional query submitted to the server will appear to be awol until one currently running finishes.

This replaces the lock with a state machine, which actually runs atomically _in redis_. Queries can be in one of several states, some of which indicate that you can safely get their SQL, and some of which indicate that you should wait to do so because it is likely to change.

 #283 is hence mitigated, since nothing ever has to wait for a lock and hence marking a query as running can safely be done in the main thread. It also has some additional benefits, in that one can now get a more accurate view of a query's current state. This includes whether it is actively running, or just expected to at some point, whether it is being wiped from cache right now, if it failed to run successfully (even if somebody else last tried to run it), and so on. 

(Looking forward to some debate on this one ;))


maxalbert:
Will need to take some time to look at this in detail, but the first thing I notice is that it includes [finist](https://pypi.org/project/finist/) as a dependency, which has only had two (pre-) releases in 2015(!) and no activity since. It also doesn't allow to create any issues on the Github repo. So I'd be very hesitant to pull this in as an external dependency.

On the upside, it basically only contains a single small class (see [here](https://github.com/sumanau7/finist/blob/master/finist/finist.py)) and is MIT licensed, so I could be convinced to include this in our own code if we wanted, so that we can ensure it is properly tested for our use case and we can adapt it as needed.

Have you looked whether there are any more recent and actively maintained packages around?

maxalbert:
Only had a quick play with the `QueryStateMachine` class so far, but my first impression is that I like it! :)

Here are a bunch of random initial observations. There are quite a few but they are pretty much exclusively suggestions for clarification or better names. In principle this looks good!

More comments on the PR to come later. ;)

- What is the difference between the states "known" and "queued"? The docstring in `QueryStateMachine` doesn't really make this clear.

  For "queued", it says that it "indicates that a query is going to be executed in future", but this should also be the case if it is "known". What is special about the "queued" state and what exactly causes the transition from known to queued?

- Can we add docstrings for methods in `QueryStateMachine`? Some are pretty self-explanatory (but good to have them anway), but some are not entirely clear. In particular, for the `trigger_event` method it's not entirely clear what the return values are.

- In `QueryStateMachine.trigger_event`, I'm a little surprised that the `trigger_success` return value will be `False` if the same event is triggered twice, even thought the query will remain in the desired state:
  ```python
  >>> state_machine = QueryStateMachine(redis, "DUMMY_QUERY_ID")
  >>> print(state_machine.enqueue())
  (<QueryState.QUEUED: 'queued'>, True)
  >>> print(state_machine.is_queued)
  True
  >>> print(state_machine.enqueue())  # Note that "success" return value is "False"...
  (<QueryState.QUEUED: 'queued'>, False)
  >>> print(state_machine.is_queued)  # ... but the query is still queued.
  True
  ```

- It wasn't clear to me before trying it out that the query is immediately put into the "known" state when creating a `QueryStateMachine`. This should be clarified in the docstring.

- It wasn't obvious to me that the state of a query would persist even when creating a new `QueryStateMachine` instance with the same query_id. In hindsight this is obvious (because the class just facilitates interaction with Redis in a structured way, so Redis persists the state), but would be good to mention this in the docstring.

- Is it easy to produce a Diagram for the transitions? Don't worry about it if not (can always add this later), but if it is then it would be great to include it in the docs.

- `is_executed` should only return True if the query was executed _successfully_ (because this is what the `executed` state represents, according to the docstring in `QueryStateMachine`). Consequently, the state `is_executed_without_error` should be removed because it is redundant. I can see the usefulness of a state that indicates that the query isn't running any more but I'd prefer a different name for this (see below).


Naming:

- The `resetting` state doesn't convey anything about what is being reset and why. From the docstring and implementation it says that it is inherently tied to the cache, can we reflect this in the name?

- The property `errored` should be renamed to `is_errored` (because all other properties start with `is_`)

- I'd be in favour of renaming the `executed` state to `completed` (and rename `is_executed` to `is_completed`) because it has more of a connotation of success.

- I can see the usefulness of a property which indicates whether the query was set running and now isn't running any more, but I think we want a differently named property for this than `is_executed`. Maybe `is_stopped_running` (or `is_not_running_any_more` if we want to be verbose)? Alternatively, we could use the existing `has_finished_operating` for this (see next item).

- The name of the function `has_finished_operating()` is fairly misleading. From the name I would expect it to return immediately with a True/False return value, but it actually blocks until some condition is met. This should definitely be reflected in the name. I'm also not sure if I like the blocking behaviour at all, but I'll defer that judgement until I have had a chance to look at how it's being used.

maxalbert:
P.S. re naming:

- I was confused by the name `finish_reset`. Would you be happy to rename this to `finish_resetting` (if this is what it does)? (Also, I need to understand these states better, but that's for later...)

- How about renaming `error` to `raise_error`? With all the other actions it's clear that they are verbs, but this one is ambiguous.

greenape:
State diagram:

screenshot 2019-02-28 at 12 43 41

```
graph LR
A[KNOWN] -->|ENQUEUE| B[QUEUED]
B -->|EXECUTE| C[EXECUTING]
C -->|ERROR| D[ERRORED]
C -->|FINISH| E[COMPLETED]
C -->|CANCEL| F[CANCELLED]
B -->|CANCEL| F[CANCELLED]
D -->|RESET| G[RESETTING]
E -->|RESET| G[RESETTING]
F -->|RESET| G[RESETTING]
G -->|FINISH_RESET| A
H[Blocking]
style B fill:#f9f
style C fill:#f9f
style G fill:#f9f
style H fill:#f9f
```

(There's no immediately obvious place to include this in the docs, but I do think it would be a good idea to.)

maxalbert:
Very nice! Love the state diagram. Agreed there doesn't currently seem to be an obvious place in the docs. Maybe add it to the Developer docs in a new subsection of the flowmachine section related to "architecture" or "design" or "internals"?

maxalbert:
The package `finist` seems to be missing from `docs/Pipfile` (it's only present in `docs/Pipfile.lock`)?

flowclient/tests/unit/test_get_status.py
```diff
@@ -4,15 +4,19 @@
...
+import pytest
+
 from flowclient.client import get_status
 
 
-def test_get_status_reports_running():
-    """ Test that status code 202 is interpreted as query running. """
+@pytest.mark.parametrize("running_status", ["queued", "executing"])
+def test_get_status_reports_running(running_status):
+    """ Test that status code 202 is interpreted as query running or queued. """
```
maxalbert: Is this test really testing anything? Aren't we just comparing `running_status` to itself because we're setting the return value on the mock to it? Where does the `status_code` come into play at all?

flowmachine/flowmachine/core/model_result.py
```diff
@@ -150,25 +152,50 @@ def to_sql(
...
+            if this_thread_is_owner:
                 logger.debug("Obtained storage lock.")
                 con = self.connection.engine
-                if force:
-                    self.invalidate_db_cache(name, schema=schema)
                 try:
                     with con.begin():
                         logger.debug("Using pandas to store.")
                         self._df.to_sql(name, con, schema=schema, index=False)
+                        # Mark as finished before writing meta to avoid getting blocked when
```
maxalbert:
```suggestion
                        # Mark as finished before writing cache metadata to avoid getting blocked when
```

flowmachine/flowmachine/core/model_result.py
```diff
@@ -150,25 +152,50 @@ def to_sql(
 
         def do_query() -> ModelResult:
             logger.debug("Getting storage lock.")
```
maxalbert: Should this line be removed now that we're not using an explicit lock any more?

flowmachine/flowmachine/core/model_result.py
```diff
@@ -150,25 +152,50 @@ def to_sql(
...
-            with rlock(self.redis, self.md5):
+            q_state_machine = QueryStateMachine(self.redis, self.md5)
+            current_state, this_thread_is_owner = q_state_machine.execute()
+            if this_thread_is_owner:
                 logger.debug("Obtained storage lock.")
```
maxalbert: Does this line about the storage lock actually still apply?

flowmachine/flowmachine/core/model_result.py
```diff
@@ -150,25 +152,50 @@ def to_sql(
...
+
+            if q_state_machine.is_completed:
+                return self
+            elif q_state_machine.is_cancelled:
+                logger.error(f"Model result write '{self.md5}' was cancelled.")
+                raise QueryCancelledException(self.md5)
+            elif q_state_machine.is_errored:
+                logger.error(f"Model result write '{self.md5}' finished with an error.")
+                raise QueryErroredException(self.md5)
+                logger.debug("Released storage lock.")
```
maxalbert: Where exactly is the storage lock released, and why does this only happen in the `is_errored` branch?

flowmachine/flowmachine/core/model_result.py
```diff
@@ -150,25 +152,50 @@ def to_sql(
...
                         "No dataframe to store, presumably because this"
                         " was retrieved from the db."
                     )
-            logger.debug("Released storage lock.")
+            elif q_state_machine.is_executing:
+                logger.debug(
+                    f"Model result '{self.md5}' being written elsewhere, waiting for it to finish."
+                )
+                while q_state_machine.is_executing:
+                    sleep(5)
```
maxalbert: Could make this poll frequency an optional kwarg to `do_query()`? Maybe not important, but somehow it feels odd to have an arbitrary hard-coded number of seconds here that the call will be blocked for without the user being able to change it.

flowmachine/flowmachine/core/query_state.py
```diff
@@ -0,0 +1,346 @@
...
+        tuple of QueryState, bool
+            Returns a tuple of the new query state, and a bool indicating whether the caller
+           marked the reset as complete with this call.
+
+        """
+        return self.trigger_event(QueryEvent.FINISH_RESET)
+
+    def wait_until_complete(self):
+        """
+        Blocks while the query is in any state that makes how to get
```
maxalbert: This docstring is a little difficult to understand. How about rephrasing it as follows?
```
Blocks until the query is in a state where its result is determinate (i.e., one of "know", "errored", "completed", "cancelled").
```

flowmachine/flowmachine/core/query_state.py
```diff
@@ -0,0 +1,346 @@
...
+        Returns
+        -------
+        tuple of QueryState, bool
+            Returns a tuple of the new query state, and a bool indicating whether the caller
+           marked the reset as complete with this call.
+
+        """
+        return self.trigger_event(QueryEvent.FINISH_RESET)
+
+    def wait_until_complete(self):
```
maxalbert: Not a big deal, but `wait_until_complete()` sounds suspiciously like "wait until query is in the 'completed'" state, which is slightly misleading because it can end up in other states, too. I can't think of a better name but if you have an idea feel free to rename it.

flowmachine/flowmachine/core/query.py
```diff
@@ -190,7 +198,7 @@ def get_query(self):
         try:
             table_name = self.fully_qualified_table_name
             schema, name = table_name.split(".")
-            with rlock(self.redis, self.md5):
+            if QueryStateMachine(self.redis, self.md5).wait_until_complete():
```
maxalbert: This is confusing to the reader because `wait_until_complete()` sounds like an instruction which I wouldn't expect to return anything. Can we split this into two separate instructions?
```suggestion
            query_state_machine = QueryStateMachine(self.redis, self.md5)
            query_state_machine.wait_until_complete()
            if query_state_machine.is_completed:
```

flowmachine/flowmachine/core/query_state.py
```diff
@@ -0,0 +1,346 @@
...
+        """
+        Blocks while the query is in any state that makes how to get
+        the result of it indeterminate (i.e. currently running, resetting,
+        or scheduled to run).
+
+        Returns
```
maxalbert: See comment above - it's a bit confusing (at least in Python land) that a function called `wait_until_complete()` has a return value. Can we remove this return value, since the user can just as easily check the `is_completed` state themselves, which is also clearer to read?

flowmachine/flowmachine/core/query_state.py
```diff
@@ -0,0 +1,346 @@
...
+        -------
+        bool
+            True if the query has been executed successfully and is in cache.
+        """
+        if self.is_executing or self.is_queued or self.is_resetting:
+            while not (
+                self.is_finished_executing or self.is_cancelled or self.is_known
+            ):
+                _sleep(1)
```
maxalbert: See above, worth making this poll frequency a parameter?

flowmachine/flowmachine/core/query.py
```diff
@@ -714,6 +729,7 @@ def _db_store_cache_metadata(self, compute_time=None):
                     ),
                 )
                 con.execute("SELECT touch_cache(%s);", self.md5)
+                con.execute("SELECT pg_notify(%s, 'Done.')", self.md5)
```
maxalbert: Out of interest, where do these notifications end up? In the postgres logs, or in the flowmachine logs, or both/elsewhere?

flowmachine/flowmachine/core/query.py
```diff
@@ -503,20 +503,21 @@ def to_sql(
 
         def do_query() -> Query:
             logger.debug("Getting storage lock.")
-            with rlock(self.redis, self.md5):
+            q_state_machine = QueryStateMachine(self.redis, self.md5)
```
maxalbert: There is a lot of repetition between the `do_query()` methods in the `ModelResult` and `Query` classes. Is this an indication that this functionality should live somewhere else where it can be shared?

flowmachine/flowmachine/core/query.py
```diff
@@ -862,6 +885,15 @@ def invalidate_db_cache(self, name=None, schema=None, cascade=True, drop=True):
             logger.debug("Dropping {}".format(full_name))
             with con.begin():
                 con.execute("DROP TABLE IF EXISTS {}".format(full_name))
+            q_state_machine.finish_resetting()
+        elif q_state_machine.is_resetting:
+            logger.debug(
+                f"Query '{self.md5}' is being reset from elsewhere, waiting for reset to finish."
+            )
+            while q_state_machine.is_resetting:
```
maxalbert: Hmm, I'm not fully convinced we want to block here. I can see it being useful _if_ it is guaranteed that once we return the table has actually been removed, but I don't think we're checking that this has successfully happened So why not return immediately since we know the table is being dropped somewhere else?

flowmachine/flowmachine/core/query.py
greenape: Hah. That should be removed - holdover from a previous angle of attack.

The notify stuff in Postgres is quite cool though! You can register as a listener on a channel, and then receive asynchronously any notifications sent on that channel.
Would be _more_ useful, if not for the fact that notifications have transaction level isolation, so any sent inside a transaction aren't delivered until after it completes.

flowmachine/flowmachine/core/model_result.py
greenape: Good point. Will change.

maxalbert:
Looks pretty good now! I added a few more inline comments, let me know when you have addressed them (or if they are not applicable) and I'll have another final look.

flowmachine/flowmachine/core/cache.py
```diff
@@ -23,6 +34,168 @@
...
+        con = connection.engine
+        with con.begin():
+            try:
+                plan_time = write_func(query_ddl_ops, con)
+                logger.debug("Executed queries.")
+            except Exception as e:
+                q_state_machine.raise_error()
+                logger.error(f"Error executing SQL. Error was {e}")
+                raise e
+            if schema == "cache":
```
maxalbert: Is there a reason to only conditionally write cache metadata? I'm a little confused about the meaning of this `schema` argument. I assumed its purpose was to allow a different name for the cache schema (e.g. for testing), but it looks like it's passed in from `Query.to_sql()` and `ModelResult.to_sql()` and is basically the schema that the queries are stored in? I trust that it does the right thing but I'm not sure I fully follow the logic.

flowmachine/flowmachine/core/server/query_proxy.py
```diff
@@ -510,11 +504,32 @@ def get_sql(self):
...
-            else:
-                return get_sql_for_query_id(query_id)
-        except (AttributeError, ValueError):
+            query_state = self.redis_interface.query_status(query_id)
+        except ValueError as e:
+            raise QueryProxyError(
+                f"Got a bad state for '{query_id}'. Original exception was {e}"
+            )
+
+        if query_state == QueryState.EXECUTING:
```
maxalbert: Just as a thought, I wonder whether it would be useful to have the error messages live in QueryState directly and simply propagate them here? Would allow us to get rid of the if/elif chain. But haven't looked into it deeply, it's possble this wouldn't work out.

flowmachine/tests/server/test_query_proxy.py
```diff
@@ -168,10 +176,60 @@ def dummy_construct_query_object(query_kind, params):
...
     # Get SQL code and check it is as expected
     #
     sql = query_proxy.get_sql()
     assert "SELECT * FROM dummy_table" == sql
+
+
+@pytest.mark.parametrize(
+    "current_state, expected_error",
+    [
+        (QueryState.KNOWN, MissingQueryError),
```
maxalbert: Similar to my comment above, I'm wondering whether it's worth always raising the same error. I'll keep that in mind for my re-working / removal of the QueryProxy class.

maxalbert:
@greenape FYI, I have added a couple more commits on top  of yours, but literally only cosmetics.

PR #443, opened 2019-03-04
Fixups

### LINKED ISSUE #189: Different environment variables with similar names and same(?) purpose are confusing
There are some cases where multiple environment variables are used that appear to control the same thing (although I'm not entirely sure there aren't subtle differences) but have different names. 

The example that I keep stumbling across is `DB_PORT` vs. `FLOWDB_PORT`. Do these have the same purpose? If not, we should clarify and document their meaning and the differences. If yes, can we merge them and only keep one (e.g. the more explicit `FLOWDB_PORT`)?

There may be other environment variables where a similar overlap exists. In any case, it would be helpful to document all relevant environment variables and their meaning.

greenape:
Closes #189

### I have:

### Description

Various minor fixups.

- All usages of `DB_PORT` are now `FLOWDB_PORT`
- Flowmachine docker build no longer relocks the Pipfile during build
- Integration tests now spin up flowmachine and flowapi using pytest. This means they no longer need docker, and are a. faster and b. we can collect coverage of flowapi and flowmachine during them
- Build order on CI has changed to allow more things to run in parallel.

In combination, these changes reduce the CI build by about _10_ minutes.


docker-compose-dev.yml
```diff
@@ -19,7 +19,7 @@ services:
       context: ./flowdb/
       dockerfile: Dockerfile
     ports:
-      - ${FLOWDB_PORT:-9000}:5432
+      - ${FLOWFLOWDB_PORT:-9000}:5432
```
maxalbert: Overeager auto-search/replace? ;)

flowmachine/Dockerfile
```diff
@@ -8,10 +8,10 @@ WORKDIR /app
 COPY Pipfile* /app/
 RUN pip install pipenv && pipenv install --deploy
 COPY . /app/
-RUN pipenv install .
+RUN pipenv run pip install .
```
maxalbert: Why do you prefer to install flowmachine via pip run through pipenv, instead of just plain pipenv?

greenape: Just to record the rationale here - we should not be relocking in the dockerfile, because the build won't be what we've just tested, and it bloats the resulting image massively. Skipping the lock is fine, because we've explicitly installed flow machine's dependencies.

integration_tests/tests/conftest.py
```diff
@@ -2,17 +2,61 @@
...
+    import sys
+
+    sys.path.insert(
+        0,
+        os.path.abspath(os.path.join(os.path.dirname(__file__), "..", "..", "flowapi")),
+    )
+
+    api_thread = Process(
+        target=hypercorn.__main__.main,
+        args=(["--bind", "0.0.0.0:9090", "app.main:create_app()"],),
```
maxalbert: Is this hard-coded port going to conflict with any other instances of flowapi that are also running? How about reading the port from an environment variable (e.g. `FLOWKIT_INTEGRATION_TESTS_FLOWAPI_PORT` or similar)? (See the `zmq_port` fixture below.)

Similarly for the `flowmachine_server` above - this will start on port 5555 and might conflict with an existing flowmachine server, right?

greenape: Good point. I think actually the right thing to do here is to not specify them and use something similar to pytest-asyncio's [unused_tcp_port](https://github.com/pytest-dev/pytest-asyncio#unused_tcp_port) fixture to specify + try to bind on [port 0](https://www.dnorth.net/2012/03/17/the-port-0-trick/) in the actual code. But I'm gonna spin that off into an issue I think.

Issue #351, opened 2019-02-07
Refactor logging (and ensure documentation and tests are complete)


maxalbert:
Currently the logging functionality exists in a few places and uses slightly different mechanisms. We should move towards using `structlog` everywhere and set everything up from the same place to ensure it is consistent.

We should also ensure that everything is fully tested (e.g. to avoid glitches if `LOG_DIRECTORY` is set to a non-default location; not sure if there can be problems if that location is outside the container and/or not writeable?) and it is clearly documented how to access and post-process the logs if needed.

maxalbert:
Closing this because the first part has been addressed. The remaining issues are tracked in #468 and #469.

PR #460, opened 2019-03-07
Structured logs

greenape:
Against #351

### I have:

### Description

This is for #351, but I think doesn't quite close it out.

- FlowMachine's debug and error logs are now structlog!
- FlowAPI's debug logs are now structlog!!
- All JSON writing is now a tiny bit faster*!!! (This is a pre-emptive optimisation to avoid hitting issues when under a lot of load, which I think would be evident with the very slow python json serialiser)
- Structlog is now configured once, in `flowmachine.__init__` which _should_ mean it is inescapably always configured.

*This is due to switching to `python-rapidjson`. Previously we used a mix of `ujson` and the stdlib `json` module. `ujson` isn't compatible with structlog, and `python-rapidjson` is faster than `ujson` and _much_ faster than `json`.


docs/source/5-administrator.md
```diff
@@ -60,15 +60,18 @@ These values can be overridden when creating a new FlowDB container by setting t
 
 [^1]:By default, this uses the value set for `cache_size` in `cache.cache_config`.
 
-<a name="logging">
 ## Logging and Audit Trails
 
-FlowKit supports structured logging in JSON form for all interactions with FlowAPI, primarily to facilitate audit trails.
+FlowKit supports structured logging in JSON form throughout FlowMachine and FlowAPI. By default both FlowMachine and FlowAPI will only log errors, and audit logs.
+
+Audit trail logs are _always_ written to both stdout and a log file. Error logs are, by default, only written to stdout.
```
maxalbert: Should error logs be written to stderr instead of stdout?

docs/source/5-administrator.md
```diff
@@ -234,6 +237,14 @@ FlowMachine usage log:
 
 Note that the `request_id` field is identical across the five log entries, which lets you match the request across the multiple services.
 
+### Error and Debugging Logs
+
+By default, FlowMachine and FlowAPI write error logs to stdout. For more verbose logging, set the `LOG_LEVEL` environment variable to `debug` when starting the docker container.
+
+FlowMachine also supports writing debug and error logs to a file. To enable this, set the `WRITE_LOG_FILE` environment variable to `TRUE` when starting the FlowMachine container. Logs will be written to `$LOG_DIRECTORY/flowmachine-debug.log` (by default `LOG_DIRECTORY=/var/log/flowmachine`). If the directory specified does not exist, FlowMachine will attempt to create it.
```
maxalbert: Just to be extra explicit, we should mention that this directory lives inside the docker container, e.g.: "Logs will be written to `$LOG_DIRECTORY/flowmachine-debug.log` inside the docker container (by default `LOG_DIRECTORY=/var/log/flowmachine`)."

Come to think of it, why do we need to allow configuring this directory at all if it's inside the container? Wouldn't it be simpler and more predictable to always write to the same directory? The user can always mount a custom host directory at `/var/log/flowmachine` if they also want to store the logs outside the container.

docs/source/5-administrator.md
```diff
@@ -234,6 +237,14 @@ FlowMachine usage log:
...
+
+Log messages from FlowMachine will show the `logger` field of the log entry as the Python module that emitted the log entry (e.g. `{'logger':'flowmachine.core.query'..}`. FlowAPI debugging messages set `logger` to `flowapi`.
```
maxalbert: For consistency with flowapi, replace `logger` with `flowmachine` in the flowmachine logs?

greenape: I think the use case is a little different though - in flowmachine, the logger tells you which class wrote the message, useful for debugging. That's not really relevant in flowapi, cos the debug logger is a global app wide one.

docs/source/5-administrator.md
jch1g10: Is there any situation in which we would ever want to run flowmachine outside of a container? If so, it might be necessary to change the log directory in that situation.

greenape: Fine with mentioning explicitly that this is inside the container, but I think worth leaving it configurable because otherwise it makes it a pain should you wish to run it _outside_ a container. Not our presumed deployment scenario, but a totally reasonable thing to do (and indeed, we do it ourselves in the tests).

flowapi/flowapi/main.py
```diff
@@ -42,7 +42,12 @@
 async def connect_logger():
     log_level = current_app.config["LOG_LEVEL"]
     log_root = current_app.config["LOG_DIRECTORY"]
-    current_app.logger.setLevel(logging.getLevelName(log_level))
+    logger = logging.getLogger("flowapi")
+    logger.setLevel(log_level)
+    ch = logging.StreamHandler()
+    ch.setLevel(log_level)
+    logger.addHandler(ch)
+    current_app.log = structlog.wrap_logger(logger)
```
maxalbert: This may just be personal preference, but I find `current_app.logger` slightly more readable than `current_app.log` (perhaps because it doesn't have the noun vs verb ambiguity). Any objections to changing it back to `current_app.logger`?

docs/source/5-administrator.md
maxalbert: Oh sorry, ignore me. I misunderstood what this sentence was trying to explain...

Do we need to update the "Complete Logging Cycle" example then? I'm not sure the `logger` fields in this example represent what is being explained here?

flowapi/flowapi/main.py
greenape: Can't, quart has special rights over `app.logger`. Could change to `debug_logger`?

docs/source/5-administrator.md
greenape: Personally, I don't think so. In 'normal' operation, you should never see debugging messages, and they aren't relevant to following the execution for audit purposes.

flowapi/flowapi/main.py
maxalbert: Ah, I see. Perhaps `flowapi_logger` then (since we may not just be logging debug messages)?

As an aside, this seems like a possibility for a trip-up. I was unaware that quart uses `app.logger` internally. Probably good to add a comment to that effect in `flowapi/main.py` where we set up the logger?

flowmachine/flowmachine/core/init.py
```diff
@@ -216,14 +219,10 @@ def _init_logging(log_level, write_log_file):
     except (AttributeError, TypeError):
         log_level = logging.ERROR
     true_log_level = logging.getLevelName(log_level)
-    formatter = logging.Formatter(
```
maxalbert: Not a big deal, but what's the rationale for removing this formatter? (In particular, what do messages look like after this change?) It should only be relevant for on-screen formatting, right? Why not keep the structure of those messages as we had them before?

docs/source/5-administrator.md
```diff
@@ -60,17 +60,20 @@ These values can be overridden when creating a new FlowDB container by setting t
 
 [^1]:By default, this uses the value set for `cache_size` in `cache.cache_config`.
 
-<a name="logging">
 ## Logging and Audit Trails
 
-FlowKit supports structured logging in JSON form for all interactions with FlowAPI, primarily to facilitate audit trails.
+FlowKit supports structured logging in JSON form throughout FlowMachine and FlowAPI. By default both FlowMachine and FlowAPI will only log errors, and audit logs.
+
+Audit trail logs are _always_ written to both stderr and a log file. Error logs are, by default, only written to stderr.
```
maxalbert: Thanks for changing this to `stderr`. Just to check, does structlog automatically output to stderr? I don't see anywhere in the code that we explicitly configure this (and couldn't find it in the structlog docs at a glance).

greenape: iirc by default, it goes to stdout, but we're wrapping stdlib's loggers so it goes where they do (stderr).

flowmachine/flowmachine/core/init.py
greenape: Makes the messages easier to parse. Without removing the formatter, you get the structlog message jammed after the stuff the formatter puts in. So, if you want to be able to parse all the logs you'd then need to do special handling to be able to parse any debugging ones. Sans formatter, every log message is a valid json object.

maxalbert: 👍

PR #487, opened 2019-03-16
Rename env vars for consistency

### LINKED ISSUE #485: Inconsistent use of environment variable name for connecting with FlowDB
**Describe the bug**
The [`connect`](https://github.com/Flowminder/FlowKit/blob/314caa4a19ccb480abed072a07dd76a1ddbe5ecb/flowmachine/flowmachine/core/init.py) method of `flowmachine` will read the following environment variables to connect to the DB:

- `FLOWDB_PORT`
- `DB_USER`
- `DB_PW`
- `DB_HOST`
- `DB_NAME`
- ...

Now I find it inconsitent to have `FLOWDB_PORT` and the other parameters prefixed with `DB`. Is there any reasons for that? 

**Product**
Flowmachine

**Version**
master branch

**Expected behavior**
Can we perhaps be consistent and use `FLOWDB` as the prefix?

maxalbert:
Closes #485 

### I have:

### Description

As pointed out in #485, the current environment variable names are not very consistent, and sometimes it is not fully obvious what they mean or there is ambiguity.

I also feel it is good to be explicit in the names to avoid conflicts with env vars used by other software (e.g. `DB_USER` might clash, whereas `FLOWDB_USER` won't).

This PR makes the following renamings to improve consistency, understandability and potential clashes:

- `DB_USER` -> `FLOWDB_USER`
- `DB_USER` -> `FLOWDB_HOST`
- `DB_PASS` -> `FLOWDB_PASS`
- `DB_PW` -> `FLOWDB_PASS`

- `API_DB_USER` -> `FLOWAPI_DB_USER`
- `API_DB_PASS` -> `FLOWAPI_DB_PASS`
- `FM_DB_USER` -> `FLOWMACHINE_DB_USER`
- `FM_DB_PASS` -> `FLOWMACHINE_DB_PASS`

I also removed the env var `DB_NAME` because it doesn't have any use at the moment - the database is always called `flowdb`, so there is no point in allowing to choose a different name (we can always re-introduce it if there ever is a need to change it, but for now good to keep it simple).

As an unrelated change, I have removed the parameter `strategy="threadlocal"` in an internal call to `sqlalchemy.create_engine()` because it is [deprecated](https://docs.sqlalchemy.org/en/latest/changelog/migration_13.html#change-4393-threadlocal)


maxalbert:
Actually, I had to re-add `strategy="threadlocal"` for now because removing it made the docs build fail (not sure why).

gzagatti:
@maxalbert, I would also consider renaming the `REDIS` env variables to something more meaningful such as `FLOWMACHINE_REDIS`, or something shorter.

maxalbert:
@gzagatti Good point, I'd be in favour of being explicit about this. I'll add it as a comment to #485 so we can track it there.

PR #503, opened 2019-03-25
New synth data generator

greenape:
### I have:

### Description

The primary purpose of this PR is to add an additional data generator to the synth data container. This a SQL based, in DB generator which is faster (2-3x in my tests) than the existing python one, and more efficient on disk space. Together these mean it can be used to generate _much_ larger data volumes. 
It also has some nice bonuses, with the tradeoff of not being seedable:

- Number of sites & cells can be different
- mds and sms data can be generated
- TACs are generated
- Subscribers have a 'home' region, and call from there with configurable probability
- Subscribers move regions with configurable probability
- Subscribers have a fixed set of partners they call & text with
- Mass relocation events can be simulated (everybody in a designated admin2 region moves, and nobody can enter it for a set period of time)

The PR retains the existing data generator, and allows setting which is used via an env var. Also cleans up the circle config a little, uses the synth data in the docs, and boosts the speed of the existing python generator a bit as an added bonus.


.circleci/config.yml
```diff
@@ -95,6 +40,81 @@ defaults:
...
 
+
+executors:
+  python_with_flowdb: # Synthetic data generation execution environment
+    parameters:
+      flowdb_image:
+        type: enum
+        enum: ["synthetic-data", "testdata"]
+        default: "testdata"
+      generator:
```
maxalbert: Maybe call this `synthetic_data_generator` for clarity (in line with the env var `SYNTHETIC_DATA_GENERATOR`)?

flowdb/bin/build/002_tweak_config.sh
```diff
@@ -24,5 +24,14 @@
 #
 set -e
 
+# Allows WAL to be skipped for CREATE TABLE & CLUSTER
+# we turn this on in the autogenerated config anyway, but nice
```
maxalbert:
```suggestion
# we configure this in the autogenerated config anyway, but nice
```

flowdb/bin/build/002_tweak_config.sh
```diff
@@ -24,5 +24,14 @@
...
+# to have it off from the start for the benefit of the testdata containers.
```
maxalbert:
```suggestion
# to have it off from the start for the benefit of the testdata containers
# for increased performance during startup.
```
(That's the benefit as I understand it? Might be nice adding a half-sentence to indicate why that's the case.)

Issue #270, opened 2019-01-22
Tidy up logic in claims checks


maxalbert:
The logic in the [check_claims](https://github.com/Flowminder/FlowKit/blob/master/flowapi/app/run_query.py#L13) decorator isn't always clear, and it makes some implicit assumptions that are not evident from reading this piece of code alone.

For example, the logic of the `if` branch in [L58](https://github.com/Flowminder/FlowKit/blob/master/flowapi/app/run_query.py#L58) depends on the fact that either `query_id` or `query_kind` is present, and it's not clear that if this is not the case this is correctly handled.

Would be good to tidy this up and make it easier to read and understand.

jch1g10:
It would also be good to separate the two stages of checking claims: getting the appropriate claim type, query kind etc., and checking against the token. This would simplify the `check_claims` decorator, and reduce code duplication between `check_claims` and ``check_geography_claims`.

Issue #224, opened 2019-01-10
FlowMachine should reply to malformed ZMQ messages which have a reply address


greenape:
**Describe the bug**
At present, if you send a bad message to FlowMachine via zmq (e.g. missing a top level key), then you don't get a reply, and the sender will get hung when it tries to receive. This is fine _if_ the zmq message is so wrong that a reply address can't be found, but an error message should be sent back if there _is_ a reply address.

**Product**
FlowMachine

**Version**
master

**To Reproduce**
Unskip `integration_tests.flowmachine_server_tests.test_action_run_query.test_wrongly_formatted_zmq_message`

PR #508, opened 2019-03-26
Refactor params validation and query construction

### LINKED ISSUE #507: Query construction and parameter validation should be harmonised
The flowmachine server needs to construct query objects from parameters sent via zmq messages. This query construction and the associated parameter validation currently happens in [construct_query_object()](https://github.com/Flowminder/FlowKit/blob/master/flowmachine/flowmachine/core/server/query_proxy.py#L98) in a fairly ad-hoc manner. This is not very scalable and should be harmonised to make it easier to expose additional queries, and to introduce a "single place of truth" for the validation of common parameters such as aggregation units, date periods, subscriber subsets, etc.

maxalbert:
Closes #507 and provides steps towards addressing #270, #224.

### I have:

### Description

This PR refactors the flowmachine server internals to provide a much more harmonised way of validating query parameters and constructing query objects which is less ad-hoc and more scalable.

Main changes:

- Removed the existing QueryProxy implementation and the `construct_query_object()` helper function. This is replaced with a collection of files defining marshmallow schemas for the exposed queries in `flowmachine/core/server/query_schemas/`.

- The flowmachine server was refactored and simplified. The various action handlers have been moved into `flowmachine/core/server/action_handlers.py`. Currently there is an `ACTION_HANDLERS` lookup which matches the action name to its handler function, but @greenape suggested we could use decorators to mark actions, which would allow to seamlessly add new actions.

- The API spec can now be generated automatically using the toplevel `FlowmachineQuerySchema`, which recursively expands all exposed flowmachine queries (see the `action_handler__get_query_schemas()` function in `flowmachine/core/server/action_handlers.py`). The resulting spec is verified as part of the integration test and can be inspected in 
`integration_tests/tests/flowmachine_server_tests/test_server.test_api_spec_of_flowmachine_query_schemas.approved.txt`.

- The general JSON structure for a query has slightly changed in that the parameters are now "inlined" (at the same level as the `query_kind` key), rather than explicitly nested using an additional `params` key.

Knock-on changes:

- Changing the zmq reply message structure made it necessary to adapt the claims checks in `flowapi/flowapi/check_claims.py`. It would be good for someone else to take a look that these changes make sense and I didn't accidentally introduce any bugs.

- Because we can now pass the query parameters directly to the flowmachine server and it constructs queries recursively using the marshmallow schema definitions, some code in `flowclient` could be drastically simplified (in particular, the `_meaningful_locations()` helper function was removed).

Minor changes:

- Renamed `daily_location_method` to `method`.
- I added a `DummyQuery` class which behaves like any other proper flowmachine query but doesn't store anything to the database. This is helpful in integration tests, so I added it to `flowmachine/core/` (rather than just the flowmachine tests) so that it is accessible via the zmq server from e.g. flowapi and flowclient, too. Let me know if you think there is a better place for it.
- Some of the tests used `0` as the dummy query id, which took me a while to realise. I have made this more explicit by using `DUMMY_QUERY_ID`.

A few things to note:

- With the changes in this PR, each internal flowmachine query which is exposed via the API has a wrapping `...Exposed` class in `flomwachine/core/server/query_schemas/` (e.g. `DailyLocationExposed`). This makes things easy to handle and consistent, but it adds an extra wrapping layer. I feel we will probably want to keep these `Exposed` classes, but they will likely allow us to do some cleanup of intermediate layers which we should do in subsequent PRs.

- For a few parameters (such as `subscriber_subset`, `event_types`, ...) we allow special values such as the string `"all"`. I feel we should move away from this and only accept `None` as the special value which indicates "the most general default". This is the route I have chosen for the exposed queries, but internally the special values are still used all over the place (and I had to add a couple of explicit conversions to avoid big ripple effects that would have made this PR even bigger). Again, this should be tidied up in subsequent PRs.

- It was quite difficult to debug some of the knock-on effects which the tweaks to the flowmachine server internals had on the tests. The main reason is that before these changes the JSON message structure was pretty much arbitrary, and the tests which mock out the zmq replies only specified the bits necessary to make the tests pass. This was mostly ok for reading and broadly understanding the tests but made them difficult to change because it wasn't clear which messages the mock replies were responding to. I tried to mitigate this by explicitly constructing mock replies using the new `ZMQReply` class, but currently this makes it necessary to copy the `zmq_helpers.py` module in multiple places (`flowmachine/flowmachine/core/server/zmq_helpers.py`, `flowapi/flowapi/zmq_helpers.py`, `flowclient/flowclient/zmq_helpers.py`). I suggest to create a `flowkit_common` module where this shared code can live, but I'd rather do that in a separate PR.

- Related to the previous point, currently the `ZMQReply` class uses manual validation internally to ensure it has the right structure. Once this PR is merged we should immediately move towards using marshmallow for this as well, which will eliminate most of the code in `zmq_helpers.py` and reduce it to a few lines.

- There is currently a dependency resolution issue when installing `marshmallow`, `marshmallow-oneofschema`, `apispec` and `apispec-oneofschema` using pipenv. I haven't quite figured out why this is but it is somehow related to the fact that marshmallow is on a pre-release for version 3 (which we want to use). I suspect it's simply the way in which the dependencies are specified in one of the `-oneofschema` libraries. We can possibly fix it with a small PR once we have figured out the root cause, but for now we need to work around it by using `--skip-lock` when installing flowmachine inside the docker containers.

- In `flowapi/tests/unit/test_get_query.py`, there is a parametrised test called `test_get_json_status_code` which checks that flowapi returns the expected http status code depending on the current state of the flomwachine query. I had to disable the 200 case for this because

- The `query_id <-> (query_kind, query_params)` lookup in redis is now handled by a new class called `QueryInfoLookup`. Currently this also has the responsibility of determining whether a query is "missing", but this responsibility should be transferred to `QueryStateMachine` asap.


flowmachine/flowmachine/core/server/server.py
```diff
@@ -2,198 +2,55 @@
...
-async def recv(port):
-    """
-    Listen for messages coming in via zeromq on the given port, and dispatch them.
-    """
 
-    ctx = Context.instance()
-    socket = ctx.socket(zmq.ROUTER)
-    socket.bind(f"tcp://*:{port}")
-
-    def shutdown():
```
greenape: I reckon this is why the coverage has dropped - without this, no coverage is collected for things in flowmachine which get hit in the integration tests.

flowmachine/flowmachine/core/server/action_handlers.py
```diff
@@ -0,0 +1,327 @@
...
+    """
+    spec = APISpec(
+        title="FlowAPI",
+        version="1.0.0",
+        openapi_version="3.0.2",
+        plugins=[MarshmallowPlugin()],
+    )
+    spec.components.schema("FlowmachineQuerySchema", schema=FlowmachineQuerySchema)
+    schemas_spec = spec.to_dict()["components"]["schemas"]
+    schemas_spec = sort_recursively(schemas_spec)  # needed for reproducible output
```
greenape: The reproducibility part is significant for the tests, right? Wouldn't we rather sort in the test and avoid paying for that here?

flowmachine/flowmachine/core/server/action_handlers.py
```diff
@@ -0,0 +1,327 @@
...
+    """
+    Handler for 'get_available_queries' action.
+
+    Returns a list of available flowmachine queries.
+    """
+    available_queries = list(FlowmachineQuerySchema.type_schemas.keys())
+    return ZMQReply(status="success", payload={"available_queries": available_queries})
+
+
+def action_handler__get_query_schemas():
```
greenape: Worth memoizing this? No reason obvious to me that you'd need to generate it per-request.

flowmachine/flowmachine/core/server/action_handlers.py
```diff
@@ -0,0 +1,327 @@
...
+    parameters needed to construct the query.
+    """
+    try:
+        query_obj = FlowmachineQuerySchema().load(action_params)
+    except ValidationError as exc:
+        # The dictionary of marshmallow errors can contain integers as keys,
+        # which will raise an error when converting to JSON (where the keys
+        # must be strings). Therefore we transform the keys to strings here.
+        error_msg = "Parameter validation failed."
+        validation_error_messages = convert_dict_keys_to_strings(exc.messages)
```
greenape: Boggles my mind a little that this is necessary. Silly marshmallow :P

flowmachine/flowmachine/core/server/action_handlers.py
```diff
@@ -0,0 +1,327 @@
...
+        return None
+
+
+def action_handler__poll_query(query_id):
+    """
+    Handler for the 'poll_query' action.
+
+    Returns the status of the query with the given `query_id`.
+    """
+    query_kind = _get_query_kind_for_query_id(query_id)
```
greenape: One thing to consider for the future is whether this is actually something that should go via zmq as well, particularly if we make all this a bit more more async.

flowmachine/flowmachine/core/server/action_handlers.py
```diff
@@ -0,0 +1,327 @@
...
+    # avoid this separate treatment.
+    q_info_lookup = QueryInfoLookup(Query.redis)
+    if not q_info_lookup.query_is_known(query_id):
+        msg = f"Unknown query id: '{query_id}'"
+        payload = {"query_id": query_id, "query_state": "awol"}
+        return ZMQReply(status="error", msg=msg, payload=payload)
+
+    query_state = QueryStateMachine(Query.redis, query_id).current_query_state
+
+    if query_state == QueryState.COMPLETED:
```
greenape: As noted, probably better to shove the messages into `QueryState` so this whole thing can vanish.

flowapi/tests/unit/test_geography.py
```diff
@@ -48,6 +51,7 @@
     )
 
 
+@pytest.mark.xfail(reason="This test may need to be reworked...")
```
greenape: Why?

integration_tests/tests/flowmachine_server_tests/test_action_poll_query.py
```diff
@@ -1,18 +1,63 @@
...
+        "payload": {
+            "query_id": expected_query_id,
+            "query_kind": "dummy_query",
+            "query_state": "completed",
+        },
+    }
+    assert expected_reply == reply
+
+
+@pytest.mark.xfail(
```
greenape: Why's this actually fail again?

flowmachine/flowmachine/core/server/query_schemas/base_exposed_query.py
```diff
@@ -0,0 +1,122 @@
...
+from abc import ABCMeta, abstractmethod
+from copy import deepcopy
+
+from flowmachine.core import Query
+from flowmachine.core.query_info_lookup import QueryInfoLookup
+
+__all__ = ["BaseExposedQuery"]
+
+
+class BaseExposedQuery(metaclass=ABCMeta):
```
greenape: I need to revisit why there's this distinction, I did understand it but the reason is currently lost to me.

flowmachine/flowmachine/core/server/query_schemas/base_exposed_query.py
```diff
@@ -0,0 +1,122 @@
...
+        # code block should be removed and simply `q.md5` be returned as the
+        # query_id.
+        try:
+            # In addition to the actual query, also set an aggregated version running.
+            # This is the one which we return via the API so that we don't expose any
+            # individual-level data.
+            q_agg = q.aggregate()
+            q_agg.store()
+            query_id = q_agg.md5
+            # FIXME: this is a big hack to register the query info lookup also
```
greenape: Since all `q.aggregate()` does is return `SpatialAggregate(q)`, this should be pretty easy to fix. Should do sooner rather than later, as it will change the api quite a bit.

flowmachine/flowmachine/core/server/query_schemas/meaningful_locations.py
```diff
@@ -0,0 +1,311 @@
...
+    tower_cluster_radius = fields.Float(required=False, default=1.0)
+    tower_cluster_call_threshold = fields.Integer(required=False, default=0)
+    subscriber_subset = SubscriberSubset(required=False)
+
+    @post_load
+    def make_query_object(self, params):
+        return MeaningfulLocationsAggregateExposed(**params)
+
+
+def _make_meaningful_locations_object(
```
greenape: One main reason this wasn't flattened in my original implementation is that 99% of the time, you want (like with the aggregates) to put the queries it composes into cache before starting to run, to avoid paying a massive performance penalty. Obviously the signature is nicer this way, but I do have a concern that this puts a barrier in the way of efficiently caching stuff in order to make a primarily machine facing interface more readable to humans.

flowmachine/flowmachine/core/server/query_schemas/modal_location.py
```diff
@@ -0,0 +1,57 @@
...
+from marshmallow import Schema, fields, post_load
+from marshmallow.validate import OneOf, Length
+from marshmallow_oneofschema import OneOfSchema
+
+from .base_exposed_query import BaseExposedQuery
+from .custom_fields import AggregationUnit, SubscriberSubset
+from .daily_location import DailyLocationSchema, DailyLocationExposed
+
+
+class InputToModalLocationSchema(OneOfSchema):
```
greenape: I don't think there's actually a better way to do this, but I don't like it much.. Primarily because it adds hoops to creating a new thing that you can feed into modal locations.

flowmachine/Pipfile
```diff
@@ -37,6 +40,10 @@ asynctest = "*"
 cachey = "*"
 approvaltests = "*"
 watchdog = "*"
+ipdb = "*"
 
 [requires]
 python_version = "3.7"
+
+[pipenv]
+allow_prereleases = true
```
greenape: Need to be careful here, because this is globally applicable.

flowmachine/flowmachine/core/server/server.py
```diff
@@ -2,198 +2,55 @@
...
-from .zmq_interface import ZMQMultipartMessage, ZMQInterfaceError
 
+import flowmachine
+from .action_handlers import perform_action
+from .exceptions import FlowmachineServerError
+from .zmq_helpers import ZMQReply, parse_zmq_message
 
 logger = structlog.get_logger(__name__)
-# Logger for all queries run or accessed
-query_run_log = logging.getLogger("flowmachine-server")
```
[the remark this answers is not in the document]
maxalbert: Oh, good point! I took it out initially to be able to better understand the code and forgot to put it back in.

flowmachine/flowmachine/core/server/zmq_helpers.py
```diff
@@ -0,0 +1,144 @@
...
+        self.payload = ZMQReplyPayload(payload)
+
+    def as_json(self):
+        """
+        Return a JSON object
+        """
+        return {"status": self.status.value, "msg": self.msg, "payload": self.payload}
+
+
+def parse_zmq_message(msg_str):
```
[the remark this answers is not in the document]
maxalbert: Yes, and we should. I'll open a separate follow-up issue for this.

flowmachine/flowmachine/core/server/query_schemas/daily_location.py
```diff
@@ -0,0 +1,52 @@
...
+from flowmachine.features import daily_location
+from .base_exposed_query import BaseExposedQuery
+
+__all__ = ["DailyLocationSchema", "DailyLocationExposed"]
+
+
+class DailyLocationSchema(Schema):
+    date = fields.Date(required=True)
+    method = fields.String(required=True, validate=OneOf(["last", "most-common"]))
+    aggregation_unit = fields.String(
```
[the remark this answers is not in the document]
maxalbert: Nope, good catch. 👍

flowmachine/flowmachine/core/server/server.py
maxalbert: Oh crikey, yes. Need to put that back in too. 🙈

flowmachine/flowmachine/core/server/action_handlers.py
maxalbert: Yes, good point. It's also kind of nice to return the schemas in the order they are defined internally.

flowmachine/flowmachine/core/server/action_handlers.py
maxalbert: Yeah... Maybe I missed something how to work around this, but by default marshmallow sometimes exports error messages with an integer as the dict key (presumably because it refers to the index in a list). If we can would be good to find a way to tidy this up a little - in particular the forwarding of error messages received from marshmallow. While they are accurate and reflect all the validation errors, they are a bit inconvenient to interpret from the user's point of view, so we probably want to massage them a little before sending them back.

flowmachine/flowmachine/core/server/action_handlers.py
maxalbert: Sure. Are you thinking of something like `functools.lru_cache`?

flowmachine/flowmachine/core/server/action_handlers.py
maxalbert: Not sure I see what exactly you have in mind. Can you elaborate?

flowmachine/flowmachine/core/server/action_handlers.py
maxalbert: Agreed. I'll open a follow-up issue.

flowapi/tests/unit/test_geography.py
maxalbert: Hmm, I took a look at this again because I was wondering whether I had marked it `xfail` during refactoring. But after having a second look I think it's either not needed any more, or we need to adjust what exactly it tries to test. A third option is that the logic in the `get_geography()` route needs tweaking.

integration_tests/tests/flowmachine_server_tests/test_action_poll_query.py
maxalbert: I thought we can't currently test this because the `QueryStateMachine` doesn't know how to handle missing queries, but it turns out we can test it anyway (just the internal implementation uses a different approach), so I have re-enabled this.

flowmachine/flowmachine/core/server/query_schemas/base_exposed_query.py
maxalbert: Agreed. Do we already have an issue for this?

flowmachine/flowmachine/core/server/query_schemas/meaningful_locations.py
maxalbert: That's a good point, although I feel the user should not need to worry about this. My hunch is that we can actually pretty easily achieve this by adding marshmallow schemas for these sub-queries which we simply don't expose via the API. Might need a little re-rigging of the marshmallow machinery but it should actually be fairly easy (I hope).

flowmachine/flowmachine/core/server/query_schemas/modal_location.py
maxalbert: I used to share this sentiment. I agree it feels a little clunky at first, but the more I think about it the more I actually like it because it creates a meaningful abstraction that states: "these are precisely the types of things that can be fed into `ModalLocation`".

I don't think it creates extra hoops actually, because the only difference to if this class didn't exist is that now you need to add a new class to the list`InputToModalLocationSchema.type_schemas`, and otherwise you'd need to add it to `FlowmachineQuerySchema.type_schemas`. But apart from that the process of adding something new that you can feed to modal locations is identical.

flowmachine/Pipfile
maxalbert: Yes, I was slightly concerned about this too, but I didn't think pipenv would allow us to install a pre-release version of marshmallow without specifying this flag. But actually when I tried it again with the pinned version it seems to work, so maybe this can be removed? I'll leave it in for now but may try another CI run with it removed once we have addressed everything else.

flowmachine/flowmachine/core/server/query_schemas/modal_location.py
greenape: I _think_ what I'm wondering is why one can't have all the schemas inherit from the base one, deal with the type field bit by marking it as `dump_only` or something like that, and then define the things that can be accepted in terms of inheritance and, in openapi world, `allOf`, e.g. you have a 'meta' schema/mixin scenario, so if your schema wants to be accepted as input to modal locations it includes the `ReferenceLocation` schema in the bases and modal location then `InputToModalLocationSchema(FlowmachineQuerySchema, ReferenceLocation)`, or something like that.

But I think it just bugs me because it introduces a weird bidirectional inheritance situation, where the superclass is in charge of declaring child classes :P

maxalbert: Interesting points, I'll mull over this when I have a bit more time and headspace. :)

My first approach was actually to use `dump_only` but I couldn't quite get it to work (which doesn't mean it's impossible, of course) and even the incomplete attempts didn't feel very tidy. But there may well be a better way of doing things along the lines you suggested. Happy to revisit after this is merged if we find a better approach.

maxalbert:
Thanks a lot for the comments! Fully agreed on revisiting and cleaning up immediately. I'll open separate issues for any cleanup tasks we have identified. I think I have addressed everything, let me know if there is anything else you'd like changed before merging.

PR #546, opened 2019-04-01
Marshmallow pre fix

greenape:
This is a temporary workaround for the pre-release status of marshmallow, and the version conflict that this introduces.

Rather than use pre-release mode for all the pipfiles, this pins against marshmallow using git tags, which allows us to use their release candidates without the need for pre or lock skipping.


maxalbert:
Looks good, except that file `test_daily_location_results.test_daily_location_1_sql.received.txt` should not be committed to version control. (It contains the outcome of the actual test run, rather than the reference result.) I thought we already had a rule in `.gitignore` to exclude any `*.received.txt` files but apparently not...

maxalbert:
(I just pushed changes for both of these things.)

maxalbert:
Just as a question: I saw that you explicitly added `marshmallow` and `marshmallow-oneofschema` to the integration_tests and docs Pipfiles. Shouldn't these rather be added to `flowmachine/setup.py` and then they would be installed automatically since they are flowmachine dependencies?

greenape:
Can't use vcs packages in setup.py anymore, because pip no longer processes `dependency_links`

PR #552, opened 2019-04-02
Remove default passwords in flowmachine

### LINKED ISSUE #551: Remove password defaults
Both the Dockerfile and the `connect` function have defaults set for various passwords. These should be removed, and not setting a password should be an error.

greenape:
Closes #551 

### I have:

### Description

Removes default passwords for Redis & Flowdb and makes not providing one of a secret, env var, or parameter an error.


flowmachine/tests/conftest.py
```diff
@@ -106,7 +106,7 @@ def skip_datecheck(request, monkeypatch):
 
 @pytest.fixture(autouse=True)
 def flowmachine_connect():
-    con = flowmachine.connect()
+    con = flowmachine.connect(db_pass="foo", redis_password="fm_redis")
```
maxalbert: I'm fine with explicitly providing a password in the tests. Just thinking that an alternative could be to _require_ the env vars to be set instead (and abort the tests immediately if they are not). No strong opinion really which is better, but it would help us to avoid sprinkling dummy passwords in a number of places?

PR #554, opened 2019-04-02
Claims check

### LINKED ISSUE #270: Tidy up logic in claims checks

### LINKED ISSUE #535: Refactor claims checks (retrieve knowledge about which permissions need checking from each query kind)
Until know all query kinds needed the same permission checks (whether they are allowed to perform `run`/`poll`/`get_result` on the query, and the spatial aggregation permissions).

However, not all queries will require the same kinds of permissions. For example, `available_dates` does not require any spatial aggregation so the claims checker would break in its current form because it tries to access the `aggregation_unit` parameter which doesn't exist.

To keep things more general and extendable, we should make each query aware about which permissions are required for it, so that the claims checker can simply retrieve this knowledge from the query and perform only the necessary checks.

greenape:
Closes #270 closes #535

### I have:

### Description

Not a user facing change, but this does away with the claims checking decorators all together and instead moves logic for this into a user object, which integrates better with flask-jwt-extended. Rather than do lots of fiddly checking in the decorator, this raises an exception indicating that claims verification failed. That then gets picked up the (previously unused) user claims verification failed handler, which returns a 401 to the user and logs the event.

Usage is pretty simple - call the requisite method at the start of the route on the `current_user` object. This either returns True (which can be ignored), or raises the exception and aborts the route.

For different logic needed claims, a new method would need to be added to `UserObject`.


flowapi/flowapi/user_model.py
```diff
@@ -0,0 +1,285 @@
...
+            True if the user can do 'action' with this query
+
+        Raises
+        ------
+        UserClaimsVerificationError
+            If the user cannot do action with this kind of query at this level of aggregation
+        """
+        try:
+            action_rights = self.claims[query_kind]["permissions"][action]
+            unit_rights = (
```
maxalbert: Nitpick, but how about calling this `aggregation_rights` instead of `unit_right`?

flowapi/flowapi/user_model.py
```diff
@@ -0,0 +1,285 @@
...
+        """
+        request.socket.send_json(
+            {
+                "request_id": request.request_id,
+                "action": "get_query_params",
+                "params": {"query_id": query_id},
+            }
+        )
+        reply = await request.socket.recv_json()
+        if reply["status"] == "success":
```
maxalbert: What happens if `reply["status"]` is `"error"`? Shouldn't this raise an exception as well? (And similarly below?)

Issue #467, opened 2019-03-12
Integration tests fail with default flowdb/redis port values in .env file


maxalbert:
The `integration_tests/.env` file contains the following default values:
```
FLOWDB_PORT=9997
REDIS_PORT=9996
```
These are different to the default values in the toplevel `docker-compose-dev.yml` file (where the values are `FLOWDB_PORT=9000` and `REDIS_PORT_6379`, respectively). As a consequence, the integration tests fail when run against the default local dev setup.

To reproduce:
```bash
$ make flowdb_testdata-up redis-up
$ cd integration_tests/ && pytest -x tests/flowmachine_tests/
```
Error:
```
Traceback (most recent call last):
[...]
sqlalchemy.exc.OperationalError: (psycopg2.OperationalError) could not connect to server: Connection refused
	Is the server running on host "localhost" (::1) and accepting
	TCP/IP connections on port 9997?
could not connect to server: Connection refused
	Is the server running on host "localhost" (127.0.0.1) and accepting
	TCP/IP connections on port 9997?
```

The most immediate fix I can think of is to change the values in `integration_tests/.env` to 9000 and 6379 (or to remove these vars from that file altogether and force the user to set them explicitly if flowdb/redis are not running on the default ports).

However, it is slightly annoying that the docker-compose file and the `.env` file can get out of sync with each other. Ideally these defaults should only be configured in a single place, but I can't see where this would be.

greenape:
There's a totally separate compose file for the integration tests, and a script for running them (`cd integration_tests && pipenv run run-tests`). The different values in the `.env` are actually there to avoid colliding with any other running flowdb & Redis instances, but I'm definitely up for a better solution here.

maxalbert:
Well, I've been thinking for a while that the integration tests should really use the toplevel docker-compose file anyway (and we should remove the one in that folder). This would have the added benefit that this docker-compose file works as expected.

I don't really see any potential for conflict with already running flowdb/redis instances. During the CI process you only have one instance (which would be spun up with the default port values), and locally you typically want to test against the running instances anyway (and if not you can always spin up new instances using the toplevel docker-compose file by passing different env vars).

PR #562, opened 2019-04-03
Unique subscriber counts

### LINKED ISSUE #588: Add API route for unique subscriber counts
`flowmachine.features.location.unique_subscriber_counts.UniqueSubscriberCounts` should have an exposed API route.

OwlHute:
Closes #588

### I have:

### Description

Description of what this PR _does_, and what it _changes_.


flowclient/flowclient/client.py
```diff
@@ -935,3 +935,38 @@ def flows(
...
+
+    Returns
+    -------
+    dict
+        Dict which functions as the query specification
+
+    Example
+    -------
+    >>> unique_subscriber_counts( { "start_date" : "2019-03-20", "end_date" : "2019-03-31", "aggregation_unit" : "admin3" } )
+    [ ]  # TODO - Fill this in
```
maxalbert: Accidentally forgot to delete this TODO item?

flowclient/flowclient/client.py
```diff
@@ -935,3 +935,38 @@ def flows(
...
+) -> dict:
+    """
+    Return query spec for unique subscriber counts
+
+    Parameters
+    ----------
+    start_date : str
+        ISO format date of the first day of the count, e.g. "2016-01-01"
+    end_date : str
+        ISO format date of the day _after_ the final date of the count, e.g. "2016-01-08"
```
maxalbert: I think the `end_date` is actually inclusive:
```suggestion
        ISO format date of the final date of the count, e.g. "2016-01-08"
```

maxalbert:
The additions to the API spec (in the file `tests/flowmachine_server_tests/test_server.test_api_spec_of_flowmachine_query_schemas.approved.txt`) look correct, but they are in the "wrong" order (the keys in the file are sorted alphabetically so that the result is reproducible for the tests).

I assume you edited this file manually? While this is possible, it's easier to let ApprovalTests auto-generate it for you. When you run the relevant integration test (e.g. via `cd integration_tests && pipenv run -svx tests/flowmachine_server_tests/test_server`) it will open a diff tool and allow you to automatically accept the changes.

However, currently it is configured to use `opendiff` which I'm not sure is available on Windows. You will want to install some diff tool (see [here](https://www.git-tower.com/blog/diff-tools-windows/) for some choices) and temporarily change the setting in`integration_tests/tests/conftest.py`(near the bottom) from `opendiff` to whatever tool you installed. Apologies that this is a bit clunky at the moment - we should look into supporting a variety of tools out of the box (which we can do in a separate PR once we have got it working for you on Windows).

flowclient/flowclient/client.py
OwlHute: I entirely agree that it seems more intuitive for a date range to be inclusive. But I copied that comment from another entry point, location_event_counts(). Maybe the idea is that date ranges should be like python substrings, where the "to index" points to the byte after the final one desired?! Can I leave this one for a definitive judgement from someone (Jono?) familiar with the code?

greenape: It is a slightly whacky legacy thing, because otherwise for a single day you need to write the date twice, or specify the second date in date time form as just before midnight. So yes, end_date is exclusive.

maxalbert: Ah thanks, I wasn't aware that that's how things are supposed to work. However, I'm pretty sure this is not the current behaviour because the `end_date` parameters is passed all the way down to the flowmachine library query, which passes it on to `EventTableSubset` where the comparison uses `<=` (see [here](https://github.com/Flowminder/FlowKit/blob/master/flowmachine/flowmachine/features/utilities/event_table_subset.py#L209); or [here](https://github.com/Flowminder/FlowKit/blob/master/flowmachine/flowmachine/features/utilities/event_table_subset.py#L246) for the old version).

In any case, I'm happy to keep the docstring unchanged for this PR and if it needs tweaking we can do this in a separate PR.

flowclient/flowclient/client.py
OwlHute: I wasn't sure what goes there. Is there a systematic way to find out? I assume I can try actually running the code and extracting the relavent string from the output.

maxalbert: Yes, you'd need to actually run it and copy & paste the output.

Note that in order to be able to successfully run it, you will need to change the dates to something between `2016-01-01` and `2016-01-07` because the data in the test database is between those dates.

flowclient/flowclient/client.py
greenape: Yeah, but see https://github.com/Flowminder/FlowKit/blob/master/flowmachine/flowmachine/features/utilities/event_table_subset.py#L157

maxalbert:
@OwlHute FYI, in PR #566 I added a bit of documentation about the approvaltests-based tests and added a toplevel `approvaltests_diff_reporters.json` file which allows to configure the diff tool more easily. See [here](https://github.com/Flowminder/FlowKit/pull/566/files#diff-42dfce04a254cd3fffef010621e82be6) for the addition to the docs about this.

OwlHute:
Re Max's flowmachine_server_tests comment, for now I have manually moved the unique_subscriber_counts entries into alphabetical order, as I have had trouble running the integration_tests:

Firstly, I have established that these tests definietly won't run in Windows, with or without the FlowKit integration test  pipenv environment, because two 3rd party modules they use, "pglast" and "ujson" are Unix-only. (The only other Unix dependency I've come across in FlowKit is the use of "pwd", for dealing with the Unix password database.)

I am able to run them (without missing module failures) in "Ubuntu for Windows" in their FlowKit development pipenv environment. However, this fails with hordes of errors, all I think due to the same thing: a database port not being configured, or its config not being picked up by the test scripts. Here is a sample error message:

>       conn = _connect(dsn, connection_factory=connection_factory, **kwasync)
E       sqlalchemy.exc.OperationalError: (psycopg2.OperationalError) could not connect to server: Connection refused
E               Is the server running on host "localhost" (127.0.0.1) and accepting
E               TCP/IP connections on port 9997?

greenape:
@OwlHute There's a readme file that explains how to run the integration tests, located in the integration tests folder. Note that you'll need/want to use Pipenv, because various settings are in `.env` files that pipenv and docker both pick up automatically. You should be able to run the tests using `pipenv install && pipenv run run-tests` from the integration tests folder. 

The test that is currently failing (which I believe @maxalbert has just fixed for you), is the autogenerated api schema.

Once the tests are all passing, we'll also want to look at the test coverage to assess whether you need to add some more tests to cover the new code.

maxalbert:
Thanks for the changes @OwlHute! The toplevel keys are now in the correct order. The nested keys were still in non-alphabetical order; for the sake of expediency I just pushed a change to fix this (you will want to run `git pull` to bring your local branch up to date).

Apologies for the test run failures. This is due to #467, I'll submit a PR to fix this in a minute. In the meantime, you should be able to run the tests by setting the following environment variables explicitly in your shell (you will need to run `pipenv shell` first to activate the pipenv environment, otherwise the `integration_tests/.env` file overwrites them again).
```
cd integration_tests
pipenv shell
export FLOWDB_PORT=9000
export REDIS_PORT=6379
export FLOWMACHINE_PORT=5555
```
Are you able to run the tests with these settings?

If you want to run just the API spec test (either for speed, or to avoid failures with Unix-only modules) you can do it like this (from within the integration tests pipenv environment as above):
```
pytest -svx tests/flowmachine_server_tests/test_server.py 
```

maxalbert:
@greenape Is there a need for any more test coverage or are all the changes here covered by existing tests?

flowmachine/flowmachine/core/server/query_schemas/unique_subscriber_counts.py
```diff
@@ -0,0 +1,45 @@
...
+        # Note: all input parameters need to be defined as attributes on `self`
+        # so that marshmallow can serialise the object correctly.
+        self.start_date = start_date
+        self.end_date = end_date
+        self.aggregation_unit = aggregation_unit
+
+    @property
+    def _flowmachine_query_obj(self):
+        """
+        Return the underlying flowmachine daily_location object.
```
greenape:
```suggestion
        Return the underlying flowmachine UniqueSubscriberCounts object.
```

flowclient/flowclient/client.py
```diff
@@ -935,3 +935,36 @@ def flows(
...
+
+    Returns
+    -------
+    dict
+        Dict which functions as the query specification
+
+    Example
+    -------
+    >>> unique_subscriber_counts( { "start_date" : "2019-03-20", "end_date" : "2019-03-31", "aggregation_unit" : "admin3" } )
+    [ ]  # TODO - Fill this in
```
greenape: Example needs to be filled in as mentioned, and should actually work when run (args are wrapped in a dict here, so this would error when run?)

OwlHute:
@Max, many thanks for these port settings. I ran the test in the pipenv environment in Ubuntu, and the output looks a lot better, in that it is much shorter! But there are still a few errors around connections, e.g.:

>               sock.connect(socket_address)
>  E               ConnectionRefusedError: [Errno 111] Connection refused

E               ConnectionRefusedError: [Errno 111] Connection refused

> /home/jr/.local/share/virtualenvs/integration_tests-sEjdA8oo/lib/python3.7/site-packages/redis/connection.py:538: ConnectionRefusedError

greenape:
For reference, I'm looking at the coverage report for the added schema here: https://codecov.io/gh/Flowminder/FlowKit/src/73abcbf0213132937f6cb1d3b11640bf4e0a5b6e/flowmachine/flowmachine/core/server/query_schemas/unique_subscriber_counts.py

and the client, here: https://codecov.io/gh/Flowminder/FlowKit/src/unique_subscriber_counts/flowclient/flowclient/client.py

flowmachine/flowmachine/core/server/query_schemas/unique_subscriber_counts.py
OwlHute: I've implemented the suggested alternative, but that _flowmachine_query_ob j() method was already present and the only change I made was to fix the object name in theader. So that won't change the test coverage. Presumably something must be added to actually call _flowmachine_query_obj() ?! (I assumed some test would run it automatically, as its name looks pretty generic.)

maxalbert: @OwhHute This particular suggestion about changing `daily_location` to `UniqueSubscriberCounts` in the docstring is independent of @greenape's comment about code coverage and the need to add an additional test. (It may look related because Github displays it right underneath, but it has nothing to do with it).

integration_tests/tests/test_queries.py
```diff
@@ -26,6 +26,22 @@
...
+        (
+            "unique_subscribers_count",
+            {
+                "start_date": "2016-01-01",
+                "end_date": "2016-01-02",
+                "aggregation_unit": "admin3",
+            },
+        ),
+        (
+            "unique_subscribers_count",
```
maxalbert: Why was this added twice? It seems to be an exact copy of the previous one?

maxalbert:
Thanks for the changes @OwlHute. This looks (almost) good, but we need a changelog entry please - see the checkbox in the PR template at the top.

(Ideally we'd also want a Github issue that this PR closes, but we can add that separately.)

OwlHute:
I have a horrible feeling I may have morphed the name "unique_subscriber_counts" to "unique_subscribers_count" in the added test - Checking now ..

OwlHute:
Fixed!

integration_tests/tests/test_queries.py
OwlHute: DOH! Looking at what I thought was one previous test case, I assumed it was the input arguments and the expected output arguments. But on closer inspection I see it is two previous test cases and only the input arguments are required. I've removed the duplicate entry now and re-committed & pushed to github

flowclient/flowclient/client.py
```diff
@@ -935,3 +935,31 @@ def flows(
         "to_location": to_location,
         "aggregation_unit": aggregation_unit,
     }
+
+
+def unique_subscriber_counts(
+    start_date: str, stop_date: str, aggregation_unit: str
```
maxalbert: Sorry, I missed this earlier: this parameter should actually be named `end_date` rather than `stop_date`. (This was caught by the most recent integration test you added, which is why they are currently failing.)

maxalbert:
This looks good now, as far as I can tell. Happy with the test coverage @greenape?

greenape:
Hm. Looks like needs a case added to one or both of `test_get_query_params` or `test_get_query_kind` in integration tests, in order to hit the `get_obj_type` method.

OwlHute:
> 
> 
> Hm. Looks like needs a case added to one or both of `test_get_query_params` or `test_get_query_kind` in integration tests, in order to hit the `get_obj_type` method.

In test_get_query_params.py I added a block within @pytest.mark.parametrize but the automated test objected to this. So I presume the new block needs to be within a second @pytest.mark.parametrize block of its own.

maxalbert:
@OwlHute Thanks for adding the extra test. It failed because the assert statement contained a hard-coded `"daily_location"` but your new test was against `unique_subscriber_counts`. I just pushed a change to fix this, so hopefully it will all pass now.

maxalbert:
I'm happy to merge this now. The code coverage is still every so slightly reduced, but if I'm interpreting the codecov reports correctly then this seems to be a code path that's also not hit by other queries, so I'm happy to deal with this separately.

maxalbert:
FYI, I have removed the tests in `test_action_get_params.py` and `test_action_get_query_kind.py` again because they are not meant to test specific queries (but rather the action handler functions), see #584.

PR #574, opened 2019-04-04
Fix hang with error

### LINKED ISSUE #514: Server hangs if creating a query raised an error
When running a query via the api, a query which raises an error on creating the object (e.g. daily locations for a date with no data) will just hang. Server continues fine for subsequent responses, but you'd expect 400 status back, perhaps with the error message.

greenape:
Closes #514

### I have:

### Description

Fixes hanging the client when a query object couldn't be created. This arose because any exception raised in the run_query action handler got swallowed by AsyncIO. It'll now respond with an error message instead of trying to propagate the exception.


flowmachine/tests/server/test_action_handlers.py
```diff
@@ -9,10 +9,26 @@
...
+    action_handler__run_query,
 )
 from flowmachine.core.server.zmq_helpers import ZMQReplyStatus
 
 
+def test_run_query_error_handled(monkeypatch, dummy_redis):
+    """
+    Run query handler should return an error status if query construction failed.
+    """
+    monkeypatch.setattr(Query, "redis", dummy_redis)
```
maxalbert: It just occured to me: wouldn't it be simpler if we moved this monkeypatch into the fixture itself?

maxalbert:
Looks like code coverage hasn't changed, so the error path on the flowapi side seems to be tested through other means already. Happy to merge this.

PR #580, opened 2019-04-05
Location introversion

### LINKED ISSUE #577: Add API route for location introversion
`flowmachine.features.location.location_introversion.LocationIntroversion` should have an exposed API route.

I'm thinking same parameterisation as the other exposed routes - start, stop, aggregation unit + direction, but I'd like direction to be optional with default `both`.

- [ ] Add `LocationIntroversionExposed` class
- [ ] Add marshmallow schema
- [ ] Add schema to `FlowmachineQuerySchema`
- [ ] Add flowclient function
- [ ] Add tests to integration tests
    - [ ] Add test case to `test_get_query_params`
    - [ ] Add test case to `test_run_query`
- [ ] Add query endpoint to Flowauth demo data
- [ ] Update approval test data for integration tests

OwlHute:
Closes #577 

### I have:

### Description

Description of what this PR _does_, and what it _changes_.


flowclient/flowclient/client.py
```diff
@@ -935,3 +935,34 @@ def flows(
         "to_location": to_location,
         "aggregation_unit": aggregation_unit,
     }
+
+
+def location_introversion(
+    start_date: str, stop_date: str, aggregation_unit: str, direction: str = "both"
```
maxalbert: In flowclient, the exposed parameter should be called `end_date`:
```suggestion
    start_date: str, end_date: str, aggregation_unit: str, direction: str = "both"
```

flowmachine/flowmachine/core/server/query_schemas/location_introversion.py
```diff
@@ -0,0 +1,52 @@
...
+__all__ = ["LocationIntroversionSchema", "LocationIntroversionExposed"]
+
+
+class LocationIntroversionSchema(Schema):
+
+    start_date = fields.Date(required=True)
+    end_date = fields.Date(required=True)
+    aggregation_unit = AggregationUnit()
+    direction = fields.String(
+        required=True, validate=OneOf(["in", "out", "both"])
```
maxalbert: We probably want `default="both"` here, too.

integration_tests/tests/flowmachine_server_tests/test_server.test_api_spec_of_flowmachine_query_schemas.approved.txt
```diff
@@ -271,6 +275,41 @@
...
+        "type": "string"
+      }
+    },
+    "required": [
+      "aggregation_unit",
+      "end_date",
+      "start_date"
+    ],
+    "type": "object"
+  }
```
maxalbert: Missing comma after this closing `}`.

integration_tests/tests/flowmachine_server_tests/test_server.test_api_spec_of_flowmachine_query_schemas.approved.txt
```diff
@@ -271,6 +275,41 @@
...
+        "format": "date",
+        "type": "string"
+      },
+      "start_date": {
+        "format": "date",
+        "type": "string"
+      }
+    },
+    "required": [
+      "aggregation_unit",
```
maxalbert: The `"direction"` parameters is missing after `"aggregation_unit"`.

flowclient/flowclient/client.py
OwlHute: Fixed (and I'll fix that in TotalNetworkObjects too)

flowmachine/flowmachine/core/server/query_schemas/location_introversion.py
OwlHute: Added

integration_tests/tests/flowmachine_server_tests/test_server.test_api_spec_of_flowmachine_query_schemas.approved.txt
OwlHute: Fixed

integration_tests/tests/flowmachine_server_tests/test_server.test_api_spec_of_flowmachine_query_schemas.approved.txt
OwlHute: Ah, that's because "direction" is an optional parameter so I thought it would not belong in a group named "required". But I can add it if "required" means something slightly different in this context.

flowmachine/flowmachine/core/server/query_schemas/location_introversion.py
OwlHute: Hmm, the automated test doesn't like the "default" keyword:
> flowmachine/core/server/query_schemas/location_introversion.py:21: in
> LocationIntroversionSchema
>   required=True, validate=OneOf(["in", "out", "both"], default="both")
> E   TypeError: __init__() got an unexpected keyword argument 'default'

flowmachine/flowmachine/core/server/query_schemas/location_introversion.py
```diff
@@ -0,0 +1,52 @@
...
+__all__ = ["LocationIntroversionSchema", "LocationIntroversionExposed"]
+
+
+class LocationIntroversionSchema(Schema):
+
+    start_date = fields.Date(required=True)
+    end_date = fields.Date(required=True)
+    aggregation_unit = AggregationUnit()
+    direction = fields.String(
+        required=True, validate=OneOf(["in", "out", "both"], default="both")
```
maxalbert: This should be `required=False`, since it's an optional argument:
```suggestion
        required=False, validate=OneOf(["in", "out", "both"], default="both")
```

integration_tests/tests/flowmachine_server_tests/test_server.test_api_spec_of_flowmachine_query_schemas.approved.txt
maxalbert: You are right, good catch! I ran approvaltests to auto-generate the correct spec, and it included `"direction"` here because you had set `required=True` in the definition of `LocationIntroversionSchema`. So the above needs changing, and then the spec needs to be re-generated. It would be helpful if you could set up your system so that you can do this rather than manually editing the file and us having to re-check it every time the tests fail.

flowmachine/flowmachine/core/server/query_schemas/location_introversion.py
maxalbert: Also, it looks like you added `default` inside the brackets of the call to `OneOf` rather than outside, which will throw an error because `default` is an argument to `fields.String`.

integration_tests/tests/flowmachine_server_tests/test_server.test_api_spec_of_flowmachine_query_schemas.approved.txt
maxalbert: And yes, just to confirm "direction" should not be included in this list of required parameters.

Issue #599, opened 2019-04-08
Add SpatialAggregate to api


greenape:
Modal locations and daily locations are presently dealt with by calling `aggregate()` on any query created by the api. This _implicitly_ returns them wrapped in `SpatialAggregate`. This should be made explicit, and the hidden call to `aggregate()` removed. This also paves the way for a broader classification of things-that-are-aggregated, which I think might warrant being a marshmallow spec in itself (probably allOf).

- [x] Add marshmallow spec for SpatialAggregate
- [ ] Remove call to `aggregate()`
- [ ] Update client
- [ ] Update claims checking in FlowAPI

greenape:
One thing to consider here is that `modal_location` operates on `daily_location` specs, so this presumably needs to be a separate function in FlowClient. Also means a bare `daily_location` query spec should no longer be allowable, and means any examples using `modal_location` or `daily_location` will need to be updated.

greenape:
#524

OwlHute:
Am I right in thinking the idea is to make this aggregation process user-callable, so an API user can use it more generally on any pair of applicable data sets one of which is a subscriber list, and produce a sort of aggregated "product" of the sets?

greenape:
Think that's in reference to #600 @OwlHute?

greenape:
Yes. The API requests that previously went:

```json
{"query_kind": "daily_location" ...}
```

Should now go:

```json
{"query_kind":"spatial_aggregate", "locations":{"query_kind": "daily_location" ...}}
```

OwlHute:
If that is the only change, it sounds pretty simple. So I'm guessing other changes are also required (including obviously changes to the relevant test cases)

OwlHute:
Following class added in flowmachine/flowmachine/core/server/query_schemas/custom_fields.py, preserving alphabetic order of existing classes:

class SpatialAggregate(fields.List):
    """
    List representing a spatial aggregate subset.
    """

    def __init__(self, required=False, allow_none=True, validate=None, **kwargs):
        if validate is not None:
            raise ValueError(
                "The SpatialAggregate field provides its own validation "
                "and thus does not accept a the 'validate' argument."
            )

        super().__init__(
            required=required, allow_none=allow_none, validate=OneOf([None]), **kwargs
        )

greenape:
Not sure I follow?

PR #607, opened 2019-04-10
Aggregate network objects

### LINKED ISSUE #601: Expose AggregateNetworkObjects via api
- [x] Refactor `AggregateNetworkObjects` to take `TotalNetworkObjects` as main argument
- [x] Create new marshmallow spec file under `server/query_schemas`
- [x] Add newly created spec to `FlowMachineQuerySchema`
- [x] Add function to `flowclient.client`
- [x] Add function to `flowclient.__init__`
- [x] Add test to `integration_tests/tests/test_queries.py`
- [x] Add `docs/source/4-developer.md` with newly added query

OwlHute:
Closes #601

### I have:

### Description

Add API route aggregate_network_objects()


docs/source/4-developer.md
```diff
@@ -90,6 +90,10 @@ The API exposes four routes:
 
 At present, the following query types are accessible through FlowAPI:
 
+- `aggregate_network_objects`
```
greenape: :+1:

flowclient/flowclient/client.py
```diff
@@ -1022,3 +1022,31 @@ def total_network_objects(
         "end_date": end_date,
         "aggregation_unit": aggregation_unit,
     }
+
+
+def aggregate_network_objects(
```
greenape: I've been giving the params for this some thought - I actually think what would be most useful is to have it take: network_objects,  statistic, by. That can be achieved without touching the underlying Flowmachine class, by constructing a tno object, and then calling aggregate on it with the by and statistic arguments.

flowmachine/flowmachine/features/network/total_network_objects.py
```diff
@@ -203,7 +203,7 @@ def aggregate(self, by=None, statistic="avg"):
         )
 
 
-class AggregateNetworkObjects(GeoDataMixin, Query):
+class AggregateNetworkObjects(TotalNetworkObjects, Query):
```
greenape: What's going on here?

OwlHute: This relates to your comment in #601 "Refactor AggregateNetworkObjects to take TotalNetworkObjects as main argument"

flowclient/flowclient/client.py
OwlHute: You'll have to expand on that a bit, in terms of roughly how the call would look. I gather tno is TotalNetworkObject, and "by" is possibly the date range?! But also, I thought you were trying to do away with calls to aggregate() (#599)

flowmachine/flowmachine/features/network/total_network_objects.py
greenape: Ah. I thing I've failed to communicate what I meant there - the suggestion was that rather than taking all the arguments necessary to create a `TotalNetworkObjects` object in the `__init__` method, that this class should instead just take a `TotalNetworkObjects` object instead of those arguments.

This change actually alters the class hierarchy of the class (https://www.digitalocean.com/community/tutorials/understanding-class-inheritance-in-python-3 is a decent overview of class inheritance in python), and I imagine will cause a few issues because `TotalNetworkObjects` also inherits from `Query`, which would give you a somewhat odd inheritance hierarchy.

OwlHute: OK I'll amend the code accordingly. (I do understand python class hierarchies, but I'll have a quick skim of that page)

flowclient/flowclient/client.py
OwlHute: As you were, I understand now what you are driving at! (in view of the review comment further down)

flowmachine/flowmachine/features/network/total_network_objects.py
```diff
@@ -249,39 +238,9 @@ class AggregateNetworkObjects(GeoDataMixin, Query):
...
-            column_name=column_name,
-            size=size,
-            polygon_table=polygon_table,
-            geom_col=geom_col,
-            hours=hours,
-            subscriber_subset=subscriber_subset,
-            subscriber_identifier=subscriber_identifier,
-        )
+    def __init__(self, total_objs, statistic="avg", by=None):
+        self.total_objs = copy.deepcopy(total_objs)
```
greenape: Don't need a deepcopy here, just an assign is fine.

greenape: Below this, period will now need to refer to the period attribute of the passed in object.

flowmachine/flowmachine/core/server/query_schemas/aggregate_network_objects.py
```diff
@@ -0,0 +1,47 @@
...
+    end_date = fields.Date(required=True)
+    aggregation_unit = AggregationUnit()
+
+    @post_load
+    def make_query_object(self, params):
+        return AggregateNetworkObjectsExposed(**params)
+
+
+class AggregateNetworkObjectsExposed(BaseExposedQuery):
+    def __init__(self, *, start_date, end_date, aggregation_unit):
```
greenape: This needs to match the `AggregateNetworkObjects` form now

flowmachine/flowmachine/core/server/query_schemas/aggregate_network_objects.py
```diff
@@ -0,0 +1,47 @@
...
+from flowmachine.features import AggregateNetworkObjects
+from .base_exposed_query import BaseExposedQuery
+from .custom_fields import AggregationUnit
+
+__all__ = ["AggregateNetworkObjectsSchema", "AggregateNetworkObjectsExposed"]
+
+
+class AggregateNetworkObjectsSchema(Schema):
+
+    start_date = fields.Date(required=True)
```
greenape: Needs to have the right fields, one of which is a bit tricky because it should refer to another schema. See FlowsSchema, or ModalLocationSchema for an example of how to do that.

flowclient/flowclient/client.py
greenape: Arguments need to be `total_network_objects_query: dict`, and `by: str` (or a more explanatory name than `by` would also be fine!)

flowmachine/flowmachine/core/server/query_schemas/aggregate_network_objects.py
```diff
@@ -0,0 +1,47 @@
...
+    @property
+    def _flowmachine_query_obj(self):
+        """
+        Return the underlying flowmachine aggregate_network_objects object.
+
+        Returns
+        -------
+        Query
+        """
+        return AggregateNetworkObjects(
```
greenape: Basic idea here is fine, but need to create the `TotalNetworkObjects` object and return the result of calling `aggregate` on it. Or, wait for a merge on #609 which includes a refactor of `TotalNetworkObjects`, because it was necessary to making `spatial_aggregate` work!

maxalbert:
@OwlHute FYI, PRs #609 (spatial aggregate) and #611 (joined spatial aggregate) have now been merged, so hopefully that helps with this PR as well (see Jono's inline comment).

Github is indicating a couple of merge conflicts with this branch due to the changes in #609 and #611 so these will need to be resolved (but should be very simple).

OwlHute:
Hi Max. I've merged the lastest "master" branch with my aggregate_network_objects private branch & pushed the latter to github for #607

Running the tests on my workspace, there is one failure. But I am not clear on the reason for this & could use a bit of help to diagnose & fix it. (I think the code is 99% OK and expect only a very small fix is needed, but the trick is knowing where and what! )

maxalbert:
Thanks @OwlHute. Which tests are failing for you locally? When I click on the "Details" link next to the integration checks failure here on Github (see [here](https://circleci.com/gh/Flowminder/FlowKit/39881?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)) it shows two failing tests:
- `tests/flowmachine_server_tests/test_server.py`
- `tests/query_tests/test_queries.py`

The first one seems to be from a missing update to the API spec. Have you been able to get the approvaltests running locally? If so, it should be just a matter of running this particular test (e.g. via `pytest -svx tests/flowmachine_server_tests/test_server.py` from within the `integration_tests` pipenv environment) and it will update the spec for you (via a GUI popup in whichever diff viewer you have configured as explained in the [developer docs](https://flowminder.github.io/FlowKit/dev_environment_setup/#diff-tool-for-verifying-changes-in-approvaltests-based-tests)), so you just need to apply the changes in the diff viewer and commit the result.

For the second failing tests, if you run it via `pytest -svx tests/query_tests/test_queries.py` (again from within the `integration_tests` pipenv environment, it will succeed for the first few parameters and then fail on the `aggregate_network_objects` parameter (also see the CircleCI failure linked above). This is due to some of Jono's recent changes in #609 or #611. If you take a look at how the other JSON parameters have changed now (with a toplevel `"spatial_aggregate"` key rather than having the `query_kind` directly as the toplevel key) you should be able to deduce how your addition must be tweaked to make the tests pass.

OwlHute:
Thanks for the review comment Max. I have fixed the first error. However, I am still slightly puzzled by the second issue.

From the test output, the following is the query being sent to flowapi :

> query_kind = 'aggregate_network_objects'
> params = {'by': 'day', 'statistic': 'median', 'total_network_objects': {'aggregation_unit': 'admin3', 'end_date': '2016-01-02', 'query_kind': 'total_network_objects', 'start_date': '2016-01-01'}}

and referring to total_network_objects() in flowclient/flowclient/client.py that 'total_network_objects' dictionary is correct:

    return {
        "query_kind": "total_network_objects",
        "start_date": start_date,
        "end_date": end_date,
        "period": period,
    }

It also matches the parameters in the AggregateNetworkObjects class in flowmachine/flowmachine/core/server/query_schemas/total_network_objects.py :

> def __init__(self, *, total_network_objects, statistic="avg", by=None):
>   :::

Finally, in flowmachine/flowmachine/core/server/query_schemas/aggregate_network_objects.py the actual query is run as:

>    def _flowmachine_query_obj(self):
>       """
>       Return the underlying flowmachine aggregate_network_objects object.
>
>        Returns
>        -------
>        Query
>        """
>        tot_network_objs = self.total_network_objects._flowmachine_query_obj
>
>       return AggregateNetworkObjects(
>            tot_network_objs, statistic=self.statistic, by=self.by
>        )

I assume the latter call must introduce SpatialAggregate into the mix, but I'm not clear where and what need amending for this. (Once I've fixed this issue, I should be able to handle any similar changes in future.)

It's easy to see what the code is trying to achieve in principle, but all the same the details are damned intricate!

OwlHute:
I've got a bit further, and identified the reason for the test failure. (A bug in the test code prevents this being displayed properly!):

>            try:
>                error = response.json()["msg"]
>                # payload_info = f" Payload: {response.json()['payload']}"
>                payload_info = str(response.json()) + ' (response.status_code = ' + str(response.status_code) + ')'
>            except ValueError:
>                # Happens if the response body does not contain valid JSON
>                # (see http://docs.python-requests.org/en/master/api/#requests.Response.json)
>                error = f"the response did not contain valid JSON"
>                payload_info = ""
>            raise FlowclientConnectionError(
>               f"Something went wrong: {error}. API returned with status code: {response.status_code}.{payload_info}"
            )
E           flowclient.client.FlowclientConnectionError: Something went wrong: Aggregation unit must be specified when running a query.. API returned with status code: 400.{'msg': 'Aggregation unit must be specified when running a query.'} (response.status_code = 400)

The bug is that the dict returned by response.json() contains no 'payload' key, so the error was being displayed as just "KeyError: 'payload'".

So the next step is to find out where that message "Aggregation unit must be specified when running a query" is generated. (For the new aggregate_network_objects() API route, it is no longer appropriate!)

maxalbert:
Thanks @OwlHute! Yes, you are right that the missing `payload` key was shadowing the actual error message.

I had a look and the error happens in `flowapi/flowapi/user_model.py`. The issue is that the code in `UserObject._get_query_kinds_and_aggregation_units()` assumes that unless the `query_kind` is `spatial_aggregate` or `joined_spatial_aggregate`, it must have an aggregation unit itself.

However, `aggregate_network_objects` isn’t a _spatial_ aggregate (but rather a statistical aggregate) so it doesn't contain an `aggregation_unit` itself. The easiest fix that I can think of for the time being is to add another else branch which checks for `aggregate_network_objects` and extracts the `aggregation_unit` from the underlying `total_network_objects` (@greenape: any better ideas?).

This is a bit of a hack for the time being, but unless I'm missing something it would require larger refactorings to have a more elegant solution and I guess this is outside the scope of this particular PR. Let me know if this explanation makes sense and is sufficient to make the required tweaks.

flowmachine/flowmachine/core/server/query_schemas/aggregate_network_objects.py
```diff
@@ -0,0 +1,56 @@
...
+        Return the underlying flowmachine aggregate_network_objects object.
+
+        Returns
+        -------
+        Query
+        """
+        tot_network_objs = self.total_network_objects._flowmachine_query_obj
+
+        return AggregateNetworkObjects(
+            tot_network_objs, statistic=self.statistic, by=self.by
```
maxalbert: Jono's latest changes require this to be a keyword argument now:
```suggestion
            total_network_objects=tot_network_objs, statistic=self.statistic, by=self.by
```

maxalbert:
@OwlHute FYI, I just pushed a couple of commits to fix the bug you mentioned regarding the missing `payload` argument and to make the error message a bit more readable, so you will want to pull the latest changes.

OwlHute:
Ah many thanks Max. I was hot on the trail, as you'll have seen from an edit I made to my last message a couple of minutes ago! Yes, I was about to add a "special case" hack to check for an aggregation unit within "local_network_objects" if that was present.

OwlHute:
I fixed the "missing aggregation unit" error, but now the same query is failing with "Error message: 'Internal flowmachine server error: wrong arguments passed to handler for action 'run_query'". So I'll have to chase that down now. I hope this process of tracking down and fixing each error uncovered in turn is finite!

maxalbert:
@OwlHute Ah, that's actually a tricky one to debug, and sadly a somewhat misleading error message.

The error is thrown [here](https://github.com/Flowminder/FlowKit/blob/master/flowmachine/flowmachine/core/server/action_handlers.py#L353). In practice (on the stable master branch) this will only occur if the arguments to the action handler are wrong.

However, sadly during development it _can_ occur that a mistake is introduced during refactoring and the `TypeError` is actually triggered by some code executed _inside_ the function `action_handler__run_query()` (rather than at the point in time when it is being called).

My guess is that this happens in [L86](https://github.com/Flowminder/FlowKit/blob/master/flowmachine/flowmachine/core/server/action_handlers.py#L86) and is likely due to a missing parameter in the query spec that you added for `aggregate_network_objects`. When I debug this locally (by adding a debugger breakpoint inside the run_query action handler), the error message says: `TypeError: __init__() missing 1 required keyword-only argument: 'period'`. The reason for this is that the `period` argument in the class `TotalNetworkObjectsExposed` does not provide a default argument, so this needs to be added.

maxalbert:
@OwlHute FYI, I just pushed a change which catches the `TypeError` earlier so that it can return a meaningful error message. Apologies for the misleading one earlier!

OwlHute:
Ah, many thanks again Max. That sounds plausible! I was slowly converging on the reason, but would have taken a while longer to crack it!

OwlHute:
There is still one test error, and I have established this occurs in flowmachine/flowmachine/core/server/query_schemas/aggregate_network_objects.py, in make_query_object() whose input parameters are as follows:

> {'statistic': 'median', 'total_network_objects': <flowmachine.core.server.query_schemas.total_network_objects.TotalNetworkObjectsExposed object>, 'period': 'day'}

When the method feeds this to AggregateNetworkObjectsExposed, the result is an error as follows:

> argument of type 'TotalNetworkObjectsExposed' is not iterable''.

This "object as a dictionary value" in the middle of the JSON stems from the use of InputToAggregateNetworkObjectsSchema in the above source file, to create a schema within the schema as suggested by Jono in another review comment above.  I probably could jump up and down on the code, hacking it until the test passes. But as my "fix" might not be optimal or suitable, I think it is better to leave it until Monday am for Jono & Max to ponder.

flowmachine/flowmachine/features/network/total_network_objects.py
```diff
@@ -221,7 +210,7 @@ class AggregateNetworkObjects(GeoDataMixin, Query):
     """
 
     def __init__(self, *, total_network_objects, statistic="avg", by=None):
-        self.total_objs = total_network_objects
+        self.total_objs = copy.deepcopy(total_network_objects)
```
greenape: Don't need a copy here.

flowclient/flowclient/client.py
```diff
@@ -1099,3 +1103,36 @@ def joined_spatial_aggregate(
...
+
+    Returns
+    -------
+    dict
+        Query specification for an aggregated network objects query
+    """
+    total_network_objs = total_network_objects
+
+    if "period" not in total_network_objs:
+        total_network_objs["period"] = period
```
greenape: Why is this here?

flowmachine/flowmachine/core/server/query_schemas/aggregate_network_objects.py
```diff
@@ -0,0 +1,61 @@
...
+        Returns
+        -------
+        Query
+        """
+        tot_network_objs = self.total_network_objects._flowmachine_query_obj
+
+        return AggregateNetworkObjects(
+            total_network_objects=tot_network_objs,
+            statistic=self.statistic,
+            period=self.period,
```
greenape: The name of this argument doesn't match the kwarg that `AggregateNetworkObjects` is expecting.

flowmachine/flowmachine/core/server/query_schemas/aggregate_network_objects.py
```diff
@@ -0,0 +1,61 @@
...
+    type_schemas = {"total_network_objects": TotalNetworkObjectsSchema}
+
+
+class AggregateNetworkObjectsSchema(Schema):
+
+    total_network_objects = fields.Nested(
+        InputToAggregateNetworkObjectsSchema, required=True
+    )
+    statistic = Statistic()
+    period = Period()
```
greenape: Ah no, right. I see, they're used here but not in TNO. I don't like the inconsistency, also the two arguments (`by` and `period`) have different allowable values. The `period` for TNO goes up to year, `by` for ANO goes to century. Looks like that isn't apparent from the docstring of the class, but is in fact the behaviour one wants.

flowmachine/flowmachine/core/server/query_schemas/aggregate_network_objects.py
```diff
@@ -0,0 +1,61 @@
...
+class AggregateNetworkObjectsExposed(BaseExposedQuery):
+    def __init__(self, *, total_network_objects, statistic, period):
+        # Note: all input parameters need to be defined as attributes on `self`
+        # so that marshmallow can serialise the object correctly.
+        self.total_network_objects = total_network_objects
+        self.statistic = statistic
+        self.period = period
+
+        if "period" not in self.total_network_objects:
+            self.total_network_objects["period"] = self.period
```
greenape: This won't work, because by this point `total_network_objects` has been deserialised by marshmallow and is an instance of `TotalNetworkObjects` exposed (so, regular object, not a dict).

flowmachine/flowmachine/core/server/query_schemas/aggregate_network_objects.py
OwlHute: Yes I noticed the multiple "if" where the default ANO "period" is stepped up a notch relative to the TNO "by".

flowclient/flowclient/client.py
OwlHute: Because without it the tests error as follows:

> E           flowclient.client.FlowclientConnectionError: Something went wrong. API returned with status code 400.
> Error message: 'Internal flowmachine server error: could not create query object using query schema.
> The original error was: '__init__() missing 1 required keyword-only argument: 'period'''.
> Payload: {'orig_error_msg': "__init__() missing 1 required keyword-only argument: 'period'",
> 'params': {'period': 'day', 'query_kind': 'aggregate_network_objects', 'statistic': 'median', 
> 'total_network_objects': {'aggregation_unit': 'admin3', 'end_date': '2016-01-02', 'query_kind': 'total_network_objects', 'start_date': '2016-01-01'}}}

flowmachine/flowmachine/core/server/query_schemas/aggregate_network_objects.py
OwlHute: Yes, I have removed this

flowclient/flowclient/client.py
greenape: Solution is to pass in valid json for the nested query, rather than silently patch it up here I think.. if you look at the total_network_objects function (which is what would be supplying the value for the total_network_objs argument), it has a default value of ‘day’ for the period.

OwlHute:
To move forward with this now, I think the best thing is for me to push the latest changes (a couple of small ones addressing review comments) and have a sort of audit of the parameters, tracing them from flowclient to flowmachine, with particular reference to these blasted "period" and "by" arguments!  "period" occurs in both AggregateNetworkObjects and TotalNetworkObjects, and I don't think TotalNetworkObjects is defined or used quite right somewhere

integration_tests/tests/query_tests/test_queries.py
```diff
@@ -70,6 +79,19 @@
...
             },
         ),
+        (
+            "aggregate_network_objects",
+            {
+                "total_network_objects": {
+                    "query_kind": "total_network_objects",
+                    "start_date": "2016-01-01",
+                    "end_date": "2016-01-02",
+                    "aggregation_unit": "admin3",
```
greenape: Missing a parameter here.

OwlHute: Ah but the missing parameter "total_by" is supposed to be optional, with a default of "day" !

greenape: Ahhh. Looks like default=‘day’ should actually be missing=‘day’ in the totalnetworlobjects schema.

OwlHute: Yes, now you mention it, I don't see any references to default value setting in custom_fields.py. I'll try running the tests with:

> class TotalNetworkObjectsSchema(Schema):
>      :::
>    total_by = TotalBy(required=False, default="day")

replaced by:

> total_by = TotalBy(required=False)

But the default must be set somewhere.

greenape: https://github.com/Flowminder/FlowKit/blob/89851e2c9198e2f2d6378cd810b14cac9e969582/flowmachine/flowmachine/core/server/query_schemas/total_network_objects.py#L20

That line, `default='day'`, should be `missing='day', required=False` - suspect that's what's going on. (I misread the marshmallow docs when I wrote that bit, evidently :))

flowmachine/tests/server/test_action_handlers.py
```diff
@@ -14,6 +17,21 @@
 from flowmachine.core.server.zmq_helpers import ZMQReplyStatus
 
 
+def test_run_query_type_error(monkeypatch):
```
greenape: I've added this test because the `TypeError` code path wasn't touched, but I think we still want to keep it around for "the dev screwed up" situations, which is pretty much the only time it would occur.

flowmachine/flowmachine/core/server/query_schemas/custom_fields.py
```diff
@@ -46,6 +46,60 @@ def __init__(self, required=False, validate=None, **kwargs):
         )
 
 
+class TotalBy(fields.String):
+    """
+    A string representing a period type, e.g. "day"
+    """
+
+    def __init__(self, required=False, **kwargs):
+        # if validate is not None:
```
greenape: Can we remove these lines, or at least add an explanation for them being commented out?

flowclient/flowclient/client.py
```diff
@@ -1099,3 +1103,33 @@ def joined_spatial_aggregate(
...
+def aggregate_network_objects(
+    *, total_network_objects: Dict[str, str], statistic: str, aggregate_by: str = "day"
+) -> dict:
+    """
+    Return query spec for aggregate network objects
+
+    Parameters
+    ----------
+    total_network_objects : dict
+        TotalNetworkObjects query result
```
greenape:
```suggestion
        Query spec produced by total_network_objects
```

flowclient/flowclient/client.py
```diff
@@ -1099,3 +1103,33 @@ def joined_spatial_aggregate(
...
+    statistic : str
```
greenape:
```suggestion
    statistic : {"avg", "max", "min", "median", "mode", "stddev", "variance"}
```

flowclient/flowclient/client.py
```diff
@@ -1099,3 +1103,33 @@ def joined_spatial_aggregate(
...
+        Statistic type
+    aggregate_by : str
```
greenape:
```suggestion
    aggregate_by : {"second", "minute", "hour", "day", "month", "year", "century"
```

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 ### Added
+- Added new flowclient API entrypoint, aggregate_network_objects(), to access (with simplified parameters) equivalent flowmachine query [#601](https://github.com/Flowminder/FlowKit/issues/601)
```
```suggestion
- Added new flowclient API entrypoint, `aggregate_network_objects`, to access equivalent flowmachine query [#601](https://github.com/Flowminder/FlowKit/issues/601)
```

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
 
 ### Changed
```
greenape: Can we have a changed entry here as well to reflect that the names of the arguments to `TotalNetworkObjects` and `AggregateNetworkObjects` have been altered?

CHANGELOG.md
```diff
@@ -14,10 +15,12 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [0.5.3]
 ### Added
+- Amended time aggregation parameter names used in existing `total_network_objects` query from "period" in flowclient & "by" in flowmachine to "total_by" in both.
```
greenape: Needs to go in the [Unreleased] section

CHANGELOG.md
```diff
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 ### Added
+- Amended time aggregation parameter names used in existing `total_network_objects` query from "period" in flowclient & "by" in flowmachine to "total_by" in both.
```
greenape: Sorry to be pedantic, but can this go in the changed section of [Unreleased], and can we explicitly refer to `TotalNetworkObjects` and `AggregateNetworkObjects`?

CHANGELOG.md
```diff
@@ -5,8 +5,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 ### Added
+- Added new flowclient API entrypoint, `aggregate_network_objects`, to access equivalent flowmachine query [#601](https://github.com/Flowminder/FlowKit/issues/601)
 
 ### Changed
+- Amended time aggregation parameter names used in existing `TotalNetworkObjects` query from "period" in flowclient & "by" in flowmachine to "total_by" in both.
```
```suggestion
- The `period` argument to `TotalNetworkObjects` in FlowMachine has been renamed `total_by`
- The `period` argument to `total_network_objects` in FlowClient has been renamed `total_by`
```

CHANGELOG.md
```diff
@@ -5,8 +5,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
+  Likewise time aggregation parameter name in new `AggregateNetworkObjects` query is "aggregate_by", also used in both flowclient & flowmachine [#601](https://github.com/Flowminder/FlowKit/issues/601)
```
```suggestion
  - The `by` argument to `AggregateNetworkObjects` in FlowMachine has been renamed to `aggregate_by`
```

CHANGELOG.md
```diff
@@ -5,8 +5,11 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
```
```suggestion
 - The `by` argument to `AggregateNetworkObjects` in FlowMachine has been renamed to `aggregate_by`
```

OwlHute:
The CHANGELOG Unreleased "Changed" entry now reads:

> ### Changed
> - The `period` argument to `TotalNetworkObjects` in FlowMachine has been renamed `total_by`
> - The `period` argument to `total_network_objects` in FlowClient has been renamed `total_by`
> - The `by` argument to `AggregateNetworkObjects` in FlowMachine has been renamed to `aggregate_by`
> - The `by` argument to `AggregateNetworkObjects` in FlowMachine has been renamed to `aggregate_by

So I think in that 4th line "FlowMachine" should read "FlowClient"

CHANGELOG.md
```diff
@@ -5,19 +5,23 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 ### Added
+- Added new flowclient API entrypoint, `aggregate_network_objects`, to access equivalent flowmachine query [#601](https://github.com/Flowminder/FlowKit/issues/601)
 
 ### Changed
-
+- The `period` argument to `TotalNetworkObjects` in FlowMachine has been renamed `total_by`
+- The `period` argument to `total_network_objects` in FlowClient has been renamed `total_by`
+  - The `by` argument to `AggregateNetworkObjects` in FlowMachine has been renamed to `aggregate_by`
```
```suggestion
 - The `by` argument to `AggregateNetworkObjects` in FlowMachine has been renamed to `aggregate_by`
```

CHANGELOG.md
```diff
@@ -5,19 +5,23 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
+ - The `by` argument to `AggregateNetworkObjects` in FlowMachine has been renamed to `aggregate_by`
```
```suggestion
```

PR #609, opened 2019-04-11
Add spatial_aggregate api endpoint

### LINKED ISSUE #599: Add SpatialAggregate to api

### LINKED ISSUE #603: TotalNetworkObjects api route is actually AggregateNetworkObjects
At present, because server.py defaults to calling `aggregate()` on everything, the thing referring to itself as `TotalNetworkObjects` is actually `AggregateNetworkObjects`. This will be fixed by #599, at which point the `aggregate()` call will no longer happen.

greenape:
Closes #599

### I have:

### Description

- Adds a `spatial_aggregate` endpoint
- Adds a `spatial_aggregate` function to flowclient
- Disallows using `daily_location` or `modal_location` as top level queries
- Makes all flowclient funcs kw only
- Makes `modal_location` take a list of daily locations instead of varargs
- Removes the implicit call to aggregate
- Fixes #603 
- Makes `AggregateNetworkObjects` take a `TotalNetworkObjects` argument (this turned out to be necessary to #603, which was necessitated by removing the call to aggregate, as it turns out `TotalNetworkObjects` couldn't actually be written to cache because `index_cols` was mistakenly including subscriber)


flowapi/tests/unit/test_run_query.py
```diff
@@ -42,13 +42,13 @@
...
+                "query_kind": "spatial_aggregation",
+                "locations": {"query_kind": "daily_location", "date": "2016-01-01"},
+            },
             "Aggregation unit must be specified when running a query.",
         ),
-        (
-            {"params": {"date": "2016-01-01"}},
-            "Query kind must be specified when running a query.",
-        ),
+        ({"date": "2016-01-01"}, "Query kind must be specified when running a query."),
```
maxalbert: Do we want/need an additional test case where the toplevel query kind is `"spatial_aggregation"` but then the query kind is missing inside the `"locations"`? When I try this locally I currently get a 401 but presumably we want 400 as for the other cases in this test?

flowmachine/flowmachine/core/query.py
```diff
@@ -941,7 +941,7 @@ def index_cols(self):
         try:
             if self.subscriber_identifier in cols:
                 ixen.append(self.subscriber_identifier)
-            else:
+            elif "subscriber" in cols:
```
maxalbert: Can we please have an explicit `else` branch (which presumably should raise an error)?

flowmachine/flowmachine/core/server/query_schemas/base_exposed_query.py
```diff
@@ -88,35 +64,3 @@ def query_id(self):
...
-        #
-        #     DailyLocationSchema -> DailyLocationExposed -> BaseExposedQuery -> FlowmachineQuerySchema -> DailyLocationSchema
-        #
-        # We break the dependency BaseExposedQuery -> FlowmachineQuerySchema here by using
-        # this dynamic import. There is possibly a better way to do this...
-        from flowmachine.core.server.query_schemas import FlowmachineQuerySchema
-
-        return FlowmachineQuerySchema().dump(self)
-
-    def _create_query_info_lookup(self):
```
maxalbert: Hmm, I was certain we needed this (or at least _wanted_ to create the query info lookup here) but you are correct that it doesn't seem to be used. As far as I can tell, the query lookup creation currently happens in the run_query action handler. @greenape Just to confirm you think that's the right place for it to happen (so that it's fine to delete this here)?

flowmachine/flowmachine/core/server/query_schemas/base_exposed_query.py
```diff
@@ -88,35 +64,3 @@ def query_id(self):
         # by flowmachine.core.Query.md5, and the latter is currently being used by
         # the QueryStateMachine, so we need to use it to check the query state.
         return self._flowmachine_query_obj.md5
-
-    @property
-    def query_params(self):
```
maxalbert: Fine with me to delete this (because we can't use it in its current form; see the comment in the `query_id` property above), but I'd be curious what your thinking is in general about this. Do you see a chance that in the future we may derive the query id purely from the query parameters (at least for exposed queries)? Or is this not possible/desirable?

flowmachine/flowmachine/core/server/query_schemas/base_exposed_query.py
greenape: If we turn out to need it, we can put this back in. I figure right now, more of a liability have a dead code path hanging about.

flowmachine/flowmachine/core/server/query_schemas/base_exposed_query.py
greenape: Well, if we want to continue to support flowmachine as a library, the method of deriving ids needs to be viable in both cases.

flowmachine/flowmachine/core/server/query_schemas/flowmachine_query.py
```diff
@@ -49,36 +51,5 @@ class FlowmachineQuerySchema(OneOfSchema):
         "location_introversion": LocationIntroversionSchema,
         "total_network_objects": TotalNetworkObjectsSchema,
         "dfs_metric_total_amount": DFSTotalMetricAmountSchema,
+        "spatial_aggregate": SpatialAggregateSchema,
     }
-
-    def get_obj_type(self, obj):
```
maxalbert: Kinda nice that this boilerplate can be removed as a consequence of removing the `query_params` property (because this is only needed by the call to `FlowmachineQuerySchema().dump()`. 👍

flowmachine/flowmachine/core/server/query_schemas/spatial_aggregate.py
```diff
@@ -0,0 +1,51 @@
...
+from flowmachine.features.utilities.spatial_aggregates import SpatialAggregate
+from .base_exposed_query import BaseExposedQuery
+from .daily_location import DailyLocationSchema, DailyLocationExposed
+from .modal_location import ModalLocationSchema, ModalLocationExposed
+from .custom_fields import AggregationUnit
+
+__all__ = ["SpatialAggregateSchema", "SpatialAggregateExposed"]
+
+
+class InputToSpatialAggregate(OneOfSchema):
```
maxalbert: At some point we should perhaps come up with more meaningful names for these "input" schemas. But the only alternative I can think of off the top of my head is `SpatiallyAggregatableQuery`, which isn't really much better. 😄

flowmachine/flowmachine/core/server/utils.py
```diff
@@ -53,11 +53,14 @@ def send_zmq_message_and_receive_reply(msg, port=5555, host="localhost"):
 FM_EXAMPLE_MESSAGE = {
     "action": "run_query",
     "params": {
-        "query_kind": "daily_location",
-        "date": "2016-01-01",
-        "method": "last",
-        "aggregation_unit": "admin3",
-        "subscriber_subset": None,
+        "query_kind": "spatial_aggregate",
+        "locations": {
```
maxalbert: It just occurred to me that when reading this spec, I expected the value of the `"locations"` key to be a list because "locations" is plural. I don't think making it singular makes sense, but something to keep in mind because it may lead to confusion. Maybe we want to rename it to `"locations_query"` in the future?

flowmachine/flowmachine/features/network/total_network_objects.py
```diff
@@ -249,39 +220,8 @@ class AggregateNetworkObjects(GeoDataMixin, Query):
 
     """
 
-    def __init__(
-        self,
-        start=None,
```
maxalbert: Oh yeah baby! Delete those params... 💥💥💥

integration_tests/tests/flowmachine_server_tests/test_action_run_query.py
```diff
@@ -65,12 +67,12 @@
     #
 
     first_few_rows_expected = [
-        ("524 3 09 50", 18),
-        ("524 5 13 67", 17),
+        ("524 1 01 04", 13),
```
maxalbert: Why do we expect these values to change? Shouldn't the internal calculation still yield the same result?

flowmachine/flowmachine/core/server/utils.py
greenape: Maybe, yeah. I had exactly that originally, but decided that because it is the locations of multiple _subscribers_ that the plural made sense.

integration_tests/tests/flowmachine_server_tests/test_action_run_query.py
greenape: They do! But they are now in a consistent order, so the first three rows are different.

integration_tests/tests/utils.py
```diff
@@ -35,7 +39,9 @@ def make_token(username, secret_key, lifetime, claims):
     )
 
 
-query_kinds = ["daily_location", "modal_location", "flows"]
+query_kinds = list(FlowmachineQuerySchema.type_schemas.keys()) + list(
```
maxalbert: I feel like the right-hand side is useful enough to add to flowmachine itself. Feel free to add to this PR, or we can create a separate issue for it.

flowmachine/flowmachine/core/server/utils.py
maxalbert: Yes, I feel the plural in locations is appropriate. Would just be nice from a user's perspective to have a name that makes it reasonably obvious what kind of value is expected.

flowclient/flowclient/client.py
```diff
@@ -835,10 +858,11 @@ def meaningful_locations_between_dates_od_matrix(
 
 
 def modal_location(
-    *locations: Dict[str, Union[str, Dict[str, str]]], aggregation_unit: str
+    *, locations: Dict[str, Union[str, Dict[str, str]]], aggregation_unit: str
```
maxalbert: Should the first `Dict` be a `List`?

flowapi/flowapi/user_model.py
```diff
@@ -76,16 +77,44 @@ def has_access(
...
+        -------
+        str, str
+            Query kind and aggregation unit
+
+        """
+
+        try:
+            query_kind = query_json["query_kind"]
+            if query_kind == "spatial_aggregate":
+                return self._get_query_kind_and_aggregation_unit(
```
maxalbert: Hmm, here we are extracting the aggregation unit from the input locations query. Is this really what we want in all cases? It works for e.g. daily/modal locations, but shouldn't it be possible to do a spatial aggregation on something that was previously non-aggregated? Or, say, to run a further aggregation to admin2-level of daily locations that are on admin3 level?

I don't consider this a blocker for this PR but I think I'd be much happier if we added a separate `aggregation_unit` parameter to the `spatial_aggregate` query itself, and only did some consistency checks with aggregation unit of the input query.

greenape: Well at this time, `SpatialAggregate` doesn't support that, so it would seem a bit moot really. I think should we add support for a different approach to spatial aggregation, we'd want to revisit and see if we could find a better solution.

PR #610, opened 2019-04-11
Dev setup cleanup

### LINKED ISSUE #265: Better development setup/workflow needed

### LINKED ISSUE #467: Integration tests fail with default flowdb/redis port values in .env file

### LINKED ISSUE #485: Inconsistent use of environment variable name for connecting with FlowDB

### LINKED ISSUE #493: Integration tests should use env vars FLOWAPI_HOST and FLOWAPI_PORT to connect to flowapi
The pytest fixture `api_host` in the integration tests currently returns one of the two hard-coded strings `"http://flowapi:9090"` or `"http://localhost:9090"` to determine the url on which flowapi is running. This is a legacy from a while ago and should really use the environment variables `FLOWAPI_HOST` and `FLOWAPI_PORT`.

### LINKED ISSUE #532: Default environment variables in FlowMachine Dockerfile
The FlowMachine `Dockerfile` and `Dockerfile-dev` set default values for the environment variables `LOG_LEVEL`, `FLOWDB_PORT`, `FLOWDB_HOST`, `FLOWDB_USER`, `FLOWDB_PASS`, and `DEBUG`. These all have defaults in the code (some of which differ from the values in the Dockerfile, which can be confusing when looking for the default value).

We shouldn't set any defaults in the Dockerfile - there are defaults in the code, which can be overridden in the `docker-compose` files where necessary. `FLOWDB_PASS` shouldn't have a default anywhere, except in `docker-compose-dev.yml`.

### LINKED ISSUE #533: Docker-compose files are not working
The top-level `docker-compose.yml` and `docker-compose-dev.yml` currently fail to spin up a working FlowKit system. There are a few reasons for this:

- We use the default `FLOWDB_PASS` password in the FlowMachine container. This is set to `flowflow` in the Dockerfile, which is inconsistent with the value we set in the FlowDB container. As mentioned in #532, this default should be removed and we should instead set a default value in `docker-compose-dev.yml`, and require the user to set the value in `docker-compose.yml`.
- We set the `SERVER` env var in the FlowAPI container, but this should be `FLOWMACHINE_SERVER`.
- The parameters for `flowdb_synthetic_data` are incompatible with the new synthetic data generator - `N_SITES` defaults to 1000, which is greater than `N_CELLS=500`.
- If the FlowMachine container starts up before FlowDB is ready, it errors but the container stays running, so doesn't keep restarting until FlowDB is ready. I'm not sure why this is happening.

greenape:
Closes #533 closes #532 closes #493 closes #467 closes #265 closes #485 

### I have:

### Description

- Puts all the env vars in one, visible file at the project root
- Removes the dev Dockerfiles that we don't use
- Removes all the compose files we don't use
- Uses the makefile in the pipenv scripts for running integration tests and building docs
- Harmonises and disambiguates the variously named env vars
- Removes _all_ nonessential defaults from Dockerfiles


docs/source/dev_environment_setup.md
```diff
@@ -65,6 +65,8 @@ will try to find. Currently only `opendiff` is listed there (but we may add addi
 manually add it to this file. See the ApprovalTests [README](https://github.com/approvals/ApprovalTests.Python) for
 examples (illustrative particularly for file paths on Windows).
 
+!!! warning
```
maxalbert: Hmm, at least on Github this warning is kinda lost at the bottom of the "Diff tool" subsection. How about moving this prominently towards the top of this file, or into its own sub-section?

Also, the `!!!` doesn't play nicely with Github's markup rendering. I assume it will displayed nicely in the rendered docs, but maybe we can find a way of displaying it which works well in both? Again, perhaps a subsection title including the word "warning" is a good compromise?

flowmachine/tests/test_init.py
```diff
@@ -48,7 +48,7 @@ def test_param_priority(mocked_connections, monkeypatch):
     # Use monkeypatch to set environment variable only for this test
     monkeypatch.setenv("LOG_LEVEL", "DUMMY_ENV_LOG_LEVEL")
     monkeypatch.setenv("FLOWDB_PORT", "7777")
-    monkeypatch.setenv("FLOWDB_USER", "DUMMY_ENV_FLOWDB_USER")
+    monkeypatch.setenv("FLOWMACHINE_DB_USER", "DUMMY_ENV_FLOWDB_USER")
```
jch1g10:
```suggestion
    monkeypatch.setenv("FLOWMACHINE_FLOWDB_USER", "DUMMY_ENV_FLOWDB_USER")
```

secrets_quickstart/docker-stack.yml
```diff
@@ -47,10 +47,10 @@ services:
             - FLOWDB_PORT=5432
             - FLOWDB_HOST=flowdb
             - LOG_LEVEL=${LOG_LEVEL:-error}
-            - DEBUG=${FM_DEBUG:-False}
+            - FLOWMACHINE_SERVER_DEBUG_MODE=${FLOWMACHINE_SERVER_DEBUG_MODE:-False}
         secrets:
           - FLOWMACHINE_DB_USER
```
jch1g10:
```suggestion
          - FLOWMACHINE_FLOWDB_USER
```

secrets_quickstart/secrets-quickstart.sh
```diff
@@ -12,18 +12,18 @@ echo "Removing existing secrets_test_stack"
 docker stack rm secrets_test
 # Remove existing secrets
 echo "Removing existing secrets"
-docker secret rm FLOWMACHINE_DB_PASS
+docker secret rm FLOWMACHINE_FLOWDB_PASS
 docker secret rm FLOWMACHINE_DB_USER
```
jch1g10:
```suggestion
docker secret rm FLOWMACHINE_FLOWDB_USER
```

secrets_quickstart/secrets-quickstart.sh
```diff
@@ -12,18 +12,18 @@ echo "Removing existing secrets_test_stack"
...
 docker secret rm FLOWAPI_DB_PASS
```
jch1g10:
```suggestion
docker secret rm FLOWAPI_FLOWDB_PASS
```

secrets_quickstart/secrets-quickstart.sh
```diff
@@ -12,18 +12,18 @@ echo "Removing existing secrets_test_stack"
...
-docker secret rm FLOWAPI_DB_USER
+docker secret rm FLOWAPI_FLOWDB_USER
 docker secret rm POSTGRES_PASSWORD_FILE
 docker secret rm cert-flowkit.pem
 docker secret rm JWT_SECRET_KEY
 docker secret rm REDIS_PASSWORD_FILE
 echo "Adding secrets"
-openssl rand -base64 16 | docker secret create FLOWMACHINE_DB_PASS -
+openssl rand -base64 16 | docker secret create FLOWMACHINE_FLOWDB_PASS -
 echo "fm" | docker secret create FLOWMACHINE_DB_USER -
```
jch1g10:
```suggestion
echo "fm" | docker secret create FLOWMACHINE_FLOWDB_USER -
```

secrets_quickstart/secrets-quickstart.sh
```diff
@@ -12,18 +12,18 @@ echo "Removing existing secrets_test_stack"
...
-echo "api" | docker secret create FLOWAPI_DB_USER -
+echo "api" | docker secret create FLOWAPI_FLOWDB_USER -
 openssl rand -base64 16 | docker secret create FLOWAPI_DB_PASS -
```
jch1g10:
```suggestion
openssl rand -base64 16 | docker secret create FLOWAPI_FLOWDB_PASS -
```

development_environment
```diff
@@ -0,0 +1,124 @@
...
+# To avoid mpl errors
+MPLBACKEND="agg" 
+# Password for redis
+REDIS_PASSWORD=fm_redis 
+# Error and debugging lo level
+LOG_LEVEL=debug 
+# Flowmachine db user
+FLOWMACHINE_FLOWDB_USER=analyst 
+# Flowmachine db password
+FLOWMACHINE_FLOWDB_PASS=foo
```
jch1g10: `FLOWMACHINE_FLOWDB_USER` and `FLOWMACHINE_FLOWDB_PASS` are also set below in the "FlowDB Settings" section

docs/source/2-install.md
```diff
@@ -219,9 +219,9 @@ conn = flowclient.Connection("https://localhost:9090", "JWT_STRING", ssl_certifi
 cd secrets_quickstart
 docker login
 docker swarm init
-openssl rand -base64 16 | docker secret create FLOWMACHINE_DB_PASS -
+openssl rand -base64 16 | docker secret create FLOWMACHINE_FLOWDB_PASS -
 echo "fm" | docker secret create FLOWMACHINE_DB_USER -
```
jch1g10:
```suggestion
echo "fm" | docker secret create FLOWMACHINE_FLOWDB_USER -
```

docs/source/2-install.md
```diff
@@ -219,9 +219,9 @@ conn = flowclient.Connection("https://localhost:9090", "JWT_STRING", ssl_certifi
...
-echo "api" | docker secret create FLOWAPI_DB_USER -
+echo "api" | docker secret create FLOWAPI_FLOWDB_USER -
 openssl rand -base64 16 | docker secret create FLOWAPI_DB_PASS -
```
jch1g10:
```suggestion
openssl rand -base64 16 | docker secret create FLOWAPI_FLOWDB_PASS -
```

docs/source/2-install.md
```diff
@@ -219,9 +219,9 @@ conn = flowclient.Connection("https://localhost:9090", "JWT_STRING", ssl_certifi
 cd secrets_quickstart
 docker login
 docker swarm init
-openssl rand -base64 16 | docker secret create FLOWMACHINE_DB_PASS -
+openssl rand -base64 16 | docker secret create FLOWMACHINE_FLOWDB_PASS -
 echo "fm" | docker secret create FLOWMACHINE_DB_USER -
-echo "api" | docker secret create FLOWAPI_DB_USER -
+echo "api" | docker secret create FLOWAPI_FLOWDB_USER -
```
jch1g10: Use "analyst" and "reporter" instead of "fm" and "api" here, for consistency with what we use elsewhere?

flowapi/flowapi/config.py
```diff
@@ -37,8 +37,8 @@ def get_config():
     flowmachine_server = os.getenv("FLOWMACHINE_SERVER", "localhost")
     flowmachine_port = os.getenv("FLOWMACHINE_PORT", "5555")
 
-    flowdb_user = getsecret("API_FLOWDB_USER", os.getenv("FLOWDB_USER"))
-    flowdb_password = getsecret("FLOWAPI_DB_PASS", os.getenv("FLOWDB_PASS"))
+    flowdb_user = getsecret("FLOWAPI_FLOWDB_USER", os.getenv("FLOWAPI_FLOWDB_USER"))
+    flowdb_password = getsecret("FLOWAPI_DB_PASS", os.getenv("FLOWAPI_DB_PASS"))
```
jch1g10:
```suggestion
    flowdb_password = getsecret("FLOWAPI_FLOWDB_PASS", os.getenv("FLOWAPI_FLOWDB_PASS"))
```

flowdb/bin/build/last_create_roles.sh
```diff
@@ -18,12 +18,12 @@ set -e
 #               This user is mainly designed for ingesting 
 #               data into the events table.
 #
-#   * $FM_USER:  the for $FM_USERs that need access to raw data.
+#   * $FLOWMACHINE_FLOWDB_USER:  the for $FLOWMACHINE_FLOWDB_USERs that need access to raw data.
```
jch1g10:
```suggestion
#   * $FLOWMACHINE_FLOWDB_USER:  for users that need access to raw data.
```

flowdb/bin/build/last_create_roles.sh
```diff
@@ -41,134 +41,134 @@ fi
 if [ -e /run/secrets/FLOWMACHINE_DB_USER ];
 then
     echo "Using secrets for analyst user."
-    FM_USER=$(< /run/secrets/FLOWMACHINE_DB_USER)
+    FLOWMACHINE_FLOWDB_USER=$(< /run/secrets/FLOWMACHINE_DB_USER)
```
jch1g10:
```suggestion
    FLOWMACHINE_FLOWDB_USER=$(< /run/secrets/FLOWMACHINE_FLOWDB_USER)
```

flowdb/bin/build/last_create_roles.sh
```diff
@@ -41,134 +41,134 @@ fi
...
     echo "Using secrets for reporter user."
-    API_USER=$(< /run/secrets/FLOWAPI_DB_USER)
+    FLOWAPI_FLOWDB_USER=$(< /run/secrets/FLOWAPI_FLOWDB_USER)
 fi
 
 if [ -e /run/secrets/FLOWAPI_DB_PASS ];
 then
     echo "Using secrets for reporter password."
-    API_PASSWORD=$(< /run/secrets/FLOWAPI_DB_PASS)
+    FLOWAPI_FLOWDB_PASS=$(< /run/secrets/FLOWAPI_DB_PASS)
```
jch1g10:
```suggestion
    FLOWAPI_FLOWDB_PASS=$(< /run/secrets/FLOWAPI_FLOWDB_PASS)
```

docs/README.md
```diff
@@ -9,12 +9,14 @@ FlowKit's documentation is built using [Mkdocs](http://mkdocs.org).
 - Docker
 - Pipenv
 
-To build the docs, first install the necessary dependencies by running `pipenv install` (in the `docs/` subfolder). Once this completes, you can build the docs by running `pipenv run build`.
+To build the docs, first install the necessary dependencies by running `pipenv install` (in the `docs/` subfolder). Once this completes, you can build the docs by running `pipenv run build`, after sourcing the top level `development_environment` file.
```
jch1g10: Sourcing `development_environment` is not sufficient - this will only set shell variables, not export them as env vars.

greenape: set -a &&. ./development_environment && set +a is what’s actually needed I think 🤔

docs/source/2-install.md
greenape: Would be inclined to change elsewhere to reflect this rather than other way round.

docs/source/dev_environment_setup.md
greenape: Well, we use these elsewhere in the docs (they’re rendered by the admonition markdown extension). Is there much benefit in sticking to github flavoured markdown given these are for the docs site?

maxalbert: Yeah you're right, probably not.

docs/README.md
jch1g10: Yep, that works. Probably worth mentioning this in the docs.

docs/source/dev_environment_setup.md
```diff
@@ -65,6 +65,8 @@ will try to find. Currently only `opendiff` is listed there (but we may add addi
...
+    The project root contains an env file (`development_environment`), in which are default values for _all_ of the environment variables used to control and configure the FlowDB, FlowAPI, FlowMachine and FlowAuth. You will need to source this file, or otherwise set the environment variables before running _any other command_.
```
jch1g10:
```suggestion
    The project root contains an env file (`development_environment`), in which are default values for _all_ of the environment variables used to control and configure the FlowDB, FlowAPI, FlowMachine and FlowAuth. You will need to source this file (`set -a && . ./development_environment && set +a`), or otherwise set the environment variables before running _any other command_.
```

jch1g10:
FlowMachine and FlowAPI both use the same env var (`LOG_LEVEL`) to set the log level, so the log level will always be the same for both. Do we imagine any situations in which we may want the log levels of these components to differ?

maxalbert:
@jch1g10 I think they _should_ be independently configurable (since they are configuration options for different components), even if they will typically be the same. So I'd suggest renaming them to `FLOWMACHINE_LOG_LEVEL` and `FLOWAPI_LOG_LEVEL`.

jch1g10:
`FLOWMACHINE_FLOWDB_USER` and `FLOWAPI_FLOWDB_USER` have defaults in `flowdb/Dockerfile`. Do we want this? (if so, I'd suggest changing them from `"analyst"` and `"reporter"` to `"flowmachine"` and `"flowapi"`, as @greenape suggested above)

jch1g10:
I ran into some issues trying to spin everything up because `make up` was building `flowdb-testdata` but not `flowdb`. I've modified the makefile so that `make up` will now build everything, but building flowdb takes a while so I've also added a new `up-no_build` target.

I also added `FLOWMACHINE_LOG_LEVEL` and `FLOWAPI_LOG_LEVEL` env vars, removed some duplicates from `development_environment` and added `FLOWMACHINE_FLOWDB_USER` and `FLOWAPI_FLOWDB_USER` to the flowdb environment in `docker-compose.yml`.

docs/README.md
jch1g10:
```suggestion
To build the docs, first install the necessary dependencies by running `pipenv install` (in the `docs/` subfolder). Once this completes, you can build the docs by running `pipenv run build`, after sourcing the top level `development_environment` file by running `set -a && . ./development_environment && set +a`.
```

integration_tests/README.md
```diff
@@ -14,9 +14,13 @@ pipenv install
...
 
-This will pull all necessary docker images, start the containers, bring up instances of FlowMachine (listening on port `5555`) and FlowAPI (listening on port `9090`). After the test suite has been run, the containers will be shut down agan.
+This will pull all necessary docker images, start the containers and bring up instances of FlowMachine and FlowAPI. After the test suite has been run, the containers will be shut down agan.
 
 
 If you are using an alternative environment manager, you should install the small number of packages listed in the Pipfile before running pytest.
 
-The test suite makes use of a `.env` file, found in this directory. The values in this `.env` are used both by the test suite, and to create the docker containers. `docker-compose` will make use of the `.env` automatically, but you may need to explicitly supply the values if you are using a different method to spin up containers.
\ No newline at end of file
+The test suite makes use of the environment variables defined in `development_environment` in the project root. You will need to source these variables before running the tests.
```
jch1g10:
```suggestion
The test suite makes use of the environment variables defined in `development_environment` in the project root. You will need to source these variables before running the tests, by running `set -a && . ./development_environment && set +a` in the project root.
```

jch1g10:
I'd suggest that we rename `FLOWMACHINE_SERVER` and `REDIS_PASSWORD` to `FLOWMACHINE_HOST` and `REDIS_PASS`, to be consistent with the naming of other env vars

jch1g10:
> I'd suggest that we rename `FLOWMACHINE_SERVER` and `REDIS_PASSWORD` to `FLOWMACHINE_HOST` and `REDIS_PASS`, to be consistent with the naming of other env vars

Or maybe change the other `*_PASS` to `*_PASSWORD`

jch1g10:
The `*_HOST` and `*_PORT` env vars in `development_environment` are slightly confusing, because they set the hosts and ports for connecting to services from outside the Docker containers and are not actually used to set the corresponding env vars inside the containers (which are hard-coded inside the containers). On balance I think this is the cleanest way of doing things, because this way the correct env vars will be set if we start up flowmachine or flowapi outside of a container, connecting to a running flowdb container.

One slightly more complicated issue is `FLOWDB_HOST` which, by this reasoning, should be set to `localhost` in `development_environment`, and `FLOWDB_HOST` in the flowmachine and flowapi containers should be hard-coded in `docker-compose.yml` to match the name of the flowdb container. However, we have 3 flowdb containers defined in `docker-compose.yml`, and `Makefile` sets the `FLOWDB_HOST` env var to select which one flowmachine and flowapi connect to.

As a workaround, I've changed `docker-compose.yml` to set `FLOWDB_HOST` in the flowmachine and flowapi containers to the value of the `DOCKER_FLOWDB_HOST` env var, and changed `Makefile` to set this env var instead of `FLOWDB_HOST`. I don't think it makes sense to set a default value for `DOCKER_FLOWDB_HOST` in `development_environment`, so I haven't.

jch1g10:
I've changed the `FLOWMACHINE_FLOWDB_USER` and `FLOWAPI_FLOWDB_USER` defaults from "analyst" and "reporter" to "flowmachine" and "flowapi".

jch1g10:
The env var `POSTGRES_DB` sets the name of the FlowDB database, but flowmachine and flowapi are hard-coded to connect to a database called "flowdb".  It would seem sensible to add a flowmachine env var for the name of the flowdb database, but what should we call it? It would be nice to use the same env var in each container, but it has to be called `POSTGRES_DB` in flowdb and this doesn't seem like a great name for the corresponding flowmachine/flowapi env var.

Alternatively we could keep the name hard-coded in flowmachine and flowapi, but then there's not much point setting `POSTGRES_DB` in `development_environment`, since changing it will result in a broken setup.

maxalbert:
@jch1g10 I can't see why we would ever want to make the db name configurable, so I think have a slight preference for simply hard-coding the value. On the other hand, there is an argument for not hard-coding it to make any code referring to it a bit more readable. In the latter case, we can still accept a more meaningfully named env var as an alias and simply set the value of `POSTGRES_DB` from it.

jch1g10:
> @jch1g10 I can't see why we would ever want to make the db name configurable, so I think have a slight preference for simply hard-coding the value. On the other hand, there is an argument for not hard-coding it to make any code referring to it a bit more readable. In the latter case, we can still accept a more meaningfully named env var as an alias and simply set the value of `POSTGRES_DB` from it.

I think I'd agree that there's no need to make this configurable. in which case I'd suggest we remove `POSTGRES_DB` from `development_environment` and `docker-compose.yml` (I know we'd like to keep all env vars in `development_environment`, but this is a Postgres env var rather than a FlowKit one, so I don't see this as a problem).

I'd also then suggest we remove `ARG POSTGRES_DB=flowdb` from the FlowDB `Dockerfile`, and instead just set `ENV POSTGRES_DB=flowdb`.

CHANGELOG.md
```diff
@@ -7,10 +7,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Added
 
 ### Changed
+- All environment variables are now in a single `development_environment` file in the project root, development environment setup has been simplified
+- Default FlowDB users for FlowMachine and FlowAPI have changed from "analyst" and "reporter" to "flowmachine" and "flowapi", respectively
 
 ### Fixed
 
 ### Removed
+- Docs and integration tests now use top level compose file
```
greenape: Shouldn't that go in Changed?

greenape:
Thanks @jch1g10, much better than it was when I left it!

I wonder if the CHANGELOG entries are a little sparse now, but otherwise I'm happy.

PR #611, opened 2019-04-11
Joined spatial aggregate

### LINKED ISSUE #600: Add JoinedSpatialAggregate to API
`JoinedSpatialAggregate` combines one reference location (daily location, home location or similar) with any subscriber metric, to allow returning it aggregated at a spatial level. This should be exposed via the api.

- [x] Add marshmallow schema
- [x] Update claims checking

greenape:
Closes #600

### I have:

### Description

- Adds a new `joined_spatial_aggregate` api endpoint, which takes a location type (anything admissible to `spatial_aggregate`), and a metric (none currently available, but anything with a subscriber column), and returns the result of joining one to the other, and calculating a stat over the metric as grouped by location.
- `JoinedSpatialAggregate` is now kwarg only

As with `spatial_aggregate`, permission for this endpoint is indirectly determined, but in this case it is the intersection of the permissions on the metric and location queries.


flowapi/flowapi/user_model.py
```diff
@@ -88,24 +93,39 @@ def _get_query_kind_and_aggregation_unit(self, query_json: dict) -> Tuple[str, s
...
+            raise MissingQueryKindError
+        try:
             if query_kind == "spatial_aggregate":
-                return self._get_query_kind_and_aggregation_unit(
+                return self._get_query_kinds_and_aggregation_units(
                     query_json=query_json["locations"]
                 )
-        except KeyError:
-            raise MissingQueryKindError
+            elif query_kind == "joined_spatial_aggregate":
```
maxalbert: Just as a note for future refactorings: the appearance of multiple if/else branches and special case treatments for different query kinds suggests to me that this is not the right place for this logic to live. Will give it some thought, but it should likely be each individual query's responsibility to know which types of permissions it needs and to verify these against given claims.

flowapi/flowapi/user_model.py
```diff
@@ -88,24 +93,39 @@ def _get_query_kind_and_aggregation_unit(self, query_json: dict) -> Tuple[str, s
...
+                location_spec = self._get_query_kinds_and_aggregation_units(
+                    query_json=query_json["locations"]
+                )[0]
+                # Because metric specs don't have an aggregation unit, we stitch on the one
```
maxalbert: Similar to my comment on PR #609, the fact that we need to "stitch on" an aggregation unit suggests that the joined spatial aggregate should really accept this parameter itself. (But again, let's leave this for a future PR.)

flowapi/flowapi/user_model.py
```diff
@@ -88,24 +93,39 @@ def _get_query_kind_and_aggregation_unit(self, query_json: dict) -> Tuple[str, s
 
         Returns
         -------
-        str, str
-            Query kind and aggregation unit
+        list of tuples str, str
+            List of tuples giving query kind and aggregation unit
```
maxalbert: Again, note for future refactorings: the return data structure is getting a bit more complicated than calling code should need to know. There is likely a missing object hidden here which should take on some of the responsibility about which permissions exist and how to validate them.

PR #621, opened 2019-04-15
OpenAPI Spec

### LINKED ISSUE #620: API spec
Now that we have the building blocks of an API spec, it would be nice if an actual spec for FlowAPI were exposed and added to the documentation.

greenape:
Closes #620 

### I have:

### Description

This adds an autogenerated openapi spec for flowapi, plus a nice explorable version of it (both to flowapi, and the docs site).

Makes a minor change to all exposed queries, by adding the `query_kind` as a field in their schemas, to make sure that it (correctly) shows up in the openapi spec.

Also removes the number-the-docs things, because the reason to do that is fixed in the `mktheapidocs` library.


flowapi/flowapi/api_spec.py
```diff
@@ -0,0 +1,248 @@
...
+            "version": __version__,
+            "title": "FlowAPI",
+            "contact": {"email": "flowkit@flowminder.org"},
+            "license": {
+                "name": "MPLv2",
+                "url": "https://www.mozilla.org/en-US/MPL/2.0/",
+            },
+        },
+        "paths": {
+            "/run_query": {
```
maxalbert: I think the route should be `"/run"` instead of `"/run_query"`.

How about leveraging `apispec` to directly add the paths to the spec via the dostring? (See [example](https://apispec.readthedocs.io/en/stable/) in their docs.) This would help to avoid mismatches in the route names, and also document things like status codes in the docstring for the route.

flowapi/flowapi/api_spec.py
```diff
@@ -0,0 +1,248 @@
...
+from quart import Blueprint, request, jsonify, render_template, current_app
+from zmq.asyncio import Socket
+from flowapi import __version__
+
+blueprint = Blueprint("spec", __name__)
+
+
+def remove_discriminators(spec: dict) -> dict:
+    """
+    Remove any discriminator keys from an openapi spec dict.
```
maxalbert: Can we add a sentence to the docstring about the purpose for removing discriminators?

flowapi/flowapi/api_spec.py
```diff
@@ -0,0 +1,248 @@
...
+    Returns
+    -------
+
+    """
+    msg = {"request_id": request_id, "action": "get_query_schemas"}
+    socket.send_json(msg)
+    #  Get the reply.
+    reply = await socket.recv_json()
+    flowmachine_query_schemas = reply["payload"]["query_schemas"]
+    # Need to mark query_kind as a required field
```
maxalbert: What's the reason we need the `query_kind` field?

integration_tests/tests/flowapi_tests/test_api_spec.test_generated_openapi_json_spec.approved.txt
```diff
@@ -0,0 +1,1111 @@
...
+          "aggregation_unit",
+          "end_date",
+          "query_kind",
+          "start_date"
+        ],
+        "type": "object"
+      }
+    }
+  },
+  "info": {
```
maxalbert: When I paste this file into the swagger editor, it complains about a missing version:
```
Structural error at info
should have required property 'version'
missingProperty: version
```

integration_tests/tests/flowapi_tests/test_api_spec.test_generated_openapi_json_spec.approved.txt
```diff
@@ -0,0 +1,1111 @@
...
+          "500": {
+            "description": "Server error."
+          }
+        },
+        "summary": "Get geojson for an aggregation unit"
+      }
+    },
+    "/get/{query_id}": {
+      "get": {
+        "operationId": "get",
```
maxalbert: This `operationId` (`"get"`) is the same as the one for `/geography` above.

maxalbert: (which the swagger editor complains about)

flowapi/flowapi/api_spec.py
maxalbert: Ah I see, it's needed because it's a required input when the user sets a query running and of course we want this to be reflected in the API spec.

docs/build.sh
```diff
@@ -44,6 +44,9 @@ pipenv install
 pipenv run flowmachine &
 echo "Started FlowMachine."
 pipenv run quart run --port 9090 &
+echo "Retrieving API spec"
+sleep 5
+curl http://localhost:9090/api/0/spec/openapi-redoc.json -o source/_static/openapi-redoc.json
```
maxalbert: I'm not entirely sure why, but somehow sleeping and retrieving the spec from the actual running API server feels a little awkward/brittle to me. How about copying the json spec from the `.approved.txt` file in the flowapi tests instead?

flowapi/flowapi/api_spec.py
greenape: We'd need to roll something ourselves there - the apispec flask plugin does not, at this time, support blueprints. Is mostly a wrapper around `yaml_utils.load_yaml_from_docstring` tho, so I'll give it a look.

integration_tests/tests/flowapi_tests/test_api_spec.test_generated_openapi_json_spec.approved.txt
greenape: Version is (at the minute) defined by `flowapi.__version__`, which varies by commit because of versioneer, so I've stripped it out of the ApprovalTests bit.

docs/build.sh
greenape: Approval tests need the version stripped out, which makes the result not a valid spec :(

flowapi/flowapi/api_spec.py
greenape: 🎻

flowapi/flowapi/api_spec.py
```diff
@@ -0,0 +1,121 @@
...
+        The specification object
+
+    """
+    msg = {"request_id": request_id, "action": "get_query_schemas"}
+    socket.send_json(msg)
+    #  Get the reply.
+    reply = await socket.recv_json()
+    flowmachine_query_schemas = reply["payload"]["query_schemas"]
+    # Need to mark query_kind as a required field
+    # this is a workaround because the marshmallow-oneOf plugin strips
```
maxalbert: I get the feeling that there _should_ be some option to `marshmallow-oneofschema` to not strip it out, but it doesn't seem to support it at present. Might be worth raising an issue there.

flowapi/flowapi/api_spec.py
```diff
@@ -0,0 +1,121 @@
...
+        try:
+            func = current_app.view_functions[endpoint_func_name]
+            operations = yaml_utils.load_operations_from_docstring(func.__doc__)
+            if len(operations) > 0:
+                for method, op in operations.items():
+                    op["operationId"] = f"{endpoint_func_name}.{method}"
+                spec.path(
+                    path=rule[
+                        0
+                    ].rule,  # In theory, could have multiple rules that match but will only be a single one here
```
maxalbert: Is it worth checking this explicitly and raise an error if this assumption is false?

docs/build.sh
maxalbert: What if we use a more stable version (e.g. the tagged release `0.5.2` rather than the full version `0.5.2+113.g1d4132f0`) for the API, and don't strip it out in the approval tests?

Or alternatively put the version back in here manually from the `.approved.txt` file instead of retrieving it from the API server directly?

PR #638, opened 2019-04-17
Add messages to QueryState enum

### LINKED ISSUE #523: Add error messages related to each query state to `QueryState`
The action handler `action_handler__get_sql` contains a sequence of if/else statements which [mirrors](https://github.com/Flowminder/FlowKit/blob/master/flowmachine/flowmachine/core/server/action_handlers.py#L213) the possible values of `QueryState`. These messages should be moved into the `QueryState` class so that this big if statement can disappear.

greenape:
Closes #523 

### I have:

### Description

Moves the long form messages about query states into the QueryState enum, and out of the action handler.


flowmachine/flowmachine/core/query_state.py
```diff
@@ -26,13 +26,19 @@ class QueryState(str, Enum):
...
-    KNOWN = "known"
+    QUEUED = ("queued", "is queued.")
+    COMPLETED = ("completed", "is ready.")
+    CANCELLED = ("cancelled", "was cancelled.")
+    EXECUTING = ("executing", "is currently running.")
+    ERRORED = ("errored", "finished with an error.")
+    RESETTING = ("resetting", "is being reset.")
+    KNOWN = ("known", "is known, but has not yet been run.")
+
+    def __new__(cls, name, desc, **kwargs):
```
maxalbert: That's a very clever solution. :)

The only minor downside is that the `description` attribute is not easily discoverable in an interactive IPython session because it's only added upon instance creation (so it doesn't show up in `dir(...)` and IPython can't provide tab-completion for it).

Can we add a line to the class docstring to show an example of the `value` and `description` attribute?

maxalbert: Also, how about removing the period at the end of the descriptions (and putting it at the end of the `msg` string in the action handler instead)? Might make them more easily re-usable in other situations.

greenape: I've promoted description to a property to make it more discoverable.

PR #639, opened 2019-04-17
Use SHA1 for tagging docker builds

### LINKED ISSUE #637: Rerunning from failed on CI
Sometimes, steps of CI fail, or we want to debug them via ssh. At present, this is a PITA for any step that involves one of the flowdb images, because they are tagged using the workflow ID. We should switch to using the commit hash instead, which would allow us to restart from any point in the CI process.

greenape:
Closes #637 

### I have:

### Description

Switches the CI build from tagging docker images with the workflow id (which varies between CI runs of the same commit), to the commit hash (which .. varies between commits). This means that a failed workflow doesn't need to be completely rerun.


maxalbert:
This looks good! Just wondering, I seem to remember that you _wanted_ to be able to distinguish between different CI runs of the same commit in the past. Will this break something else as a consequence, or am I mis-remembering?

greenape:
Want to be able to distinguish between branches, but can't use branch name for the tag because not all git branch names are valid docker tags. Looking at the original issue, I considered both commit hash and workflow id, but went with the latter for no particular reason.

PR #664, opened 2019-04-23
Redis sync

### LINKED ISSUE #636: Flushing Redis DB causes trouble retrieving cached queries
**Describe the bug**
After flushing the Redis DB, the retrieval of cached queries through `q.get_query()` causes a re-computation of the query, that is, a call to `q._make_query()` instead of simply retrieving the cached table. 

This is a bug because  while`q.is_stored` reports as if the query is stored, when the user attempts to fetch it one finds that the query is being re-calculated. 

The problem is basically caused because `get_query` uses redis to check whether the query has been completed, which will be False if the redis has been flushed:

```
...
state_machine = QueryStateMachine(self.redis, self.md5)
state_machine.wait_until_complete()
if state_machine.is_completed and self.connection.has_table(
                schema=schema, name=name
):
    try:
        touch_cache(self.connection, self.md5)
   ...
```

On the other hand, `is_stored` is simpler and only looks at whether the cache table exists:

```
...
try:
    schema, name = self.fully_qualified_table_name.split(".")
    return self.connection.has_table(name, schema)
```

**Product**
Flowamchine

**Version**
master branch

**To Reproduce**
1. Create a new query: `q = Query()`
2. Store the query: `q.store()`
3. Flush the DB: `q.redis.flush_db()`
4. Check that the query is indeed stored: `q.is_stored`
5. Retrieve the query to find to your surprise the query is being recomputed: `q.get_query()`

**Expected behavior**
Cached queries should not be recomputed unless the user explicit asks for it. Flushing redis should not mean that the user wants to reset the cache. Rather, the redis database should be updated accordingly.

The call to `get_query` should set the `QueryStateMachine` to completed if the cache table exists.

### LINKED ISSUE #650: Cache utils should update redis when manipulating cache
Related to #636, the cache utilities should always update redis to correspond with what they've done to the cache, e.g. reseting the cache should reset the query state machines in redis as well.

greenape:
Closes #650 closes #636

### I have:

### Description

- Adds a new `resync_redis_with_cache` function to cache utils, which resets redis and repopulates it with fast-forwarded state machines for queries in the cache
- Makes `reset_cache` reset state machines in redis as well

### DIFF
--- a/flowmachine/flowmachine/core/cache.py
+++ b/flowmachine/flowmachine/core/cache.py
@@ -23,7 +23,7 @@ from flowmachine.core.errors.flowmachine_errors import (
     QueryErroredException,
     StoreFailedException,
 )
-from flowmachine.core.query_state import QueryStateMachine
+from flowmachine.core.query_state import QueryStateMachine, QueryEvent
 
 if TYPE_CHECKING:
     from .query import Query
@@ -217,14 +217,21 @@ def touch_cache(connection: "Connection", query_id: str) -> float:
         raise ValueError(f"Query id '{query_id}' is not in cache on this connection.")
 
 
-def reset_cache(connection: "Connection") -> None:
+def reset_cache(connection: "Connection", redis: StrictRedis) -> None:
     """
-    Reset the query cache. Deletes any tables under cache schema, resets the cache count
-    and clears the cached and dependencies tables.
+    Reset the query cache. Deletes any tables under cache schema, resets the cache count,
+    clears the cached and dependencies tables.
 
     Parameters
     ----------
     connection : Connection
+    redis : StrictRedis
+
+    Notes
+    -----
+    You _must_ ensure that no queries are currently running when calling this function.
+    Any queries currently running will no longer be tracked by redis, and UNDEFINED BEHAVIOUR
+    will occur.
     """
     # For deletion purposes, we ignore Table objects. These either point to something
     # outside the cache schema and hence shouldn't be removed by calling this, or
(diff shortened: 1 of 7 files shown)


flowmachine/flowmachine/core/cache.py
```diff
@@ -239,6 +246,47 @@ def reset_cache(connection: "Connection") -> None:
...
+    Any queries currently running will no longer be tracked by redis, and UNDEFINED BEHAVIOUR
+    will occur.
+    """
+    logger.debug("Redis resync")
+    qry = f"SELECT query_id FROM cache.cached"
+    queries_in_cache = connection.fetch(qry)
+    logger.debug("Redis resync", queries_in_cache=queries_in_cache)
+    redis.flushdb()
+    logger.debug("Flushing redis.")
+    for event in (QueryEvent.QUEUE, QueryEvent.EXECUTE, QueryEvent.FINISH):
```
maxalbert: Not sure if it matters, but is it worth swapping the order of the two for loops, so that the logging messages for each `query_id` are closer together as it's propelled through the different states? Might make the logging output a tiny bit more readable. (Although again, might not matter much because if everything goes well then we're not interested in the logging output and if something goes wrong it'll throw an error anyway.)

flowmachine/tests/conftest.py
```diff
@@ -214,6 +215,10 @@ def get(self, key):
     def keys(self):
         return sorted(self._store.keys())
 
+    def flushdb(self):
+        if self.flush:
```
maxalbert: What's the purpose of this check (and of the `flush` attribute in general)? I can't see `DummyRedis.flush` being set anywhere, so this check will always succeed?

flowmachine/flowmachine/core/cache.py
greenape: My thinking was that if something goes wrong, you'd have everything in the same state which seemed slightly nicer to me.

maxalbert: 👍

flowmachine/tests/conftest.py
greenape: Added a comment, but it is used here: https://github.com/Flowminder/FlowKit/blob/8bc0ebbd901097b66c523a87aa7ed09f0c61dd28/flowmachine/tests/test_cache_utils.py#L459 to simulate a bail-out-because-the-redis-is-being-meddled-with-while-we-try-to-sync scenario.

maxalbert: Ah sorry, I missed that one. Yup, makes sense (and thanks for renaming the attribute, makes it clearer to read 👍 ).

flowmachine/tests/test_cache_utils.py
```diff
@@ -410,3 +413,59 @@ def test_cache_table_exists(flowmachine_connect):
...
+    resync_redis_with_cache(flowmachine_connect, Table.redis)
+    assert (
+        QueryStateMachine(Table.redis, stored_query.md5).current_query_state
+        == QueryState.COMPLETED
+    )
+
+
+def test_cache_reset(flowmachine_connect):
+    """
+    Test that cache and redis are reset.
```
maxalbert:
```suggestion
    Test that cache and redis are both reset.
```

flowmachine/tests/test_cache_utils.py
```diff
@@ -410,3 +413,59 @@ def test_cache_table_exists(flowmachine_connect):
...
+        == QueryState.KNOWN
+    )
+    assert not stored_query.is_stored
+
+
+def test_redis_resync_runtimeerror(flowmachine_connect, dummy_redis, monkeypatch):
+    """
+    Test that a runtime error is raised if redis is being updated from multiple places when trying to resync.
+    """
+    monkeypatch.setattr("flowmachine.core.query.Query.redis", dummy_redis)
```
maxalbert: I may have asked this before, but is it possible/sensible to move the monkeypatching into the `dummy_redis` fixture itself?

Issue #397, opened 2019-02-18
Explore feasibility of trimming docker images where possible


maxalbert:
Some of the docker images are fairly large (e.g. flowmachine is > 1.5 GB). This is typically because they are based on standard images (e.g. `python:3.7`) which in turn may contain a full Debian/Ubuntu distribution, sometimes with additional developer packages included. It would be good to explore which of our images can be based on smaller docker images (such as alpine-based ones). In some cases (e.g. flowbd) this is not easily possible, but for flowmachine and probably others it may be.

greenape:
For ref - the original reason to use the Debian based Postgres for FlowDB was because of the need to compile some extensions and added bits. That's less relevant now, but the Debian image remains necessary because arbitrary users can't be used with the alpine base image.

Should definitely be doable to slim the others, starting with switching them to alpine.

maxalbert:
Yes agreed, it makes sense to keep FlowDB on the Debian. Re slimming the others, will be good to weigh off the trade-offs to see if it's worth the effort. Switching to Alpine will likely have the advantage of much smaller size but may considerably increase the build time because some Python packages need to be compiled from source if there is no wheel available for Alpine (and _may_ also require more maintenance than the Debian/Ubuntu-based ones for the same reason). Certainly useful to investigate the option, though.

maxalbert:
Actually, `passenger-docker` (or possibly `baseimage-docker`) look like an interesting alternative to alpine.

- https://github.com/phusion/passenger-docker
- http://phusion.github.io/passenger-docker/


- https://github.com/phusion/baseimage-docker
- http://phusion.github.io/baseimage-docker/

greenape:
This (alpine, anyway) is very doable for FlowAPI, but not for FlowAuth right now (https://github.com/tiangolo/uwsgi-nginx-flask-docker#temporarily-not-supported). FlowMachine has a dependency on pandas (although I think, not a required one for the docker container), which obviously makes alpine untenable because of the need to build pandas.

PR #665, opened 2019-04-23
Slimmer docker images

greenape:
#397

### I have:

### Description

This slims down the docker images for FlowAPI and FlowMachine.

FlowAPI can be switched to an Alpine base image, so we do that here - this reduces the image size from ~406MB, to ~86MB.
FlowMachine can't be switched to Alpine, because that requires building pandas, but can be switched to Debian-slim, which drops the image size from ~479MB to ~204MB.

At present, the base image for FlowAuth doesn't support python 3.7 + alpine, because uWSGI has not been released with Python 3.7 support for Alpine 3.7 nor for Alpine 3.8. We can however drop the python version for Flowauth to 3.6, without breaking anything or suffering a big performance penalty,  because there's no asyncio magic therein. This drops the size of the FlowAuth image from ~412MB to ~118MB.


flowauth/Dockerfile
```diff
@@ -8,14 +8,21 @@ COPY frontend /
...
+RUN apk update && apk add --virtual build-dependencies build-base postgresql-dev gcc python3-dev musl-dev \
+    libressl-dev libffi-dev && \
+    pip install --no-cache-dir pipenv && pipenv install --clear --deploy --system && \
+    apk del build-dependencies
 ENV STATIC_PATH /app/static
 ENV STATIC_INDEX 1
 ENV FLASK_APP flowauth
+ENV PIPENV_COLORBLIND=1
 WORKDIR /app/static
 COPY --from=builder /build .
```
maxalbert: I'm surprised the `--from=builder` argument doesn't require us to use the `builder` alias in L12 above?

flowapi/Dockerfile
```diff
@@ -2,12 +2,14 @@
 # License, v. 2.0. If a copy of the MPL was not distributed with this
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-FROM python:3.7
+FROM python:3.7-alpine
 
 WORKDIR /flowapi
 COPY Pipfile* /
-RUN pip install pipenv && pipenv install --deploy
+RUN apk update && apk add libzmq && apk add --virtual build-dependencies build-base gcc wget git musl-dev zeromq-dev && \
```
maxalbert: Should we use the builder pattern here too (similar to the `flowauth` dockerfile below)? Or not worth the hassle because the resulting image is already fairly small?

flowauth/Pipfile
```diff
@@ -14,7 +14,7 @@ zxcvbn = "*"
 versioneer = "*"
 passlib = "*"
 argon2_cffi = "*"
-"psycopg2-binary" = "*"
+psycopg2 = "*"
```
maxalbert: Why are we switching back to the non-binary version of `psycopg2`? Is this because we need to build it from source because there is no wheel that matches the alpine distribution?

flowauth/Pipfile
```diff
@@ -23,7 +23,7 @@ pytest-cov = "*"
 "flowauth" = {editable = true, path = "./backend"}
 
 [requires]
-python_version = "3.7"
+python_version = "3.6"
```
maxalbert: I'm ok to drop the version to 3.6 for flowauth for the time being. If someone wants to build/develop flowauth locally this makes it a little less convenient because the version now differs from the one for the other components, so it needs switching via pyenv or similar. However, hopefully we can switch back to 3.7+ soon. Can we open a Github issue so we won't forget to do this?

flowauth/Dockerfile
greenape: L5

PR #667, opened 2019-04-23
Refactor subsetting by hours

### LINKED ISSUE #666: Refactor hours subsetting
Flowmachine allows subsetting queries by restricting the hours of the day. However, the interface for this is not very intuitive and flexible: the `hours` argument to `EventTableSubset` (and most other queries) is simply a tuple of numbers such as `(6, 22)`. It's not clear whether these are full hours and whether the start and/or end are inclusive or not. It's also difficult (or impossible?) to express more fine-grained filtering such as "between 6:30 and 21:45". This logic should be refactored to make the interface more intuitive to the user, allow for better validation of parameters, and to move the code responsible for this into dedicated classes/functions, rather than keeping all the logic in `EventTableSubset`.

This is related to #373, #567.

maxalbert:
Closes #666.

### I have:

### Description

This PR has the following purpose:
- Move the logic for "subsetting by hours" out of `EventTableSubset` and into dedicated classes with smaller, more targeted responsibilities and easier testing.
- Provide a more usable interface for creating "hour slices" which reflects the possibly complicated logic they have (esp. when the hours span midnight).

The core class is `HourInterval`, which represents a half-open interval of hours during the day and allows subsetting of these hours, for example `08:00 <= HH:MM < 16:30`. Such an hour interval can be repeated every day, or alternatively at a given day during the week (e.g. every Tuesday).

When spanning midnight, the hour subsetting is represented as _two_ hour intervals (the first spanning from midnight at the beginning of the day to the start hour, and the second spanning from the stop hour to midnight at the end of the day). The class `HourSlice` represents such a concept of multiple non-overlapping time intervals and is used.

There are a couple of additional helper classes (`HourOfDay`, `DayPeriod` and `DayOfWeekPeriod`) whose main purpose it is to do parameter validation and avoid duplication of filtering logic.

The logic in `EventTableSubset` has been switched to use `HourSlice` in order to filter the sqlalchemy condition. A little bit of boilerplate was added in `__init__()` to initialise an HourSlice from the `hours` argument which is currently passed in by the other query classes. This will be removed soon and other classes will be forced to pass in an `HourSlice` directly, but I'll postpone this to a subsequent PR to avoid this one getting too big.

I also added a few approval tests in order to verify that after the refactoring the SQL queries and returned data was equivalent to before.

One thing to note is that the SQL snippet for the hour comparison was changed from e.g.
```
EXTRACT(HOUR FROM events.calls.datetime) >= 7
```
to
```
to_char(events.calls.datetime, 'HH24:MI') >= '07:00'
```
This is necessary to allow comparison with times that are in-between hours (e.g. 7:30). However, I believe the to-char conversion is slightly slower so we can opt to use the first one for `HourOfDay` values that correspond to whole hours and only use `to_char()` for mixed hour values. Let me know if you'd like me to role this into the current PR, otherwise we can do it in a subsequent one (perhaps after profiling whether `to_char()` really is considerably slower).


flowmachine/flowmachine/features/utilities/event_table_subset.py
```diff
@@ -208,78 +240,15 @@ def _make_query_with_sqlalchemy(self):
...
+        select_stmt = select_stmt.where(
+            self.hour_slices.get_subsetting_condition(self.sqlalchemy_table.c.datetime)
+        )
         select_stmt = self.subscriber_subsetter.apply_subset_if_needed(
             select_stmt, subscriber_identifier=self.subscriber_identifier
         )
 
         return get_sql_string(select_stmt)
 
-    def _make_query_ORIG(self):  # pragma: no cover
```
greenape: 👋

flowmachine/flowmachine/core/hour_slice.py
```diff
@@ -0,0 +1,258 @@
...
+        self.stop_hour = HourOfDay(stop_hour)
+        self.period = make_hour_interval_period(freq, weekday=weekday)
+
+    def __repr__(self):
+        return (
+            f"HourInterval(start_hour={self.start_hour!r}, stop_hour={self.stop_hour!r}, "
+            f"freq={self.period.freq!r}, weekday={self.period.weekday!r})"
+        )
+
+    def filter_timestamp_column(self, ts_col):
```
greenape: Type annotations for this, given the `__init__` is annotated?

flowmachine/flowmachine/core/hour_slice.py
```diff
@@ -0,0 +1,258 @@
...
+    valid_weekdays = [
+        "Monday",
+        "Tuesday",
+        "Wednesday",
+        "Thursday",
+        "Friday",
+        "Saturday",
+        "Sunday",
+    ]
+    weekday_indices_postgres = {
```
greenape: Could use `isodow` in the extract and then this isn't necessary?

flowmachine/flowmachine/core/hour_slice.py
```diff
@@ -0,0 +1,258 @@
...
+        "Friday": 5,
+        "Saturday": 6,
+    }
+
+    def __init__(self, weekday):
+        if weekday is None:
+            raise ValueError(
+                "If freq='week' then the `weekday` argument must be provided."
+            )
+        weekday = weekday.capitalize()
```
greenape: How about `datetime.datetime.strptime(weekday, "%A").isoweekday()`?

flowmachine/flowmachine/core/hour_slice.py
```diff
@@ -0,0 +1,258 @@
...
+    def filter_timestamp_column_by_day_of_week(self, ts_col):
+        return (
+            func.extract("dow", ts_col) == self.weekday_indices_postgres[self.weekday]
+        )
+
+
+def make_hour_interval_period(freq, *, weekday=None):
+    """
+    Returns an appropriate instance of `DayPeriod` or `DayOfWeekPeriod`,
+    depending on the value of `freq`.
```
greenape: Docstring?

flowmachine/flowmachine/core/hour_slice.py
```diff
@@ -0,0 +1,258 @@
...
+class HourOfDay:
+    """
+    Represents an hour of the day and allows filtering a timestamp column accordingly.
+    """
+
+    def __init__(self, hour_str: Union[str, None]):
+        self._validate_hour_string(hour_str)
+        self.hour_str = hour_str
+
+    def __eq__(self, other):
```
greenape: Why not just subclass `str` here?

flowmachine/flowmachine/core/hour_slice.py
```diff
@@ -0,0 +1,258 @@
...
+    try:
+        cls = cls_lookup[freq]
+    except KeyError:
+        allowed_freqs = ", ".join([repr(x) for x in cls_lookup.keys()])
+        raise ValueError(f"Argument `freq` must be one of: {allowed_freqs}.")
+
+    return cls(weekday=weekday)
+
+
+class HourOfDay:
```
greenape:
```suggestion
class HourOfDayPeriod:
```

For consistency with the others?

flowmachine/flowmachine/core/hour_slice.py
```diff
@@ -0,0 +1,258 @@
...
+        if isinstance(other, HourOfDay):
+            return self.hour_str == other.hour_str
+        elif isinstance(other, str) or other is None:
+            return self.hour_str == other
+        else:
+            raise TypeError(
+                f"HourOfDay cannot be compared to object of type {type(other)}"
+            )
+
+    def _validate_hour_string(self, hour_str):
```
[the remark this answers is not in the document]
maxalbert: Good point! I thought the year/month/day were mandatory when using `strptime`, but apparently not. :)

flowmachine/flowmachine/core/hour_slice.py
maxalbert: Yes, I'd like to add type annotations, but I'm not entirely sure what the most general valid class for the timestamp column is. Currently when I run
```
from flowmachine.core.sqlalchemy_table_definitions import EventsCallsTable 
print(type(EventsCallsTable.datetime))
```                                                                                                                              
it returns `<class 'sqlalchemy.orm.attributes.InstrumentedAttribute'>`, but it feels to me this is a fairly specific class and the method would accept more general ones. But I can add this particular one for the time being if you think that's ok?

flowmachine/flowmachine/core/hour_slice.py
maxalbert: How do you mean? I guess I can look up the index in the `valid_weekdays` array instead, but doesn't this also work for `dow`? Or are you suggesting something different?

greenape: Well, you have to do this odd little dance because Postgres-side, extracting the day of the week gives you 0-6 Sun-Sat, but Postgres also allows you you do `EXTRACT(ISODOW FROM TIMESTAMP '2001-02-18 20:38:40')`, which gives you 1-7, Monday-Sunday, same as `datetime.datetime.strptime(weekday, "%A").isoweekday()`.

flowmachine/flowmachine/core/hour_slice.py
maxalbert: I like the idea, but isn't the `%A` locale-specific? What if someone has a non-English locale set?

flowmachine/flowmachine/core/hour_slice.py
maxalbert: Ah gotcha! Yes, good point to use the iso weekday to keep indices consistent between postgres and python. 👍

flowmachine/flowmachine/core/hour_slice.py
greenape: But that's good, no? Supports non-English locales, and you store it as the iso number of the weekday?

flowmachine/flowmachine/core/hour_slice.py
maxalbert: Although sadly it turns out the `datetime` module can't just process weekdays without an explicit date. :(
```python
>>> import datetime
>>> print(datetime.datetime.strptime("Monday", "%A").isoweekday())
1
>>> print(datetime.datetime.strptime("Tuesday", "%A").isoweekday())
1
>>> print(datetime.datetime.strptime("Friday", "%A").isoweekday())
1
```

flowmachine/flowmachine/core/hour_slice.py
maxalbert: Frankly, I'm not entirely sure how this comparison would behave if we piggyback on `str` because `hour_str` can also be `None`, so I'd rather keep the explicit implementation here. But let me know if you have a strong preference for subclassing and know how to tweak the `__eq__` method to make it work.

flowmachine/flowmachine/core/hour_slice.py
maxalbert: Hmm, somehow this class doesn't seem on par to me with the other ones. `HourOfDay` represents a specific "timestamp" (or rather "hour-and-minutes-stamp"), i.e. a single point in time, whereas the other ones represent stretches / periods of time. But I'm the first one to admit that this is quite handwavy and the naming isn't very good yet (possibly also because there are perhaps even more meaningful concepts hidden here which should replace the existing ones).

Perhaps `HourOfDay` should be renamed to `HourStamp` or something similar? Any better suggestions?

flowmachine/flowmachine/core/hour_slice.py
greenape: Nuts. Well, we do also have `dateutil` as a dependency, so you could do:

```python
from dateutil.parser import parse
parse(weekday).isoweekday()
```

maxalbert: That's a nice idea! The minor downside is that `parse()` is very liberal in what it accepts, so you _can_ fool it:
```python
>>> from dateutil.parser import parse
>>> parse("Monday").isoweekday()
1
>>> parse("Monday, 2019-04-25").isoweekday()
4
```
But I can always combine it with an additional `strptime(weekday, "%A")` check...

maxalbert:
@greenape I have tweaked things according to your suggestions (thanks for those!) and updated docstrings and type annotations. Let me know if I missed anything, or if you have any further suggestions.

flowmachine/flowmachine/core/hour_slice.py
greenape: Is the None check required anywhere aside from the `filter_timestamp_column` method?

maxalbert: Nope. And now that I think of it that should perhaps actually live in a separate sister class representing a `MissingHourAndMinutesTimestamp` or similar...

Admittedly, the `__eq__` comparison is also mainly a convenience for the tests (albeit a nice one to have). So we _could_ get rid of it if you prefer.

flowmachine/flowmachine/core/hour_slice.py
maxalbert: I have renamed this to `HourAndMinutesTimestamp`, but happy to use a better name if you can think of one.

flowmachine/flowmachine/core/hour_slice.py
greenape: Well that's the nice thing about subclassing string - you stick the validation in `__new__` and you have a string that is guaranteed to have that particular formatting and the one additional method on it. If that's the _only_ place the comparison against `None` is used, you can do the compare against `'None'`.

maxalbert: Well, I'm comparing against None in the tests, so that needs to happen in the `__eq__` method. But maybe I should just get rid of that and do an explicit comparison with the full object...

maxalbert: Actually, your suggestion to subclass from string works nicely! 👍

I have introduced a new class `MissingHourAndMinutesTimestamp` which is returned from `__new__` if the input value is `None`, which separates the responsibilities better and is actually reasonably elegant and readable. :) Let me know if you're happy with this approach.

PR #673, opened 2019-04-24
Require synth data to be set

### LINKED ISSUE #654: Synthetic data generation is silently skipped if SYNTHETIC_DATA_GENERATOR env var is not set
If the `SYNTHETIC_DATA_GENERATOR` environment variable is unset, spinning up `flowdb-synthetic-data` will result in an empty `flowdb` instance without any data.

This is because the if statement [L44](https://github.com/Flowminder/FlowKit/blob/master/flowdb/testdata/bin/z_ingest_synthetic_data.sh#L44) of `flowdb/testdata/bin/z_ingest_synthetic_data.sh` does not have an explicit clause so it gets silently ignored if the env var is unset or has a value different from `"python"` or `"sql"`.

greenape:
Closes #654

### I have:

### Description

Failing to set the `SYNTHETIC_DATA_GENERATOR` env var now causes FlowDB synth data container to exit with an error.


.circleci/config.yml
```diff
@@ -217,7 +217,7 @@ jobs:
       - run:
          name: Test that not providing a superuser password causes the container to exit
          command: |
-           if docker run flowminder/flowdb:$SAFE_TAG; then
+           if docker run -e FLOWMACHINE_FLOWDB_PASSWORD=foo -e FLOWAPI_FLOWDB_PASSWORD=foo flowminder/flowdb:$SAFE_TAG; then
```
maxalbert: Hmm, does this mean that previously it "failed" (i.e., also exited with status 1) but for the wrong reasons? Can you think of any way we can ensure that this only fails in the case that we actually want to test?

PR #684, opened 2019-04-26
Misc cleanups

maxalbert:
This PR contains a few changes as a by-product of reading through some code in `query.py` and trying to understand it. It doesn't close any existing issues, but I'm happy to open some to capture the changes made in this PR.

### I have:

### Description

- A few minor changes in `flowmachine/core/query.py`:
  - In the `md5` calculation, I removed the try/except blocks. The `TypeError` in L104 never seemed to occur anyway, and the outer except block was only triggered from L98 (`item = sorted(item)`) and was due to the fact that `Query` objects can't be compared against each other, so I added an explicit `sort_key` which sorts queries by their md5.
    The calculation now also avoids duplicate inclusion of dependent queries (previously they were included both via `self.dependencies` and `self.__getstate__()`).
  - Tweaked the reprs for `Query` and `Table` (to make them more readable when displayed in a list in an interactive prompt, and to display the query_id).
  - Renamed a couple of methods and their parameters for clarity/readability.

   As a side note, it's not entirely clear to me that the logic in `Query._get_stored_dependencies()` is fully consistent with what the docstring claims about the meaning of `exclude_self`, because we're removing `self` unconditionally in the return line L805.

- Added a helper function `print_dependency_tree()` which can be used to display the dependencies of a given query. See `test_print_dependency_tree()` for sample output.

- In the `Table` class, I removed the explicit dependency on the `parent_table` because it doesn't seem to be used anywhere and makes the dependency tree look strange.

  However, this changes the logic in a couple of the tests (`test_invalidate_cache_multi` and `test_table_parent`). In particular I'm not sure whether this change has inadvertently broken something or whether the new logic in `test_invalidate_cache_multi` is as expected.

Do we need changelog entries for any of these changes?


flowmachine/flowmachine/utils.py
```diff
@@ -377,9 +373,28 @@ def sort_recursively(d):
...
-        try:
-            return sorted(d, key=to_nested_list)
-        except:
-            breakpoint()
+        return sorted(d, key=to_nested_list)
     else:
         return d
+
+
+def print_dependency_tree(query_obj, level=0, stream=sys.stdout):
```
greenape: Can we combine this with `Query.dependency_graph`? Ideally by removing `Query.dependency_graph` in favour of this ;P

flowmachine/tests/test_cache.py
```diff
@@ -182,7 +182,7 @@ def test_invalidate_cache_multi(flowmachine_connect):
     assert not cache_table_exists(flowmachine_connect, dl1.md5)
     assert not cache_table_exists(flowmachine_connect, hl1.md5)
     has_deps = bool(flowmachine_connect.fetch("SELECT * FROM cache.dependencies"))
-    assert has_deps
+    assert not has_deps
```
greenape: This test is duped I think - remove in favour of the one on L166, which correctly inverts the last assert?

flowmachine/flowmachine/core/query.py
```diff
@@ -90,21 +90,24 @@ def md5(self):
...
-                        hashes.append(str(item))
-                    except TypeError:
-                        pass
-                except:
+                if isinstance(item, Query) and item in dependencies:
+                    # this item is already included in `hashes`
+                    continue
+                elif isinstance(item, list) or isinstance(item, tuple):
+                    item = sorted(
+                        item, key=lambda x: x.md5 if isinstance(x, Query) else x
```
greenape: Nice!

flowmachine/flowmachine/utils.py
maxalbert: How do you imagine them being combined? I actually like both options - being able to print a dependency tree in plain-text format, and also to plot it as a full graph. Should I just rip the `dependency_graph()` function out of the `Query` class and put it here in `utils.py`, or do you have something else in mind?

maxalbert:
@greenape I think you said the explicit dependency of a `Table` object on its parent is needed for cache invalidation / detection of dependent queries, but also that we can probably stop making the distinction and only ever consider `Table` objects with the full subset of columns. Do you want me to put the dependency on the parent back in for this PR?

flowmachine/flowmachine/utils.py
greenape: It isn't so much the two options, but the two options with overlapping functionality being in two different places I think.

greenape:
I think put the link to parent back in.

flowmachine/flowmachine/utils.py
maxalbert: Would you prefer both of them to live as methods on the `Query` class, or rather as separate helper functions in `utils.py`?

greenape: Definitely in utils. `Query` is already way too bloated.

maxalbert:
@greenape I have moved the former `Query.dependency_tree` method into `utils.py`. Let me know if you'd like any additional changes.

PR #689, opened 2019-04-30
Harmonised test tokens

### LINKED ISSUE #564: Harmonise and centralise test token generation
Currently there are a bunch of places where test tokens are generated, which are drifting out of sync with one another _and_ with the actual available routes of the API. These need to be centralised and we need to ensure they're kept up to date as more things get added to flowapi.

greenape:
Closes #564

### I have:

### Description

This adds a new module - `flowkit-jwt-generator`, and uses it in place of the previously somewhat ad hoc test token generators.

The module contains two pytest fixtures, one of which (`access_token_builder`) generates arbitrary tokens, the other (`universal_access_token`) generates an all-access pass to a particular running flowapi instance.

Also contains two equivalent command line tools (`generate-jwt` and `all-access-jwt`), and an importable `generate_token` function, which all of these use to produce the actual tokens.


.circleci/config.yml
```diff
@@ -364,6 +364,34 @@ jobs:
...
+      JWT_SECRET_KEY: secret
+    working_directory: /home/circleci/project/flowkit_jwt_generator
+    steps:
+      - checkout:
+          path: /home/circleci/project/
+      - run: pipenv install --deploy --dev
+      - run:
+          name: Run token generator unit tests
+          command: |
+            # Run tests twice in order to pick up coverage of pytest plugin
```
maxalbert: Can you expand on why this is needed?

docs/notebook_preamble.py
```diff
@@ -48,57 +42,10 @@ def format_dict(x):
...
-    "radius_of_gyration": {
-        "permissions": {"run": True, "poll": True, "get_result": True},
-        "spatial_aggregation": ["admin3", "admin2", "admin1"],
-    },
-}
-
-TOKEN = make_token(
-    username="testuser", secret_key="secret", lifetime=timedelta(days=1), claims=claims
+
+TOKEN = generate_token(
```
maxalbert: I trust this works, but I can't see where `generate_token()` is imported in the preamble or notebook?

greenape: Looks like it isn’t actually- nice catch!

flowkit_jwt_generator/flowkit_jwt_generator/jwt.py
```diff
@@ -0,0 +1,213 @@
...
+        )
+
+    return token_maker
+
+
+def get_all_claims_from_flowapi(
+    flowapi_url: str
+) -> Dict[str, Union[Dict[str, bool], List[str]]]:
+    """
+    Retrieve all the queries types available on an instance of flowapi and
```
maxalbert:
```suggestion
    Retrieve all the query types available on an instance of flowapi and
```

flowkit_jwt_generator/flowkit_jwt_generator/jwt.py
```diff
@@ -0,0 +1,213 @@
...
+
+    """
+    query_kinds = []
+    aggregation_types = set()
+    spec = requests.get(f"{flowapi_url}/api/0/spec/openapi.json").json()
+    for q_class, q_spec in spec["components"]["schemas"].items():
+        try:
+            query_kinds.append(q_spec["properties"]["query_kind"]["enum"][0])
+            aggregation_types.update(q_spec["properties"]["aggregation_unit"]["enum"])
+        except KeyError:
```
maxalbert: There are quite a few possibilities why a `KeyError` may be thrown here (missing `"properties"`, missing `"query_kind"`, missing `"aggregation_unit"`, missing `"enum"). Are we sure this only every happens for the reason we expect, and we're not accidentally end up missing some claims? Probably fine, just checking.

flowkit_jwt_generator/flowkit_jwt_generator/jwt.py
```diff
@@ -0,0 +1,213 @@
...
+            pass
+    query_kinds = sorted(query_kinds)
+    aggregation_types = sorted(aggregation_types)
+    all_claims = {
+        query_kind: {
+            "permissions": permissions_types,
+            "spatial_aggregation": aggregation_types,
```
maxalbert: This assigns the same set of allowed aggregation types to all query kinds. Is there ever a case where we may want different ones?

flowkit_jwt_generator/flowkit_jwt_generator/jwt.py
greenape: Anything missing any of properties, query kind, or the enum key isn't a query kind, anything missing an aggregation_unit enum doesn't itself know about age types, so I think this is fine?

flowkit_jwt_generator/flowkit_jwt_generator/jwt.py
greenape: Sure, but not in an a magic all access token scenario.

flowkit_jwt_generator/tests/test_cli.py
```diff
@@ -0,0 +1,70 @@
...
+        jwt=result.output.strip(),
+        key=os.environ["JWT_SECRET_KEY"],
+        verify=True,
+        algorithms=["HS256"],
+    )
+    assert result.exit_code == 0
+    assert decoded["user_claims"] == dummy_flowapi
+
+
+def test_token_builder():
```
maxalbert: See above: add test to ensure the user can't accidentally create a token with empty permissions?

flowkit_jwt_generator/flowkit_jwt_generator/jwt.py
```diff
@@ -0,0 +1,213 @@
...
+    Parameters
+    ----------
+    username : str
+        Username for the token
+    secret : str
+        Shared secret to sign the token with
+    lifetime : datetime.timedelta
+        Lifetime from now of the token
+    claims : dict
+        Dictionary of claims the token will grant
```
maxalbert: Can we add a simple example to the docstring of what this dictionary looks like?

flowkit_jwt_generator/flowkit_jwt_generator/jwt.py
```diff
@@ -0,0 +1,213 @@
...
+import click
+import jwt
+import pytest
+import requests
+
+
+permissions_types = {"run": True, "poll": True, "get_result": True}
+
+
+def generate_token(
```
maxalbert: Should this function actually validate at least the basic structure of the claims dict?

greenape:
Merged the two commands into one, which I think gives a better token generating experience.

flowkit_jwt_generator/flowkit_jwt_generator/jwt.py
```diff
@@ -0,0 +1,229 @@
...
+)
+@click.option(
+    "--aggregation",
+    "-a",
+    type=click.Choice(aggregation_types),
+    multiple=True,
+    help="Spatial aggregation level of access allowed.",
+)
+def named_query(query_name, permission, aggregation):
+    if len(permission) == 0 and len(aggregation) == 0:
```
maxalbert: Should this be an `or` rather than an `and`? Otherwise it's possible to generate tokens which have no run/poll/get_result permissions at all (only aggregation units), and vice versa.

maxalbert:
Awesome, thanks for the changes. Looking good now! 👍 

Not sure if I mentioned this before, but one last suggestion is to add a `__init__.py` file in the new `flowkit_jwt_generator` module (and import it a couple of the relevant functions from `jwt.py`).

Also, I'm not sure whether we should add basic verification of the structure of the claims dict, but perhaps that's best left for a separate PR (if you even want to include it).

flowkit_jwt_generator/flowkit_jwt_generator/jwt.py
greenape: I was pondering that. My conclusion was that you want to be able to generate broken tokens for test purposes.

maxalbert: Hmm, that's a good point. I feel that because our toolchains are fairly complicated already we should prioritise a good user experience and make "broken usage" for testing purposes possible but requiring an extra hoop to jump through. A useful way to do that could be to have a flag such as `allow_broken_claims=True` (probably with a better name for the argument). But again, I'm happy to leave that for a separate PR if you prefer.

PR #693, opened 2019-05-01
One line demo startup

### LINKED ISSUE #688: Quick install instructions are insufficient
As of #610, `docker-compose.yml` requires `development_environment` to be sourced (or the environment variables otherwise set) before running `docker-compose up`. The [quick install instructions](https://flowminder.github.io/FlowKit/install/#quick-install) have not been updated to reflect this.

Rather than instructing a user to download and source `development_environment` before docker-composing, it would seem preferable to provide either a bash script which would set the env vars, download the compose file and spin everything up, or a demo `docker-compose` file in which the env vars are already set (this links in with #614).

While we're on the subject, we should also review the [dev](https://flowminder.github.io/FlowKit/install/#developer-install) and [production](https://flowminder.github.io/FlowKit/install/#production-install) install instructions, to ensure they are correct and sufficient.

greenape:
Closes #688

### I have:

### Description

This introduces a bash script that starts a complete (demo mode) FlowKit system, and can be run (ala homebrew) by simply curl-ing the script direct from GitHub.

I've split the docker-compose file to facilitate this, and moved the build blocks to a separate one, so you can run as before by just using both files.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -682,7 +682,6 @@ workflows:
           component: flowmachine
           requires:
             - run_flowmachine_tests
-            - integration_tests
           <<: *run_always_org_context
       - build_docs:
           name: build_docs
@@ -702,7 +701,6 @@ workflows:
           component: flowapi
           requires:
             - run_flowkit_api_tests
-            - integration_tests
           <<: *run_always_org_context
       - build_python_wheel:
           name: build_flowclient_wheel
@@ -733,19 +731,13 @@ workflows:
       - retag_images:
           name: retag_master_branch
           requires:
-            - build_flowmachine
-            - build_flowdb
-            - build_flowapi
-            - build_flowauth
+            - test_quickstart
             - build_docs
           <<: *master_only_org_context
       - retag_images:
           name: retag_tagged_build
           requires:
-            - build_flowmachine
-            - build_flowdb
-            - build_flowapi
-            - build_flowauth
+            - test_quickstart
             - build_docs
           tag: CIRCLE_TAG
           <<: *tag_only_org_context
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -506,6 +506,17 @@ jobs:
           path: test_results
       - run: bash <(curl -s https://codecov.io/bash)
 
+  test_quickstart:
+    machine:
+      image: circleci/classic:201808-01
+    working_directory: /home/circleci/project
+    environment:
+      BRANCH: $CIRCLE_BRANCH
+    steps:
+      - checkout:
+          path: /home/circleci/project/
+      - run: ./quick_start.sh
+
   build_docs:
     parameters:
       deploy:
@@ -706,6 +717,13 @@ workflows:
           requires:
             - run_flowmachine_tests
             - integration_tests
+          <<: *run_always_org_context
+      - test_quickstart:
+          requires:
+            - build_flowdb
+            - build_flowapi
+            - build_flowmachine
+            - build_flowauth
           <<: *run_always_org_context
       - integration_tests:
           requires:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -516,6 +516,7 @@ jobs:
       - checkout:
           path: /home/circleci/project/
       - run: ./quick_start.sh
+      - run: ./quick_start.sh stop
 
   build_docs:
     parameters:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,10 +6,15 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## [Unreleased]
 
 ### Added
+- A new Ansible playbook was added in `deployment/provision-dev.yml`. In addition to the standard provisioning
+  this installs pyenv, Python 3.7, pipenv and clones the FlowKit repository, which is useful for development purposes.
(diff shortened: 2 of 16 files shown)


CHANGELOG.md
```diff
@@ -8,11 +8,13 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Added
 - A new Ansible playbook was added in `deployment/provision-dev.yml`. In addition to the standard provisioning
   this installs pyenv, Python 3.7, pipenv and clones the FlowKit repository, which is useful for development purposes.
+- Added a 'quick start' setup script for trying out a complete FlowKit system [#688](https://github.com/Flowminder/FlowKit/issues/688).
 
 ### Changed
 - FlowAPI's `available_dates` endpoint now always returns available dates for all event types and does not accept JSON
 - Hints are now displayed in the add user form of FlowAuth if the form is not completed [#679](https://github.com/Flowminder/FlowKit/issues/679)
 - The Ansible playbooks in `deployment/` now allow configuring the username and password for the FlowKit user account.
+- Default compose file no longer includes build blocks, these have been moved to
```
maxalbert: Missing end of sentence (where have they been moved to?)

docker-compose.yml
```diff
@@ -148,17 +136,14 @@ services:
...
-      dockerfile: Dockerfile
+    image: flowminder/flowauth:${CONTAINER_TAG:-latest}
     ports:
       - ${FLOWAUTH_PORT:?Must set FLOWAUTH_PORT env var}:80
     environment:
       DEMO_MODE: ${DEMO_MODE:?Must set DEMO_MODE env var}
 
   query_locker:
-    container_name: redis_flowkit
+    container_name: query_locker
```
maxalbert: Not for this PR, but perhaps we can rename `query_locker` to `flowmachine_query_locker` (or `flowkit_query_locker`)?

docs/source/install.md
```diff
@@ -23,67 +23,21 @@ These instructions assume use of [Pyenv](https://github.com/pyenv/pyenv) and [Pi
...
 Start the FlowKit test system by running 
+
 ```bash
-JWT_SECRET_KEY=secret docker-compose up -d
+bash <(curl -s https://raw.githubusercontent.com/Flowminder/FlowKit/master/quick_start.sh)
 ``` 
-This will pull any necessary docker containers, and start the system in the background with the API exposed on port `9090` by default.  
 
-In order to use the test system, now install FlowClient and FlowAuth. 
+This will pull any necessary docker containers, and start the system in the background with the API exposed on port `9090` by default, and the FlowAuth authentication system accessible by visiting http://localhost:8000 using your web browser.
```
maxalbert: Just as an observation, but I wonder whether we should use a different standard port for flowauth because 8000 is not unlikely to be used by something else already.

On that note, what happens if any of these ports is already taken? Will it fail, or put the services on some other random available port?

greenape: It'd fail I think.
Would be possible to leave the exposed ports blank in the compose, which would then find them random ones. Do then have to do some docker inspect + jq fiddling to work out what the ports _are_.

docs/source/install.md
```diff
@@ -23,67 +23,21 @@ These instructions assume use of [Pyenv](https://github.com/pyenv/pyenv) and [Pi
...
-    <p>
-7. Or:
-    <p>
-    - Under "Groups", add a new group,
-
-    - Add a server to the group, and enable/disable API permissions and aggregation units,
-
-    - Add the user to the group.
-
+bash <(curl -s https://raw.githubusercontent.com/Flowminder/FlowKit/master/quick_start.sh) synth
```
maxalbert: Not a biggie, but I tend to like slightly more expressive names (`synth` only makes sense in the context of these instructions and is otherwise a bit hard to guess/remember for the user unless you know what it refers to). But also happy to keep it for now until we have used this quick-start script a few more times and have gathered some feedback from users.

greenape: Happy to take suggestions there :)

docs/source/install.md
```diff
@@ -23,67 +23,21 @@ These instructions assume use of [Pyenv](https://github.com/pyenv/pyenv) and [Pi
...
-#### FlowAuth Quickstart
-
-To run a demonstration version of FlowAuth use:
+The default system includes a small amount of test data. For a test system with considerably more data you can run
 
 ```bash
-docker run -p 8000:80 -e DEMO_MODE=1 flowminder/flowauth
-```
-
-This will start FlowAuth at <a href="http://localhost:8000/" target="_blank">http://localhost:8000</a>, and pre-populate a disposable sqlite database with some dummy data. Log in with either `TEST_ADMIN:DUMMY_PASSWORD` or `TEST_USER:DUMMY_PASSWORD`.
```
maxalbert: What's the reason for the FlowAuth instructions to be removed? Are they not needed because the quick start demo has an appropriate token built in?

greenape: It starts up FlowAuth - I've been over aggressive with what I removed though, you still need a username and password!

.circleci/config.yml
```diff
@@ -671,7 +683,6 @@ workflows:
           component: flowmachine
           requires:
             - run_flowmachine_tests
-            - integration_tests
```
maxalbert: Just to check: am I correct to assume that these can be removed because they are implicit in the re-defined workflow?

greenape: Yeah, I think the critical bit really is that unless _everything_ works, the build will never get tagged as latest, nor receive a release tag. Doesn't matter so much if the commit-hash tagged image gets pushed up to docker cloud in my view.

PR #694, opened 2019-05-02
Dockerized worked examples

### LINKED ISSUE #614: Production docker-compose file
As of #610 the `docker-compose.yml` in the project root is well-suited for all of the dev use-cases. However, I think it would be useful to have a separate docker-compose file designed for a production setup which doesn't depend on the git repository.

The differences between this and the top-level docker-compose file would be:

- Only the essential ports exposed
- No `build` sections (since this would be intended to be used without the git repo)
- An additional env var (e.g. `RELEASE_TAG`, default `latest`) to set the tag to pull from docker cloud (this would be useful in a situation where somebody has done `pip install flowclient` and therefore has a release version of flowclient, which may not be compatible with the `latest` flowapi image)
- Default values for all env vars except things like passwords, but allow the user to override those env vars where it is sensible to do so

This docker-compose file could be tucked away in the docs somewhere, to avoid confusion with the dev setup, but prominently linked-to in the docs (making it clear that the docker-compose file is all that's required to spin up a full FlowKit system, and the git repo is not needed).

The docs should also include instructions for how the user can know when the system is up and ready (since this can take a while, e.g. when using the synthetic data image) and how to stop all the containers (`docker-compose down -v`).

I'm not sure what's the best way to allow users to choose which flowdb image to use (they may want to try running the worked examples, in which case the synthetic data image would be best, or for a production setup they would probably want the "empty" flowdb image). Perhaps we want a "demo" docker-compose file and a "production" one? Or maybe just choose the one we think will be most useful.

### LINKED ISSUE #688: Quick install instructions are insufficient

greenape:
Closes #614 closes #688

### I have:

### Description

Overall intent: make the worked examples _super_ easy to run.

- Adds a new top level Dockerfile based on the official jupyter images, with the worked examples from the docs built in, and flowmachine and FlowClient installed.
- Adds worked examples to the quick start bash script
- Squashes the CI build process down a bit so that all the docker images can be built earlier because there's no particular reason to wait until the corresponding tests pass to try and build them

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -108,6 +108,16 @@ flowauth-build:
 	docker-compose -f $(DOCKER_COMPOSE_FILE) -f $(DOCKER_COMPOSE_FILE_BUILD) build flowauth
 
 
+worked-examples-up: worked-examples-build
+	docker-compose -f $(DOCKER_COMPOSE_FILE) -f $(DOCKER_COMPOSE_FILE_BUILD) up -d --build worked_examples
+
+worked-examples-down:
+	docker-compose -f $(DOCKER_COMPOSE_FILE) rm -f -s -v worked_examples
+
+worked-examples-build:
+	docker-compose -f $(DOCKER_COMPOSE_FILE) -f $(DOCKER_COMPOSE_FILE_BUILD) build worked_examples
+
+
 flowmachine_query_locker-up:
 	docker-compose -f $(DOCKER_COMPOSE_FILE) up -d flowmachine_query_locker
 
--- a/Makefile
+++ b/Makefile
@@ -108,13 +108,13 @@ flowauth-build:
 	docker-compose -f $(DOCKER_COMPOSE_FILE) -f $(DOCKER_COMPOSE_FILE_BUILD) build flowauth
 
 
-worked-examples-up: worked-examples-build
+worked_examples-up: worked_examples-build
 	docker-compose -f $(DOCKER_COMPOSE_FILE) -f $(DOCKER_COMPOSE_FILE_BUILD) up -d --build worked_examples
 
-worked-examples-down:
+worked_examples-down:
 	docker-compose -f $(DOCKER_COMPOSE_FILE) rm -f -s -v worked_examples
 
-worked-examples-build:
+worked_examples-build:
 	docker-compose -f $(DOCKER_COMPOSE_FILE) -f $(DOCKER_COMPOSE_FILE_BUILD) build worked_examples
 
 
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@
 DOCKER_COMPOSE_FILE ?= docker-compose.yml
 DOCKER_COMPOSE_FILE_BUILD ?= docker-compose-build.yml
 FLOWDB_SERVICES ?= flowdb_testdata
-DOCKER_SERVICES ?= $(FLOWDB_SERVICES) flowapi flowmachine flowauth flowmachine_query_locker
+DOCKER_SERVICES ?= $(FLOWDB_SERVICES) flowapi flowmachine flowauth flowmachine_query_locker worked_examples
 export DOCKER_FLOWDB_HOST=$(word 1, $(FLOWDB_SERVICES))
 
 
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -115,10 +115,7 @@ services:
 
   worked_examples:
     container_name: worked_examples
-    image: flowminder/flowkit-examples:latest
-    build:
-      context: .
-      dockerfile: Dockerfile
+    image: flowminder/flowkit-examples:${CONTAINER_TAG:-latest}
     ports:
       - 8888
     tty: true
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -132,6 +132,8 @@ services:
     networks:
       - db
       - redis
+    command:
+      ["start-notebook.sh", "c.NotebookApp.token = 'b2a960c88d617bde49e592da8b60d09b3527873cb5a50db2'"]
 
   flowapi:
     container_name: flowapi
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -133,7 +133,7 @@ services:
       - db
       - redis
     command:
-      ["start-notebook.sh", "c.NotebookApp.token = 'b2a960c88d617bde49e592da8b60d09b3527873cb5a50db2'"]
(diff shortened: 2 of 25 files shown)


greenape:
To try this out prior to merging to master, you can run:

```bash
env CI=true CIRCLE_SHA1=6b91a8c2ab7079432eee0f1288f13baac342e287 bash <(curl -s https://raw.githubusercontent.com/Flowminder/FlowKit/6b91a8c2ab7079432eee0f1288f13baac342e287/quick_start.sh) examples
```

jch1g10:
A few issues I've come across so far:
- The FlowAPI URL set in `flowclient.connect()` in the example notebooks is "http://localhost:9090", which is correct for the docs build, but inside the container this needs to be "http://flowapi:9090".
- The API token available to TEST_USER in the demo FlowAuth setup only allows access to `admin0` and `admin1` aggregations, but the queries in the worked examples aggregate to `admin3` level, so a user would have to log into FlowAuth as TEST_ADMIN and modify TEST_USER's permissions before being able to create a token which allows them to run the examples.
- By default, `bash <(curl ...) examples` starts the system with `flowdb_testdata`, but the worked examples use date ranges not included in the test data so attempting to run them will give `"No data for date"` errors.

Makefile
```diff
@@ -108,6 +108,16 @@ flowauth-build:
 	docker-compose -f $(DOCKER_COMPOSE_FILE) -f $(DOCKER_COMPOSE_FILE_BUILD) build flowauth
 
 
+worked-examples-up: worked-examples-build
```
jch1g10: `worked_examples-up` instead of `worked-examples-up` would be more consistent with the names of the other targets.

docs/source/install.md
```diff
@@ -34,9 +31,26 @@ This will pull any necessary docker containers, and start the system in the back
...
+The [worked examples](worked_examples) are also available as part of the demo system. To install these run 
+
+```bash
+bash <(curl -s https://raw.githubusercontent.com/Flowminder/FlowKit/master/quick_start.sh) examples
+```  
+
+for the examples with a small dataset, or 
+
+```bash
+bash <(curl -s https://raw.githubusercontent.com/Flowminder/FlowKit/master/quick_start.sh) examples larger_data
```
jch1g10: Would be useful to also give instructions here for stopping the system cleanly (presumably `bash ... stop` is sufficient).

greenape:
Ok, let’s -

- make the url an env var
- make the default token more permissive
- switch the default flowdb to synth for the examples and allow making it smaller

I had it as synth originally, but it takes soooo long to spin up on ci that I swapped it to the testdata.

docker-compose.yml
```diff
@@ -5,7 +5,7 @@
 # DOCKER COMPOSE FOR FLOWKIT
 #
 
-version: '3.7'
+version: '3.5'
```
maxalbert: Why do we need to revert this to an earlier version?

greenape: We don't need any features of 3.7 format, only 3.5 which introduced `shm_size`. Let's us support a marginally broader range of docker releases.

quick_start.sh
```diff
@@ -65,11 +96,12 @@ then
 else
     source /dev/stdin <<< "$(curl -s https://raw.githubusercontent.com/Flowminder/FlowKit/${BRANCH:-master}/development_environment)"
     echo "Starting containers"
```
maxalbert: When I tried spinning up the quick-start examples, for a while the script just seemed to be hanging. Somehow docker-compose doesn't give any visual indication that it's downloading the images. Can we add a comment to this effect here? E.g. add: "this may take a while" (and perhaps even indicate that it may look like nothing is happening for a while)?

docker-compose.yml
```diff
@@ -113,6 +113,29 @@ services:
...
+    stdin_open: true
+    environment:
+      - FLOWDB_PORT=5432
+      - FLOWDB_HOST=${DOCKER_FLOWDB_HOST:?Must set DOCKER_FLOWDB_HOST env var}
+      - FLOWMACHINE_FLOWDB_USER=${FLOWMACHINE_FLOWDB_USER:?Must set FLOWMACHINE_FLOWDB_USER env var}
+      - FLOWMACHINE_FLOWDB_PASSWORD=${FLOWMACHINE_FLOWDB_PASSWORD:?Must set FLOWMACHINE_FLOWDB_PASSWORD env var}
+      - REDIS_HOST=query_locker
+      - REDIS_PORT=6379
+      - REDIS_PASSWORD=${REDIS_PASSWORD:?Must set REDIS_PASSWORD env var}
+      - FLOWAPI_URL="http://flowapi:9090"
```
maxalbert: This actually needs to be without the quotation marks:
```suggestion
      - FLOWAPI_URL=http://flowapi:9090
```
The reason is that otherwise the example notebooks will pick it up with an extra set of quotation marks:
```
>>> os.getenv("FLOWAPI_URL)
'"http://flowapi:9090"'
```
... and raise an error when talking to flowapi:
```
InvalidSchema: No connection adapters were found for '"http://flowapi:9090"/api/0/run'
```

quick_start.sh
greenape: That's weird.. I definitely get output from compose during the pull and the up.

maxalbert: Yes, I do get output, but it is static and doesn't change for a while until the pull is complete.

It basically looks like this for a couple of minutes:
```
Starting containers
Pulling flowdb                   ...
Pulling flowdb_testdata          ...
Pulling flowdb_synthetic_data    ...
Pulling worked_examples          ...
Pulling flowapi                  ...
Pulling flowauth                 ...
Pulling flowmachine_query_locker ...
Pulling flowmachine              ...
```
But since there is no visual interaction and nothing changes, it's easy to think that the script is hanging or something went wrong (until the next bit of output appears after a while).

How about changing the message to the following?
```suggestion
     echo "Starting containers (this may take a few minutes)"
```

PR #697, opened 2019-05-02
Refactor logger initialisation

### LINKED ISSUE #691: `flowmachine.connect()` creates multiple loggers if connection to flowdb failed
**Describe the bug**
When calling `flowmachine.connect()` multiple times without flowdb being available (so that the call to `connnect()` fails), this creates multiple duplicate loggers (and might also have other duplicate effects).

**Product**
Flowmachine.

**Version**
Latest master (`0.6.1+107.gfd229bf8`).

**To Reproduce**
Ensure that flowdb is *not* running.

Then run:
```python
import flowmachine
import sqlalchemy
from time import sleep

for i in range(1, 6): 
    try: 
        flowmachine.connect()
        break
    except sqlalchemy.exc.OperationalError: 
        print(f"Failed to connect to flowdb. Retrying (attempt {i}/5)") 
        sleep(1)
```
The output shows that multiple duplicate loggers are created:
```
Logger created with level DEBUG
Failed to connect to flowdb. Retrying (attempt 1/5)
Logger created with level DEBUG
Logger created with level DEBUG
Failed to connect to flowdb. Retrying (attempt 2/5)
Logger created with level DEBUG
Logger created with level DEBUG
Logger created with level DEBUG
Failed to connect to flowdb. Retrying (attempt 3/5)
Logger created with level DEBUG
Logger created with level DEBUG
Logger created with level DEBUG
Logger created with level DEBUG
Failed to connect to flowdb. Retrying (attempt 4/5)
Logger created with level DEBUG
Logger created with level DEBUG
Logger created with level DEBUG
Logger created with level DEBUG
Logger created with level DEBUG
Failed to connect to flowdb. Retrying (attempt 5/5)
```

**Expected behavior**
Only a single logger should be created. More generally, `flowmachine.connect()` should not do anything other than trying to establish the connection to `flowdb`. Any other initialisation code should be run separately.

maxalbert:
Closes #691

### I have:

### Description

This PR moves the initialisation of the various loggers into a single place (in `flowmachine/core/logging.py`). The two main functions in there are `init_logging()` (to create and set up the various loggers when flowmachine is imported) and `set_log_level()` (to set the log level of a particular logger at a later time).

`init_logging()` is now called once when flowmachine is imported. This previously happened in `flowmachine.connect()`, but now we only adjust the log level there.

As a tangential refactoring I also moved the `__version__` and `__min_flowdb_version__` into a separate file. This was necessary because the original import inside `Connection.__check_flowdb_version()` was `from ..__init__ import ...`, and this would trigger a re-execution of the `init_logging()` call.


flowmachine/flowmachine/core/logging.py
```diff
@@ -0,0 +1,87 @@
...
+            structlog.processors.JSONRenderer(serializer=rapidjson.dumps),
+        ],
+        context_class=dict,
+        logger_factory=structlog.stdlib.LoggerFactory(),
+        wrapper_class=structlog.stdlib.BoundLogger,
+        cache_logger_on_first_use=True,
+    )
+
+
+def set_log_level(logger_name, log_level):
```
greenape: Types?

flowmachine/flowmachine/versions.py
```diff
@@ -0,0 +1,10 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from ._version import get_versions
+
+__version__ = get_versions()["version"]
+del get_versions
+
+__min_flowdb_version__ = "0.2.0"
```
greenape: Maybe worth updating this to whichever first version with the dfs tables is?

maxalbert:
Oh, one quick note: when moving the initialisation of `query_run_log` into `init_logging()` I removed the following line because it didn't seem to do anything and the tests still pass. Let me know if it actually needs to be there, then I'll put it back in.
```python
query_run_log = structlog.wrap_logger(query_run_log)
```

greenape:
I think it should be fine, because the get_logger call in server.py going via structlog will wrap it anyhow.

PR #732, opened 2019-05-09
break the test

josh-gree:
The test for the presence of sample DAG - succeeded by chance because airflow comes with example dags the last of which happened to also be called `tutorial` - going to break the test in first commit and then fix...


josh-gree:
Should now pass...

josh-gree:
@maxalbert looks good!

PR #733, opened 2019-05-09
Audience key in JWTs

### LINKED ISSUE #727: Audience key in tokens
A prerequisite of #89

- [ ] The token generator should _optionally_ support the audience key of tokens.
- [ ] FlowAuth should _mandatorily_ set the audience key (e.g. server name)
- [ ] FlowAPI should _mandatorily_ require an audience key (e.g. required env var)

greenape:
Closes #727

### I have:

### Description

This adds an audience key to the JWTs (required to use public key signing for the link between flowauth and flowapi while not allowing tokens to be used for different servers than intended).

The audience key is the _name_ of the server, so when adding a server in FlowAuth, this will now need to match the `FLOWAPI_IDENTIFIER` env var passed into the flowapi container.


flowapi/flowapi/config.py
```diff
@@ -43,11 +43,13 @@ def get_config():
     )
     flowdb_host = os.environ["FLOWDB_HOST"]
     flowdb_port = os.environ["FLOWDB_PORT"]
+    server_id = getsecret("FLOWAPI_IDENTIFIER", os.environ["FLOWAPI_IDENTIFIER"])
```
maxalbert: Nitpick, but can we call the variable `flowapi_identifier` or `flowapi_server_id`, so that it's clear which server this is referring to? Mainly to make the line below (which sets `JWT_DECODE_AUDIENCE`) more understandable at a glance.

flowauth/backend/flowauth/token_management.py
```diff
@@ -226,3 +223,55 @@ def add_token(server):
     db.session.add(token)
     db.session.commit()
     return jsonify({"token": token_string, "id": token.id})
+
+
+def generate_token(
```
maxalbert: I feel like you mentioned this, but why not use the `generate_token` function that's already present in the new module `flowkit_jwt_generator`?

flowkit_jwt_generator/flowkit_jwt_generator/__init__.py
```diff
@@ -7,11 +7,17 @@
     get_all_claims_from_flowapi,
     aggregation_types,
     permissions_types,
+    generate_keypair,
+    load_private_key,
+    load_public_key,
 )
 
 __all__ = [
```
maxalbert: Just an observation, but I don't think we _need_ the `__all__` variable in `__init__.py` because we want all the imported names to be available anyway. So I suppose we could delete it to avoid the duplication (I get the sense this also applies to a bunch of modules in flowmachine.) But of course also fine to keep it.

flowkit_jwt_generator/flowkit_jwt_generator/jwt.py
```diff
@@ -1,24 +1,109 @@
...
+        The private key
+    """
+    try:
+        return serialization.load_pem_private_key(
+            key_string.encode(), password=None, backend=default_backend()
+        )
+    except ValueError:
+        try:
+            return load_private_key(base64.b64decode(key_string).decode())
+        except (Error, ValueError):
```
maxalbert: Again nitpick, but reading this at first I thought we're catching generic exceptions until I realised that `Error` was imported from `binascii`. How about making this explicit here so that it's clear _why_ the command might fail? (Requires tweaking the import, of course.)
```suggestion
        except (binascii.Error, ValueError):
```

flowkit_jwt_generator/flowkit_jwt_generator/jwt.py
```diff
@@ -1,24 +1,109 @@
...
+        The public key
+    """
+    try:
+        return serialization.load_pem_public_key(
+            key_string.encode(), backend=default_backend()
+        )
+    except ValueError:
+        try:
+            return load_public_key(base64.b64decode(key_string).decode())
+        except (Error, ValueError):
```
maxalbert: See above, in case you want to change it:
```suggestion
        except (binascii.Error, ValueError):
```

flowauth/backend/flowauth/token_management.py
greenape: Docker build context scope.

maxalbert: So we _could_ use it if we publish `flowkit_jwt_generator` on PyPI, for example?

flowkit_jwt_generator/flowkit_jwt_generator/jwt.py
```diff
@@ -36,12 +121,14 @@ def generate_token(
         Lifetime from now of the token
     claims : dict
         Dictionary of claims the token will grant
+    audience : str, optional
+        Optionally provide a string to identify the audience of the token
```
maxalbert: Hmm, upon reflection I wonder whether within the FlowKit context we should give this argument a more explicit name than "audience". I know this is the JWT terminology, but for a FlowKit user it is not obvious what "audience" refers to. How about naming it `flowapi_server_name` or `flowapi_target_server` or something similar? This would help clarify what should be passed here, and make the usage more intuitive. At the very least we should tweak the docstring to explain what "audience" means.

flowauth/backend/flowauth/token_management.py
greenape: You'd need to publish to Pypi part way through the CI process, and would only be viable for tagged builds - in any other case, you're running the risk that flowkit_jwt_generator has changed in that branch.

flowkit_jwt_generator/flowkit_jwt_generator/jwt.py
```diff
@@ -172,23 +271,24 @@ def print_token(username, secret_key, lifetime):
 
     For example:
 
-    generate-jwt TEST_USER SECRET 1 --all-access http://localhost:9090
+    generate-jwt TEST_USER SECRET 1 TEST_SERVER --all-access http://localhost:9090
```
maxalbert: Something for a future PR; just noticing that this command would be a little more readable if the first few arguments were passed using (mandatory) named arguments, too:
```
generate-jwt --all-access \
    --username TEST_USER \
    --secret-key SECRET \
    --lifetime 1 \
    --flowapi-server-name TEST_SERVER \
    http://localhost:9090
```

flowauth/backend/flowauth/token_management.py
maxalbert: Good point. I suppose there are workarounds (e.g. copying the `flowkit_jwt_generator/` folder into `flowauth/` before building, and aborting the build if it is empty) but certainly not worth the hassle for a small function like this.

flowkit_jwt_generator/tests/test_keypairs.py
```diff
@@ -0,0 +1,66 @@
...
+from flowkit_jwt_generator.jwt import load_private_key, load_public_key
+
+
+def test_keypair_generator():
+    """
+    Keypair generator produces a valid key pair, and we can load it from a string.
+    """
+    private_key, public_key = generate_keypair()
+    rsa_private_key = load_private_key(private_key.decode())
+    assert private_key == rsa_private_key.private_bytes(
```
maxalbert: Isn't this simply duplicating the implementation? Still good to have the test, I just wonder whether there is an independent way of testing it (e.g. to encrypt & decrypt a small sample message and test that we get the original result?) But if not this is probably the easiest way to test things.

flowauth/backend/flowauth/token_management.py
maxalbert: Given that we _are_ duplicating the `generate_token()` function, can we add a comment here and also in `flowkit_jwt_generator` that a duplicate version exists and that they should be changed in parallel?

greenape: I've added a test that covers the duped generator, but fine to add a comment as well.

maxalbert: Yes, great to have the test. 👍Just figured it's useful to have a comment too so that you don't have to wait for a failing CI run to discover you should have changed the other implementation, too.

flowauth/backend/flowauth/token_management.py
```diff
@@ -204,17 +206,12 @@ def add_token(server):
...
         secret=server.secret_key,
-        algorithm="HS256",
-        expires_delta=lifetime,
-        fresh=True,
-        user_claims=json["claims"],
-        csrf=False,
-        identity_claim_key="identity",
-        user_claims_key="user_claims",
-        json_encoder=JSONEncoder,
+        audience=server.name,
```
maxalbert: Forgot to change this to `flowapi_identifier`?

flowauth/backend/flowauth/token_management.py
```diff
@@ -226,3 +223,56 @@ def add_token(server):
     db.session.add(token)
     db.session.commit()
     return jsonify({"token": token_string, "id": token.id})
+
+
+# Duplicated in flowkit_jwt_generator
```
maxalbert:
```
# Duplicated in flowkit_jwt_generator (cannot re-use the implementation
# there because the module is outside the docker build context for flowauth).
```

flowkit_jwt_generator/flowkit_jwt_generator/jwt.py
```diff
@@ -1,24 +1,110 @@
...
+    )
+    private_key_bytes = private_key.private_bytes(
+        encoding=serialization.Encoding.PEM,
+        format=serialization.PrivateFormat.TraditionalOpenSSL,
+        encryption_algorithm=serialization.NoEncryption(),
+    )
+    return private_key_bytes, public_key_bytes
+
+
+# Duplicated in FlowAuth
```
maxalbert:
```suggestion
# Duplicated in FlowAuth (cannot use this implementation there because
# this module is outside the docker build context for FlowAuth).
```

PR #741, opened 2019-05-09
Remove inheritance of subscriber subsetters from Query

### LINKED ISSUE #740: Subscriber subsetters should not inherit from Query
The `SubscriberSubsetterBase` class currently inherits from Query. This was a workaround introduced in #375 but it comes with a few drawbacks, caused by the fact that the subsetter classes do not represent full SQL queries but rather SQL _snippets_ which can be slotted into the full query. This inheritance should be removed.

### LINKED ISSUE #797: Tests which compare against expected query_ids are brittle
There are a few flowmachine and integration tests which construct query objects and compare against expected query_ids which are currently hard-coded in the tests. These query_ids are subject to change when internals are refactored, which makes the tests brittle and fixing them somewhat tedious.

In the integration tests we could alternatively construct the query objects directly using flowmachine and compare the resulting query_ids against the ones returned from the API, which would avoid hard-coding.

In the flowmachine tests we may need to keep the explicit query_ids hard-coded, but it would be nice to turn the test into using `approvaltests`, so that we don't need to update a bunch of values manually when they change.

maxalbert:
Closes #740.
Closes #797.

### I have:

### Description

This PR removes the inheritance of `SubscriberSubsetterBase` from `Query`.

I have added a `__repr__` method  to `SubscriberSubsetterBase` which returns a unique string representation for a given subsetter to ensure that subscriber subsets contribute to the md5 hash of the full query in a unique way (this is used in `Query.md5` to construct the query_id of the full query object).

Removing the inheritance from Query allows us to get rid of a few awkward helper methods and special cases.


greenape:
I’m wondering if they actually _need_ to be involved in the hash.. the subset thing is getting stored as an attribute of the event table subset anyway, so arguably you don’t need any further info to construct the hash for that (and obvs we want to continue to store the subset on event table subset so dependency tracking works correctly).

Given that, couldn’t one drop the md5 from these and only instantiate in _make_query to similar ends?

maxalbert:
Hmm, unless I'm misunderstanding your question I think the changes in this PR already do what you suggest.

With this refactoring the subsetters are not _directly_ involved in the hash any more (because they don't show up in `Query.dependencies` any more). They only contribute indirectly because they are attributes on `EventTableSubset`, as you point out. (By "indirectly" I mean that they pass through the `else` branch in [L105](https://github.com/Flowminder/FlowKit/blob/master/flowmachine/flowmachine/core/query.py#L105) of `query.py` and then get converted to a string in [L109](https://github.com/Flowminder/FlowKit/blob/master/flowmachine/flowmachine/core/query.py#L109).

So in order to give a unique contribution to the query id this string representation of a subsetter needs to be unique. The easiest way I could think of was to include some sort of md5 hash in the subsetter string representation itself. Of course there are other options, but this seems fairly simple and clean to me.

Does this make sense, or did I misunderstand your question?

greenape:
@maxalbert Sorry, that wasn't very clear!

What I was wondering is: the 'uniqueness' part of the subsetters is covered by the inclusion of the _input_ to the subsetters in the `EventTableSubset`'s attributes (https://github.com/Flowminder/FlowKit/blob/64fcdfc14799ec383dc81c5cc4cfd3be02682c2e/flowmachine/flowmachine/features/utilities/event_table_subset.py#L127).

So having them in the hash (https://github.com/Flowminder/FlowKit/blob/64fcdfc14799ec383dc81c5cc4cfd3be02682c2e/flowmachine/flowmachine/features/utilities/event_table_subset.py#L128) doesn't, per se, add anything - you'd get a _different_ hash, obviously, but you don't need to include them to disambiguate two queries with otherwise identical parameters.

On that basis, one would skip the unique repr and simply not have them as an attribute, instantiating them only when the snippet needs to get built in `_make_query`.

(When I wrote the original, I wasn't thinking about their additional role, which is to validate the subset, and perhaps means this is the better path).

maxalbert:
@greenape Thanks for the clarification, I see what you mean now. And you are correct, at the moment there is duplicate information there due to the fact that we're still storing the original input in `self.subscriber_subset_ORIG`.

My assumption was that we'll want to move away from explicitly passing these inputs around and instead only pass the subsetter instances around (similar to how we're now passing query objects when constructing, say, modal locations).

Maybe this assumption is wrong, but if that's the path we want to take then we will need this information to be contained in the new subsetter classes themselves rather than storing it somewhere externally. Which to me also makes sense from an information encapsulation point of view because that's where the information logically belongs. How do you think about this?

greenape:
I'd tend to agree, but as noted, where the subscriber subset is another table/query you want to be able to track that dependency.

maxalbert:
That is a good point. I'd hate for this to get in the way of a cleaner design though, so I'd be tempted to track this dependency as part of the subsetter rather than as part of the original query.

My understanding is that currently this is not possible because we can only track proper `Query` objects as dependencies (and the whole point of this PR is to _not_ have the subsetters inherit from `Query`). Would you be ok with de-coupling the dependency tracking mechanism from `Query`? For example, we could have a mixin which marks a given class as "this can be part of a dependency graph". Obviously any existing query would use this, but we could also use it for the subsetters (and other similar classes) that are not technically query objects but contribute to their construction. I'm thinking that in the future the admin regions etc. would also fall into this category. I think it would be a pretty minimally invasive change. How does this sound?

greenape:
Makes the dependency tracking 'opt in', which I'm not _super_ keen on.

At present, `Query.dependencies` only includes things in the dependencies if they're a subclass of `Query`. Can we do something there? e.g. another add the dependencies of a non-query object which _does_ have dependencies, similar to how lists/tuples are dealt with?

maxalbert:
That's basically what I'm thinking. I definitely want to keep `Query` objects as automatic dependencies as you suggest, and would want to only make a small tweak to `Query.dependencies`. 

Let me tweak this PR to make a suggestion (easier to demonstrate in code than to explain in words), and then you can tell me whether you like it or not.

maxalbert:
@greenape Here is what I have in mind: https://github.com/Flowminder/FlowKit/compare/f050df69738105e970aab24fc9e2d9fa8606d850...174c26c97964455f6a9420cb2f09135d263fad70

This adds a mixin class which allows us to mark _any_ class as contributing to the query dependencies. Of course, `Query` uses this by default but now other classes can, too.

Here is what the dependency tree of a daily location (with subsetting) looks like before this change:
```
<Query of type: LastLocation, query_id: '671e8a9d9e9876d59b4c68fe14a4de8e'>
  - <Query of type: JoinToLocation, query_id: 'e0f65a3d7a9c7e58bf5c1a206477b3ff'>
     - <Query of type: _SubscriberCells, query_id: '6c3aa9414107a23609488f6eb37ea095'>
        - <Query of type: EventsTablesUnion, query_id: 'c32684737b215b4c0e7a655de549b9de'>
           - <Query of type: EventTableSubset, query_id: '45697343f2c687c59e85576cfbd85307'>
              - <Table: 'events.sms', query_id: '9de507e882f1fb6b0cfcedd324e27839'>
                 - <Table: 'events.sms', query_id: '7a7f27978925c385bc44a5ec5667d7b3'>
           - <Query of type: EventTableSubset, query_id: 'c6f1fc830c3b4d9d3f91e245f0cfca9c'>
              - <Table: 'events.calls', query_id: '70a2d392f73aedcc251a6a504f17008a'>
                 - <Table: 'events.calls', query_id: '057addedac04dbeb1dcbbb6b524b43f0'>
     - <Query of type: CellToAdmin, query_id: 'f73369cb975cfd8c13a50f9ab1754106'>
        - <Query of type: CellToPolygon, query_id: 'e0196a13e5bf8d7c0c62fe5a6c63f87e'>
```

And here it is after the change (note the additional subsetter class and its dependent custom query):
```
<Query of type: LastLocation, query_id: '436d6d6f30535346caebf2be5bd624b1'>
  - <Query of type: JoinToLocation, query_id: 'b8650e506d2012382613d78bcb5d03b1'>
     - <Query of type: _SubscriberCells, query_id: 'ef4c36472f4c80bbc5e49d8969a12d88'>
        - <Query of type: EventsTablesUnion, query_id: '61d2e504b930f12f98b9f14c6b12b480'>
           - <Query of type: EventTableSubset, query_id: '1e89f27de8fceca38fe42daa2f72bd1c'>
              - <SubscriberSubsetterForFlowmachineQuery (md5='5a6b1f074a7f1e573eff006ab70d47cd')>
                 - <Query of type: CustomQuery, query_id: '5a6b1f074a7f1e573eff006ab70d47cd'>
              - <Table: 'events.sms', query_id: '9de507e882f1fb6b0cfcedd324e27839'>
                 - <Table: 'events.sms', query_id: '7a7f27978925c385bc44a5ec5667d7b3'>
           - <Query of type: EventTableSubset, query_id: '7751269d95e9979f09cf9a14e84cec32'>
              - <SubscriberSubsetterForFlowmachineQuery (md5='5a6b1f074a7f1e573eff006ab70d47cd')>
                 - <Query of type: CustomQuery, query_id: '5a6b1f074a7f1e573eff006ab70d47cd'>
              - <Table: 'events.calls', query_id: '70a2d392f73aedcc251a6a504f17008a'>
                 - <Table: 'events.calls', query_id: '057addedac04dbeb1dcbbb6b524b43f0'>
     - <Query of type: CellToAdmin, query_id: 'f73369cb975cfd8c13a50f9ab1754106'>
        - <Query of type: CellToPolygon, query_id: 'e0196a13e5bf8d7c0c62fe5a6c63f87e'>
```

Code used to produce the query and its dependency tree:
```python
import flowmachine

flowmachine.connect(flowdb_user="flowdb", flowdb_password="flowflow", redis_password="fm_redis")

custom_query = CustomQuery(
    "SELECT duration, msisdn as subscriber FROM events.calls WHERE duration < 10",
    ["duration", "subscriber"],
)
subsetter = make_subscriber_subsetter(custom_query)
dl = daily_location(date="2016-01-01", subscriber_subset=subsetter)

print_dependency_tree(dl)
```

greenape:
I like the gist of that I think - the other bit to consider is `_get_stored_dependencies`, which has a call to `self.is_stored` in there. Need to work around that in some way, maybe by adding a default `is_stored` False property to the mixin (which query can override)? I guess there'll also be an issue with `_get_query_attrs_for_dependency_graph` as well, because that will call explain on the query.

I think the _simplest_ way to handle the whole thing really would be to treat the one subsetter where this is a factor as a special case in `dependencies`, and have it pass on any queryish calls to the query object it wraps - basically same as list and tuple. Something like:

```python
    @property
    def dependencies(self):
        """

        Returns
        -------
        set
            The set of queries which this one is directly dependent on.
        """
        dependencies = set()
        for x in self.__dict__.values():
            if isinstance(x, Query):
                dependencies.add(x)
        lists = []
        for x in self.__dict__.values():
            if instance(x, SubscriberSubsetterForFlowmachineQuery):
               lists.append(x.__dict__.values())
            elif isinstance(x, list) or isinstance(x, tuple):
                lists.append(x)
        for l in lists:
            for x in l:
                if isinstance(x, Query):
                    dependencies.add(x)

        return dependencies
```

Would solve the whole thing, unless my brain fog this morning is preventing me from seeing obvious?

greenape:
Presumably makes sense to remove the `subscriber_subset_ORIG` attribute from `EventTableSubset` as part of this now?

flowmachine/flowmachine/core/query.py
```diff
@@ -741,11 +741,17 @@ def dependencies(self):
...
-            x
-            for x in self.__dict__.values()
-            if isinstance(x, list) or isinstance(x, tuple)
-        ]
+        lists = []
+        for x in self.__dict__.values():
+            if isinstance(x, list) or isinstance(x, tuple):
+                lists.append(x)
+            else:
+                parent_classes = [cls.__name__ for cls in x.__class__.__mro__]
```
greenape: Working around a circular import?

maxalbert:
@greenape Yeah, your suggestion works (kinda - see below), and I agree for expediency it's best to go with it for the purposes of this PR.

However, it requires a bit of tweaking because we can't import the subscriber subsetters in `query.py`, otherwise we'll get a circular dependency (because `Query` needs to be imported in `subscriber_subsetter.py`). I have added a workaround for this based on the class name, but in the medium term I'd prefer a different solution (along the lines of your first suggestion).

The reason why I'm not super keen long term on adding a special case for the subsetters in `Query.dependencies` is that it leads to a dependency between far-away parts of the code base that shouldn't need to be there (or at least should go only in one direction). As a consequence, if we add other kinds of objects in the future which can _also_ contribute to the query dependency tree but aren't themselves query objects, we need to remember to update `Query.dependencies` with _another_ special case. This is ok-ish if we remember to do so, but it's easy for a contributor to only look at existing examples (e.g. the subsetters) and copy/paste/adapt those but be completely oblivious that there is a special case hanging around in `Query` that also needs tweaking. We have (or at least had) a few other examples of this in the codebase where a particular responsibility is distributed across multiple locations, and it can lead to subtle and hard-to-discover bugs which can be avoided with a tidier design that allows proper separation of responsibilities (rather than making `Query` responsible for everything).

flowmachine/flowmachine/core/query.py
maxalbert: Yup

greenape:
Yeah, I see your point there. Is there an in between solution, e.g. a mixin that provides a method for 'get the dependencies this thing wraps', and one could then use duck typing?

maxalbert:
Yeah, I think there should be. I'll try to update my first attempt in the other branch with your suggestions  above and we can discuss+iterate in a separate PR.

maxalbert:
I have made the suggested changes and also tweaked the tests so that they avoid hard-coding expected query_ids (see #797).

flowmachine/flowmachine/core/subscriber_subsetter.py
```diff
@@ -19,35 +19,33 @@
...
+    @property
+    def md5(self):
+        try:
+            return self._md5
+        except AttributeError:
+            raise NotImplementedError(
+                f"Class {self.__class__.__name__} does not implement 'md5'"
+            )
+
     def column_names(self) -> List[str]:
```
greenape: No longer needed?

flowmachine/flowmachine/features/utilities/event_table_subset.py
```diff
@@ -124,7 +124,6 @@ def __init__(
         self.start = start
         self.stop = stop
         self.hours = hours
-        self.subscriber_subset_ORIG = subscriber_subset
```
greenape: :+1:

maxalbert:
Yes, good point. I have tweaked the test for `print_dependency_tree` so that it uses a query with an explicit subset.

PR #744, opened 2019-05-10
Flowetl dag skeletons

josh-gree:
Closes N/A

### I have:

### Description

This adds the outline for the two main DAGs.
- `etl_sensor` 
- `etl`

The DAGs currently do nothing and are populated with dummy tasks. The purpose of this PR is to implement the generic logic of each of these DAGs and test that they behave as expected.

Currently the `etl_sensor` DAG upon triggering unconditionally triggers the `etl` DAG - future logic will be introduced here to determine when to actually trigger the `etl` DAG (when a new file to be ingested is found). 

The `etl` DAG is structured in the following way;

Screen Shot 2019-05-10 at 13 30 12
 
The first tasks of the DAG will deal with the process of ingestion - if any of these fail we will follow the quarantine branch of the DAG and the last task will raise an exception to signal the failure of the ingestion. If all are successful then we follow the archive branch. The clean task will run in either scenario -  `flowetl/tests/integration/test_dag_run.py` tests this logic. 

Please ask questions so I can clarify the intention!

NB: I have tried to keep the tests as far away from the container as possible so they currently all run in a local airflow scheduler.


development_environment
```diff
@@ -127,5 +127,8 @@ FLOWETL_POSTGRES_USER=flowetl
 FLOWETL_POSTGRES_HOST=flowetl_db
 FLOWETL_POSTGRES_DB=flowetl
 
+# Need this to install airflow locally
+SLUGIFY_USES_TEXT_UNIDECODE=yes
```
maxalbert: Out of interest, what does this actually do?

josh-gree: It's acceptence of some licence...

flowetl/Dockerfile
```diff
@@ -6,7 +6,7 @@
 #  FLOWETL
 #  -----
 
-FROM puckel/docker-airflow
+FROM greenape/docker-airflow
```
maxalbert: Again out of interest, what are the tweaks in this image compared to the puckel one?

development_environment
josh-gree: Can't install airflow without it... https://stackoverflow.com/a/52210596

Or distribute anything that uses it...

jch1g10: Looks like it's not actually required any more: https://github.com/apache/airflow/blob/master/UPDATING.md#airflow-1103

josh-gree: Not sure that's true but will give it a go since now using the newer version but I'm sure I got an error...

flowetl/Dockerfile
josh-gree: It uses the newest release of airflow... Not sure if should pull that into the repo in its entirety or leave it where it is?

greenape: Might want to take a look at some of the open PRs on the puckle repo before switching - I rolled this as a poc for airflow now working with python 3.7

flowetl/dags/etl.py
```diff
@@ -0,0 +1,95 @@
...
+from airflow.models import DagRun, TaskInstance
+from airflow.operators.python_operator import BranchPythonOperator, PythonOperator
+from pendulum import parse
+
+default_args = {"owner": "flowminder", "start_date": parse("1900-01-01")}
+
+# pylint: disable=unused-argument
+def dummy_callable(*, dag_run: DagRun, task_instance: TaskInstance, **kwargs):
+    """
+    Dummy python callable - possibly raising an exception
```
maxalbert:
```suggestion
    Dummy python callable - raises an exception if the environment variable TASK_FAIL is defined, otherwise succeeds silently.
```

flowetl/Dockerfile
josh-gree: There's a PR but been there a while... There is a docker file in the official airflow repo perhaps should make use of that?

flowetl/dags/etl.py
```diff
@@ -0,0 +1,95 @@
...
+    Dummy branch callable
+    """
+    previous_task_failures = [
+        dag_run.get_task_instance(task_id).state == "failed"
+        for task_id in ["init", "extract", "transform", "load"]
+    ]
+
+    logging.info(dag_run)
+
+    if sum(previous_task_failures) > 0:
```
maxalbert:
```suggestion
    if any(previous_task_failures):
```

flowetl/Dockerfile
josh-gree: puckel/docker-airflow/pull/358

josh-gree: https://github.com/apache/airflow/blob/master/Dockerfile

flowetl/dags/etl.py
```diff
@@ -0,0 +1,95 @@
...
+        task_id="clean",
+        python_callable=dummy_callable,
+        provide_context=True,
+        trigger_rule="all_done",
+    )
+    fail = PythonOperator(
+        task_id="fail", python_callable=dummy_failing_callable, provide_context=True
+    )
+
+    init >> extract >> transform >> load >> success_branch  # pylint: disable=pointless-statement
```
maxalbert: Given that this is Airflow-specific syntax, maybe worth adding a comment explaining what this does?
```suggestion
    # Define upstream/downstream relationships between airflow tasks
    init >> extract >> transform >> load >> success_branch  # pylint: disable=pointless-statement
```

maxalbert: (I wish the Airflow devs simply made `set_downstream()` return the downstream task, so that one could use method chaining for better readability, but hey...)

flowetl/tests/integration/conftest.py
```diff
@@ -0,0 +1,168 @@
...
+    """
+    return os.environ.get("FLOWETL_TAG", "latest")
+
+
+# pylint: disable=redefined-outer-name
+@pytest.fixture(scope="module")
+def flowetl_container(docker_client, flowetl_tag):
+    """
+    Fixture that starts a running flowetl container and
+    yeilds the container object.
```
maxalbert:
```suggestion
    yields the container object.
```

flowetl/tests/integration/conftest.py
```diff
@@ -0,0 +1,168 @@
...
+    """
+    container = docker_client.containers.run(
+        f"flowminder/flowetl:{flowetl_tag}", detach=True
+    )
+    # breif sleep to wait for backing DB to be ready
```
maxalbert:
```suggestion
	    # brief sleep to wait for backing DB to be ready
```

flowetl/dags/etl.py
josh-gree: I like the bit shifts...

flowetl/tests/integration/conftest.py
josh-gree: I will turn on spellcheck eh!? ;-p

development_environment
josh-gree: Waheyyyyyyy! FYI didn't doubt you James just Airflow...

flowetl/tests/integration/conftest.py
```diff
@@ -0,0 +1,168 @@
...
+    sleep(2)
```
maxalbert: I would really like to avoid random sleeps as much as possible if we can. Almost every instance of hard-coded timeouts I have seen in practice is either flaky because it's on the verge of being too short, or rather long/conservative and thus wastes time during tests or in manual usage.

Given that in general we want reliable ways of verifying that our containers have started up successfully anyway (see my comments in #745), it would be great if the tests and fixtures could also use these.

I'm ok to defer this to a follow-up PR if it's too tedious now, but would like to tackle this asap.

flowetl/tests/integration/conftest.py
```diff
@@ -0,0 +1,168 @@
...
+    yield container
+    container.kill()
+    container.remove()
+
+
+@pytest.fixture(scope="module")
+def airflow_local_setup_mdl_scope():
```
maxalbert: Took me a minute to figure out what `mdl` refers to here. I suggest to either spell it out fully as `airflow_local_setup_module_scope`, or just call it `airflow_local_setup`.

flowetl/tests/integration/conftest.py
```diff
@@ -0,0 +1,168 @@
...
+
+@pytest.fixture(scope="session")
+def flowetl_tag():
+    """
+    Get flowetl tag to use
+    """
+    return os.environ.get("FLOWETL_TAG", "latest")
+
+
+# pylint: disable=redefined-outer-name
```
maxalbert: Why is this pylint directive needed? Where is this other place where `flowetl_container` is defined?

flowetl/tests/integration/conftest.py
josh-gree: I feel you! But not clear how to know when it's ready...if you have a suggestion?

flowetl/tests/integration/conftest.py
```diff
@@ -0,0 +1,168 @@
...
+    container.kill()
+    container.remove()
+
+
+@pytest.fixture(scope="module")
+def airflow_local_setup_mdl_scope():
+    """
+    Init the airflow sqlitedb and start the scheduler with minimal env.
+    Clean up aftwards by removing the AIRFLOW_HOME and stopping the
+    scheduler. It is neccesery to set the AIRFLOW_HOME env variable on
```
maxalbert:
```suggestion
    scheduler. It is necessary to set the AIRFLOW_HOME env variable on
```

flowetl/tests/integration/conftest.py
josh-gree: Yeah hang over from having two exactly same fixtures but with different scopes since pytest can't do that!?

flowetl/tests/integration/conftest.py
josh-gree: Parse the logs? That feels hacky? Of course could completely build our own Airflow container and ignore puckel et al...

flowetl/tests/integration/conftest.py
maxalbert: Not sure I fully understand what this comment is saying. Does this mean bad things will happen if the `AIRFLOW_HOME` env var isn't defined? And where is the "somewhere else" that the comment refers to?

flowetl/tests/integration/conftest.py
josh-gree: It's not it's the fixtures...above....

flowetl/tests/integration/conftest.py
```diff
@@ -0,0 +1,168 @@
...
+            ["airflow", "scheduler"], shell=False, stdout=f, stderr=f, env=env
+        )
+
+    sleep(2)
+
+    yield
+
+    scheduler.terminate()
+
+    shutil.rmtree(env["AIRFLOW_HOME"])
```
maxalbert: I'm confused why it's fine to remove this folder here, but it isn't explicitly created anywhere beforehand? How will subsequent tests make use of this?

maxalbert: Also, what is the purpose of `AIRFLOW_HOME` anyway / what is it used for?

flowetl/tests/integration/conftest.py
josh-gree:
```
tests/integration/conftest.py:38:22: W0621: Redefining name 'docker_client' from outer scope (line 22) (redefined-outer-name)
tests/integration/conftest.py:38:37: W0621: Redefining name 'flowetl_tag' from outer scope (line 30) (redefined-outer-name)
```

flowetl/tests/integration/conftest.py
josh-gree: Airflow creates the directory as soon as you import anything from it - if you don't specify `AIRFLOW_HOME` then gets created in `~` perhaps that's fine? Could explicitly create the dir?

flowetl/tests/integration/conftest.py
maxalbert: Yeah, let's think about it properly and worry about this in a separate PR.

flowetl/tests/integration/conftest.py
josh-gree: It also needs to exist for `airflow initdb` to create the sqlite db in a non-standard location.

flowetl/tests/integration/conftest.py
maxalbert: https://github.com/Flowminder/FlowKit/issues/750

flowetl/tests/integration/conftest.py
maxalbert: Yeah, if it's easy to do then it would be much clearer to explicitly create the dir. Makes it much easier for the reader to understand what `flowetl` is doing without having to know about airflow-specific behaviour (which if you don't know it already is difficult to guess and then hunt down documentation for).

flowetl/tests/integration/conftest.py
josh-gree: `pytest` imports stuff from `airflow`. If `AIRFLOW_HOME` is not set *before* this happens then airflow creates it's home dir in `~` - if I then set the env var to the place I want it to be before running the `intidb` and `scheduler` we get two airflow homes and airflow gets confused...

There is an entry in their changelog about it but have not fully understood it - think it means you *have* to set it to have a non standard airflow home?

https://github.com/apache/airflow/blob/master/UPDATING.md#removal-of-airflow_home-config-setting

flowetl/tests/integration/conftest.py
josh-gree: Cool will do!

maxalbert: It's also not quite clear to me whether this is running inside or outside the container? I'm always a bit hesitant to remove random directories on the host.

Also, if this is running outside the container then we should probably add a fixture which checks that `AIRFLOW_HOME` is set - which is probably a good idea anyway, because we don't want any accidentally undefined behaviour coming from tests that assume the env var exists but which might get an empty value.

(Although I realise we'll get a `KeyError` if it's not set because we're accessing it via `os.environ["AIRFLOW_HOME"]`. But it's too easy to inadvertently access it via `os.getenv("AIRFLOW_HOME")`, so probably better be safe than sorry.)

josh-gree: Although I think it will already exist before you even get in to that fixture...

josh-gree: Yea so add `assert False` on L96 of the conftest then run `AIRFLOW_HOME=./test_airflow_home pipenv run pytest --pdb tests/integration/test_dag_run.py` - when in pdb you will see that the dir has already been made due to imports from airflow...

josh-gree: So it would be completely cosmetic and would need me to delete the dir before creating it anyway?

josh-gree: Which is perhaps more confusing?

josh-gree: > It's also not quite clear to me whether this is running inside or outside the container? I'm always a bit hesitant to remove random directories on the host.
> 
> Also, if this is running outside the container then we should probably add a fixture which checks that `AIRFLOW_HOME` is set - which is probably a good idea anyway, because we don't want any accidentally undefined behaviour coming from tests that assume the env var exists but which might get an empty value.
> 
> (Although I realise we'll get a `KeyError` if it's not set because we're accessing it via `os.environ["AIRFLOW_HOME"]`. But it's too easy to inadvertently access it via `os.getenv("AIRFLOW_HOME")`, so probably better be safe than sorry.)

None of the tests (other than checking that fixuid works) are happening in the container...What would a fixture that checks if it's set look like? Just something that raises an exception if not? That doesn't seem like what fixtures are for to me?

josh-gree: https://github.com/MobileDynasty/pytest-env perhaps?

flowetl/tests/integration/conftest.py
```diff
@@ -0,0 +1,167 @@
...
+    yield container
+    container.kill()
+    container.remove()
+
+
+@pytest.fixture(scope="module")
+def airflow_local_setup():
+    """
+    Init the airflow sqlitedb and start the scheduler with minimal env.
+    Clean up aftwards by removing the AIRFLOW_HOME and stopping the
```
maxalbert:
```suggestion
    Clean up afterwards by removing the AIRFLOW_HOME and stopping the
```

flowetl/tests/integration/conftest.py
```diff
@@ -0,0 +1,167 @@
...
+
+    shutil.rmtree(os.environ["AIRFLOW_HOME"])
+    os.unlink("./scheduler.log")
+
+
+@pytest.fixture(scope="function")
+def wait_for_completion():
+    """
+    Return a function that waits for the etl dag to be in a specific
+    end state. If dag does not reach this state within (arbitarily but
```
maxalbert:
```suggestion
    end state. If dag does not reach this state within (arbitrarily but
```

flowetl/tests/integration/test_fixuid.py
```diff
@@ -0,0 +1,32 @@
...
+Tests that fixuid works as expected
+"""
+
+
+def test_uid(flowetl_container):
+    """
+    test that we can run the flowetl container with a specific user
+    """
+
+    user = "1002:1003"
```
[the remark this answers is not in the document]
maxalbert: Just to check my understanding of this test: is the purpose to verify that this specific uid/gid exist in the flowetl container?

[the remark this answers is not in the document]
josh-gree: So we need airflow to be able to move files on the host that have permissions different from the standard airflow user 1001:1001 - now two possible solutions to this. 1. For each new system determine the user that will be needed and rebuild the container such that airflow user has correct uid/gid. 2) this approach use fixuid which allows you to change the airflow user uid/gid at runtime. We pass `--user $UID:$GID` to docker run and then the airflow user has those ids...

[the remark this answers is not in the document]
josh-gree: So this test is making sure that the user the container starts with has correct uid/gid and should also thinking about it test that that user is named airflow...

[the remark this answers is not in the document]
josh-gree: Since the current test would pass for any container the special thing here is that airflow has the particular uid/gid

.circleci/config.yml
```diff
@@ -525,35 +525,21 @@ jobs:
...
-            expecteddag=sampledag
-            gotdag=`docker exec -it flowetl airflow list_dags | tail -2 | head -1 | tr -d '[:space:]'`
-            if [ $gotdag != $expecteddag ]; then
-              echo "Error: expected DAG not present."
-              exit 1
-            else
-              echo "Expected DAG present."
-            fi
-            docker stop flowetl
+            cd flowetl && FLOWETL_TAG=$CIRCLE_SHA1 AIRFLOW_HOME=./test_airflow_home pipenv run pytest --junit-xml=test_results/pytest/results.xml ./tests
```
greenape: Don't need the `&&` because of the `|` above. Could also stick the env vars into the environment for the command?

flowetl/Dockerfile
greenape: The airflow official one - man, why even put it out there if you're going to start it by saying don't use it? 🙄 

https://github.com/puckel/docker-airflow/pull/356 seems to include a _bunch_ of additional fixes, which is why I'm wondering if sticking with @puckel's image might not be better? For the 3.7 thing, I wonder if we'd be better off submitting a PR over there which updates the python image/adds a 3.7 tag?

flowetl/Pipfile
```diff
@@ -0,0 +1,17 @@
+[[source]]
+name = "pypi"
+url = "https://pypi.org/simple"
+verify_ssl = true
+
+[dev-packages]
+black = "==19.3b0"
+apache-airflow = "==1.10.3"
+pylint = "*"
```
greenape: We don't use pylint anywhere else, can we get the reason we're using it here on the record?

flowetl/Pipfile
```diff
@@ -0,0 +1,17 @@
+[[source]]
+name = "pypi"
+url = "https://pypi.org/simple"
+verify_ssl = true
+
+[dev-packages]
+black = "==19.3b0"
+apache-airflow = "==1.10.3"
```
greenape: Why's this pinned, out of interest?

flowetl/tests/integration/conftest.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+@pytest.fixture(scope="module")
+def flowetl_container(docker_client, flowetl_tag):
+    """
+    Fixture that starts a running flowetl container and
+    yields the container object.
+    """
+    container = docker_client.containers.run(
+        f"flowminder/flowetl:{flowetl_tag}", detach=True
```
greenape:
```suggestion
        f"flowminder/flowetl:{flowetl_tag}", detach=True, auto_remove=True
```

greenape: Just so's I know, why do we want this as a pytest fixture?

flowetl/tests/integration/conftest.py
```diff
@@ -0,0 +1,167 @@
...
+    )
+    # brief sleep to wait for backing DB to be ready
+    sleep(2)
+    yield container
+    container.kill()
```
greenape: Why a kill instead of a stop?

flowetl/tests/integration/conftest.py
```diff
@@ -0,0 +1,167 @@
...
+    container.remove()
+
+
+@pytest.fixture(scope="module")
+def airflow_local_setup():
+    """
+    Init the airflow sqlitedb and start the scheduler with minimal env.
+    Clean up aftwards by removing the AIRFLOW_HOME and stopping the
+    scheduler. It is necessary to set the AIRFLOW_HOME env variable on
+    test invocation otherwise it gets created somewhere else...
```
greenape:
```suggestion
    test invocation otherwise it gets created somewhere else.
```

Where does it get created?

flowetl/tests/integration/conftest.py
```diff
@@ -0,0 +1,167 @@
...
+        "AIRFLOW__CORE__DAGS_FOLDER": "./dags",
+        "AIRFLOW__CORE__LOAD_EXAMPLES": "false",
+    }
+    env = {**os.environ, **extra_env}
+    initdb = Popen(
+        ["airflow", "initdb"], shell=False, stdout=DEVNULL, stderr=DEVNULL, env=env
+    )
+    initdb.wait()
+
+    with open("scheduler.log", "w") as f:
```
greenape:
```suggestion
    with open("scheduler.log", "w") as fout:
```

.circleci/config.yml
maxalbert: @greenap What do you mean by "don't need the `&&`"? Put the two commands onto separate lines?

flowetl/tests/integration/test_fixuid.py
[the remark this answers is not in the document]
josh-gree: so with vanilla `puckel/docker-airflow`;

```
>docker run puckel/docker-airflow bash -c 'id -u | id -nu' 
airflow
> docker run --user 1001:1002 puckel/docker-airflow bash -c 'id -u | id -nu'
1001
id: cannot find name for user ID 1001
```
So the user has been changed at runtime but is no longer `airflow`. Whereas with ours using fixuid;

```
> docker run flowminder/flowetl bash -c 'id -u | id -nu'
fixuid: fixuid should only ever be used on development systems. DO NOT USE IN PRODUCTION
fixuid: runtime UID '1000' already matches container user 'airflow' UID
fixuid: runtime GID '1000' already matches container group 'airflow' GID
airflow
> docker run --user 1002:1003 flowminder/flowetl bash -c 'id -u | id -nu'
fixuid: fixuid should only ever be used on development systems. DO NOT USE IN PRODUCTION
fixuid: updating user 'airflow' to UID '1002
fixuid: updating group 'airflow' to GID '1003
fixuid: recursively searching path /
fixuid: chown /usr/local/airflow
fixuid: chown /usr/local/airflow/.bash_logout
fixuid: chown /usr/local/airflow/.bashrc
fixuid: chown /usr/local/airflow/.profile
fixuid: chown /usr/local/airflow/airflow.cfg
fixuid: chown /usr/local/airflow/src
fixuid: chown /usr/local/airflow/src/etl.py
fixuid: chown /usr/local/airflow/src/etl_sensor.py
airflow
```

So the user has been changed but it is still `airflow`. Have added another test to make sure this is true...

flowetl/Pipfile
josh-gree: So that it is the same as the one in the container

flowetl/Pipfile
josh-gree: Because I do use it...

flowetl/tests/integration/conftest.py
josh-gree: Doesn't really need to be I suppose...

flowetl/tests/integration/conftest.py
josh-gree: ............................................................................................................................

josh-gree:
for some reason can't reply to;

```
The airflow official one - man, why even put it out there if you're going to start it by saying don't use it? 🙄

puckel/docker-airflow#356 seems to include a bunch of additional fixes, which is why I'm wondering if sticking with @puckel's image might not be better? For the 3.7 thing, I wonder if we'd be better off submitting a PR over there which updates the python image/adds a 3.7 tag?
```

So what are you suggesting?

flowetl/Pipfile
josh-gree: It's a dev dep I use it during dev - it doesn't hurt anyone by it being there. If it's not it makes it a bit more painful for me...

.circleci/config.yml
josh-gree:
```
    environment:
      FLOWETL_TAG: $CIRCLE_SHA1 
      AIRFLOW_HOME: ./test_airflow_home 
```

Can I access the circle env var here?

josh-gree:
Other than which container to use (I think for now let's just stick with your 3.7 one and when puckel updates change to that - I will add an issue) done? @greenape @maxalbert

flowetl/tests/integration/test_fixuid.py
```diff
@@ -0,0 +1,44 @@
...
+    out = docker_client.containers.run(
+        f"flowminder/flowetl:{flowetl_tag}", "bash -c 'id -u'", user=user
+    )
+    assert out.decode("utf-8").strip() == "1002"
+
+
+def test_gid(docker_client, flowetl_tag):
```
maxalbert:
```suggestion
def test_uid_gid(docker_client, flowetl_tag):
```

flowetl/tests/integration/test_fixuid.py
```diff
@@ -0,0 +1,44 @@
...
+    """
+    test that we can run the flowetl container with a specific user
```
maxalbert:
```suggestion
    Test that the flowetl container contains a user with a specific UID/GID.
```

flowetl/tests/integration/test_fixuid.py
```diff
@@ -0,0 +1,44 @@
...
+    user = "1002:1003"
+    out = docker_client.containers.run(
+        f"flowminder/flowetl:{flowetl_tag}", "bash -c 'id -g'", user=user
+    )
+    assert out.decode("utf-8").strip() == "1003"
+
+
+def test_uid_is_airflow(docker_client, flowetl_tag):
+    """
+    test that we can run the flowetl container with a specific user
```
maxalbert:
```suggestion
    Test that the `airflow` user is present and has the expected UID/GID.
```

flowetl/tests/integration/test_fixuid.py
josh-gree: Why?

flowetl/tests/integration/test_fixuid.py
josh-gree: That's not what's happening really...we are testing that when the container starts the airflow user gets changed to have the specified UID/GID...

flowetl/tests/integration/test_fixuid.py
```diff
@@ -0,0 +1,44 @@
...
+    """
+
+    user = "1002:1003"
+    out = docker_client.containers.run(
+        f"flowminder/flowetl:{flowetl_tag}", "bash -c 'id -g'", user=user
+    )
+    assert out.decode("utf-8").strip() == "1003"
+
+
+def test_uid_is_airflow(docker_client, flowetl_tag):
```
maxalbert: Does this test subsume the previous one? No harm having both, but if this one does the same thing as the other one (plus a bit more, because it explicitly tests for the username too) then we can remove the previous one (although on second thought I'm not sure this one actually _tests_ the gid, so perhaps best to keep both).

.circleci/config.yml
greenape: Yup.

flowetl/Dockerfile
greenape: Opened https://github.com/puckel/docker-airflow/issues/364 re: the python version.

flowetl/tests/integration/test_fixuid.py
josh-gree: The first checks the UID - and second checks the GID.

flowetl/tests/integration/test_fixuid.py
josh-gree: Well `--user X:Y` will always result in the first two tests passing regardless of image. But the third is taking the UID of the running user and making sure that the user is called 'airflow'. So perhaps only need UID test and this one. Make sure that the UID is what we expect and then check that this UID is the `airflow` user?

flowetl/tests/integration/test_fixuid.py
maxalbert: Sorry, ignore me!

flowetl/tests/integration/test_fixuid.py
maxalbert: Hmm, I see what you mean. But isn't that a technical detail of how the presense of this user comes about, and the _purpose_ of the test is to check that this user is present when the container starts up?

.circleci/config.yml
josh-gree: Doesn't seem like you can reference `$CIRCLE_SHA1`...

flowetl/tests/integration/test_fixuid.py
josh-gree: Well no cos the `airflow` user will always be present - the check is that the `airflow` user has the specific uid/gid?

PR #749, opened 2019-05-10
Flowauth serverpage validation

### LINKED ISSUE #748: Client side validation and display appropriate messages
Subcategory of #54 

Added client-side validation and appropriate error messages for new server page for admin users.

BhavinPanch:
Closes #748 

### I have:

### Description

Added client-side validation and error message on new server page for admin users.


CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## [Unreleased]
 - Error is displayed in the add user form of Flowauth if username is alredy exists. [#690](https://github.com/Flowminder/FlowKit/issues/690)
 - Error is displayed in the add group form of Flowauth if group name already exists. [#709](https://github.com/Flowminder/FlowKit/issues/709)
+- validation errors displayed for add new server form of flowauth. [#749](https://github.com/Flowminder/FlowKit/pull/749)
```
```suggestion
- FlowAuth's add new server page now shows helper text for bad inputs. [#749](https://github.com/Flowminder/FlowKit/pull/749)
```

flowauth/frontend/src/ServerAdminDetails.js
```diff
@@ -114,9 +130,66 @@ class ServerAdminDetails extends React.Component {
...
+        });
+      } else if (servername.length == 0) {
+        state = Object.assign(state, {
+          name_helper_text: "Server name can not be blank."
+        });
+      } else if (!servername.match(letters)) {
+        state = Object.assign(state, {
+          name_helper_text: "Server name may only contain letters, numbers and underscores."
+        });
+      } else {
```
maxalbert: Very minor comment: I personally like to be explicit about the condition that triggers a conditional branch. In this case it's obvious that the `else` branch can only be triggered if the servername is too long, but it's easy to later add another condition in the `if` statement and forget to update the `else` branch, and then the error message displayed is inconsistent with the condition that triggered it.

How about turning this `else` branch into another `else if`?
```suggestion
      } else if (servername.length > 120) {
```
Of course then we need to add a catch-all else branch which blows up if it's triggered (because it should never be reached). I'm not sure what a good way to "blow up" is in Javascript land, though, so maybe this isn't a feasible approach. ;)

PR #768, opened 2019-05-13
Flowetl test prod env

josh-gree:
Closes N/A

### I have:

### Description

Major changes: 

1. Moves functions related to creating etl DAG into a module called `etl` - adds this to the Pipfile
2. Allows for creating the ETL dag with dummy callables for testing structure and flow of DAG but also with production callables. Whether to construct with dummy callables or not is governed by the `TESTING` env var. 

This change will allow us to separate out as much of the callable specification as possible from the airflow running environment. Hopefully all production callables will be tested in the tests of the etl module without having to start Airflow at all.


maxalbert:
I noticed that a few of files are missing license headers:
- `etl/setup.py`
- `etl/etl/__init__.py` (currently empty, but good to add it anyway)
- both test files in `etl/tests/`

Can you add them please? (We should probably go through the rest of the codebase at some point, too, I'm sure there are a few more that are missing.)

flowetl/dags/etl.py
```diff
@@ -9,90 +9,24 @@
 import logging
 import os
 
-from airflow import DAG
-from airflow.models import DagRun, TaskInstance
-from airflow.operators.python_operator import BranchPythonOperator, PythonOperator
+from airflow import DAG  # pylint: disable=unused-import
```
maxalbert: I assume you're purposely keeping the import, but what for out of interest?

josh-gree: Airflow doesn't even look in the file for a DAG if it's not imported...

josh-gree:
An empty `__init__.py` really?

flowetl/etl/setup.py
```diff
@@ -0,0 +1,11 @@
+from setuptools import find_packages, setup
+
+setup(
+    name="etl",
+    url="https://github.com/Flowminder/FlowKit",
+    maintainer="Flowminder",
+    packages=find_packages(),
+    include_package_data=True,
+    zip_safe=False,
+    install_requires=["apache-airflow==1.10.3"],
```
maxalbert: Do I remember correctly that you said the pinned version is to ensure that it's the same version as inside the docker container? Can we add a comment here reminding us of this, so that we can update it as and when the airflow docker images gets updated?

josh-gree: Yeah

flowetl/etl/tests/test_dummy_callables.py
```diff
@@ -0,0 +1,57 @@
...
+from etl.dummy_task_callables import dummy_callable, dummy_failing_callable
+
+
+def test_dummy_callable_succeeds_with_no_TASK_TO_FAIL_env_var_set():
+    """
+    Test that when no TASK_TO_FAIL env var set the dummy callable
+    succeeds.
+    """
+    dag_run = {}
+    task_instance = MagicMock()
```
maxalbert: It's fine to use a mock here, but since all you need is the `task_id` to be present, you could also define a `DummyTask` class which provides a stub implementation:
```python
class DummyTask:
    def __init__(self, task_id):
        self.task_id = task_id
```
and then use this in the test as:
```python
task_instance = DummyTask(task_id="init")
```
I tend to find this a little bit more readable, and also it makes it a bit more obvious that the only thing your tests rely on is that the `task_id` attribute is present and has a certain value. (Wheras a mock object is typically used when you need to verify an expected method call happened, rather than just providing stubbed out behaviour).

flowetl/dags/etl.py
maxalbert: Oh gosh... 🤕 

Can we please add a comment to this effect then, otherwise I can totally see someone (probably me...) accidentally cleaning this up in the future and breaking everything as a consequence.

maxalbert:
Just added a couple of inline comments. Feel free to address them before merging this in.

flowetl/dags/etl.py
josh-gree: Already have!

flowetl/etl/tests/test_dummy_callables.py
josh-gree: cool changed!

Issue #711, opened 2019-05-08
Worked examples need to be upgraded to most recent flowclient 


maxalbert:
The worked examples currently fail to execute because of recent changes to `flowclient` (e.g. required keyword args). This should be fixed so that they can be executed out of the box (esp. relevent with the addition of the dockerised worked examples in #694).

Issue #528, opened 2019-03-28
Find an alternative to folium for map embeds


greenape:
Folium, while neat, performs pretty horribly, has issues with chrome and is a bit limited. We should look out for alternatives. One thing to track is this: https://github.com/uber/kepler.gl/issues/331

maxalbert:
Probably worth looking into [GeoViews](http://geo.holoviews.org/). For multiple reasons, not least because it supports animated timelines out of the box.

jch1g10:
For the record, I tried using GeoViews for the maps, but abandoned this due to the complication of installation and difficulty getting the maps to appear in the docs-compiled notebook output.

maxalbert:
For the record, @jch1g10's recent work suggests that Mapbox is the best alternative for our current use case. As a side note, it would be good to briefly document the pros/cons of the various solutions while it's still fresh in memory (possibly in an ADR).

jch1g10:
I'd be happy to document the pros/cons I've found of Folium/GeoViews/Mapbox in an ADR. Where should this live?

maxalbert:
We used to have a folder for ADRs but somehow this seems to have gone awol. Let's re-add this in a separate PR, and then we can add another file in this folder to document this info.

PR #770, opened 2019-05-13
Use Mapbox GL for visualisations in the worked examples

### LINKED ISSUE #526: Re-enable docs notebooks execution during CI
We had to disable execution on CI of the docs notebooks, because they weren't running in a timely enough fashion. This is obviously undesirable, as we want to be sure they continue to work. We should resolve the not-running-fast enough, and re-enable execution asap.

jch1g10:
Closes #526 #528 #711

### I have:

### Description

This PR replaces the use of Folium with Mapbox GL for producing choropleths and heatmaps in the worked examples.

Running `pipenv run build` now also pre-caches flowmachine queries by executing `docs/cache_queries.py`, meaning that the example notebooks can run quickly enough to be executed during the CI docs build.

Note: Mapbox requires an API token for producing the visualisations. The dockerised worked examples are set up to read this from the `MAPBOX_ACCESS_TOKEN` env var. I haven't made it mandatory to set this env var before spinning up the worked examples container, because some people may want to run parts of the example notebooks without having to sign up for a mapbox account, but I have added a note to the docs explaining that a token is required to produce the maps.

Another note: `mapboxgl-jupyter` is not yet as full-featured as folium, and doesn't currently include some features we were previously using, such as multi-layered choropleths and time sliders. There are open issues for both of these features and more, though (see https://github.com/mapbox/mapboxgl-jupyter/issues/34, https://github.com/mapbox/mapboxgl-jupyter/issues/82), so we should continue to monitor this.


docs/cache_queries.py
```diff
@@ -0,0 +1,178 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+# This file is executed in build.sh before building the docs.
+# The purpose is to speed up execution of the example notebooks,
+# by caching some of the queries beforehand.
```
maxalbert:
```suggestion
# by caching some of the queries beforehand (which is currently
# not possible using flowclient directly).
```

jch1g10:
Running
`CI=true CIRCLE_SHA1=a010142b007b32d5f63012a16e295536a7645fbc bash <(curl -s https://raw.githubusercontent.com/Flowminder/FlowKit/a010142b007b32d5f63012a16e295536a7645fbc/quick_start.sh) examples`
should spin up the demo system with the worked examples.

maxalbert:
@jch1g10 This command does indeed spin up the demo system with the worked examples, but running the notebooks fails because it spins up `flowdb_testdata` rather than `flowdb_synthetic_data`, so the flows calculation returns the following error:
```python
FlowclientConnectionError: Something went wrong. API returned with status code 400.
Error message: 'Unable to create query object.'.
Payload: {'exception': 'No data for date: 2016-01-08 - 2016-01-09'}
```
Do you think there is an easy fix for this? If not, I'd rather address it in a separate PR.

jch1g10:
> @jch1g10 This command does indeed spin up the demo system with the worked examples, but running the notebooks fails because it spins up `flowdb_testdata` rather than `flowdb_synthetic_data`, so the flows calculation returns the following error:
> 
> ```python
> FlowclientConnectionError: Something went wrong. API returned with status code 400.
> Error message: 'Unable to create query object.'.
> Payload: {'exception': 'No data for date: 2016-01-08 - 2016-01-09'}
> ```
> 
> Do you think there is an easy fix for this? If not, I'd rather address it in a separate PR.

Oops! Should use the synthetic data now, with the latest commit:
```
CI=true CIRCLE_SHA1=7c868eb4c71994241fa1ce1e9b3a98d1186b604a bash <(curl -s https://raw.githubusercontent.com/Flowminder/FlowKit/7c868eb4c71994241fa1ce1e9b3a98d1186b604a/quick_start.sh) examples
```
(once CI finishes)

maxalbert:
Thanks! I tried to run the worked examples using this command. I can execute part of the Flows notebook, but sadly I get an internal error. I don't think this has anything to do with this PR, though, so I'm happy to merge this.

PR #774, opened 2019-05-14
Flowetl pghooks

josh-gree:
Closes N/A

### I have:

### Description

This adds the first production task callable - `render_and_run_sql_callable`. It does two things;

1. Based on the contents of the particular dag runs conf dictionary finds a sql template file in the `config` directory and populates the template also from the dag runs conf.
2. Applies the populated sql template to a DB using a `DbApiHook` the base class for what we will end up using the `PostgresHook` but could work for other DB's by using a different subclass of `DbApiHook`.

The actual construction of config is assumed to be present although is not. The next PR will add the machinery for accessing this config and passing it to Dag runs...


flowetl/Dockerfile
```diff
@@ -27,6 +27,10 @@ RUN USER=airflow && \
 RUN mv /entrypoint.sh /defaultentrypoint.sh
 COPY ./entrypoint.sh /entrypoint.sh
 
+WORKDIR /libs
```
maxalbert: No need to set the workdir since you're using absolute paths in the next two lines, and then change the WORKDIR again below?

flowetl/Dockerfile
```diff
@@ -27,6 +27,10 @@ RUN USER=airflow && \
...
+COPY ./etl /libs/etl
```
maxalbert: Doesn't really matter which directory we use since it's inside the container, but for cosmetic reasons how about using `/opt/etl` instead? (Conforms a little better to traditional Unix folder structure.)

flowetl/etl/etl/production_task_callables.py
```diff
@@ -8,7 +8,60 @@
...
 import logging
 
-from airflow.models import DagRun
+from pathlib import Path
+
+from airflow.models import DagRun, BaseOperator
+from airflow.hooks.dbapi_hook import DbApiHook
+
+
+def render_sql_callable(
```
maxalbert: Given that this doesn't just render the sql but also runs it, would be good to clarify this in the name of the callable.
```suggestion
def render_and_run_sql_callable(
```

Btw, some of the names of these functions sound a bit confusing with the `_callable` suffix. I'm wondering whether we should just do away with the suffix, or alternatively use two underscores (`...__callable`) to clarify that we're not rendering and running a sql callable here.

flowetl/etl/etl/production_task_callables.py
```diff
@@ -8,7 +8,60 @@
...
+    This function takes information from the DagRun conf to locate
+    the correct sql template file, uses the DagRun conf to populate the
+    template and runs it against the DB.
+
+    Parameters
+    ----------
+    dag_run : DagRun
+        Passed as part of the Dag context - contains the config.
+    task : BaseOperator
+        Passed as part of the Dag context - access to the instantiated
```
maxalbert:
```suggestion
        Passed as part of the Dag context - provides access to the instantiated
```
(Is this what you mean here?)

flowetl/etl/etl/production_task_callables.py
```diff
@@ -8,7 +8,60 @@
...
+        operator this callable is running in.
+    db_hook : DbApiHook
+        A hook to a DB - will most likely be the PostgresHook but could
+        be other types of Airflow DB hooks.
+    config_path : Path
+        Location of tools global config
```
maxalbert: Sorry if I'm being dumb, but what exactly does "tools global config" mean here? Does tool refer to flowetl? (And is there a missing apostrophe?)

flowetl/etl/etl/production_task_callables.py
```diff
@@ -8,7 +8,60 @@
...
+    template_name : str
+        The file name sans .sql that we wish to template. Most likely the
+        same as the task_id.
+    """
+    # dag_run.conf["template_path"] -> where the sql templates
+    # for this dag run live. Determined nby the type of the CDR
```
maxalbert:
```suggestion
    # for this dag run live. Determined by the type of the CDR
```

flowetl/etl/etl/production_task_callables.py
```diff
@@ -8,7 +8,60 @@
...
+    # this dag is ingesting. If this is voice then template_path
+    # will be 'etl/voice'...
```
maxalbert:
```suggestion
    # will be 'etl/voice'.
```

flowetl/etl/etl/production_task_callables.py
```diff
@@ -8,7 +8,60 @@
...
+    template_path = config_path / dag_run.conf["template_path"]
+
+    # template name matches the task_id this is being used
+    # in .If this is the transform task then will be transform'
```
maxalbert:
```suggestion
	    # in. If this is the transform task then it will be 'transform'
```

flowetl/etl/etl/production_task_callables.py
josh-gree: Yeah have changed

flowetl/etl/etl/production_task_callables.py
josh-gree: ....................................................................................

flowetl/etl/etl/production_task_callables.py
```diff
@@ -8,7 +8,60 @@
...
+    # and thus the template we use will be 'etl/voice/transform.sql'
+    template_path = template_path / f"{template_name}.sql"
+    template = open(template_path).read()
+
+    # make use of the Operators templating functionality
```
maxalbert:
```suggestion
    # make use of the operator's templating functionality
```

flowetl/etl/etl/production_task_callables.py
josh-gree: Have changed to `...__callable`

flowetl/etl/tests/test_render_sql_callable.py
```diff
@@ -0,0 +1,55 @@
...
+    config_dir = tmpdir.mkdir("config")
+    etl_dir = config_dir.mkdir("etl")
+    voice_dir = etl_dir.mkdir("voice")
+
+    template_name = "init"
+    init_template_path = voice_dir.join(f"{template_name}.sql")
+    init_template = "select {{number}}"
+    init_template_path.write(init_template)
+
+    # Mocks so we can use the templating feature of the operator this
```
maxalbert: This is a slightly nitpicky comment, but in the testing world there is actually a fairly specific definition for what a "mock" is (as opposed to other kinds of test doubles, such as dummy/fake/stub/spy/...). Martin Fowler explains some of the differences [here](https://martinfowler.com/articles/mocksArentStubs.html#TheDifferenceBetweenMocksAndStubs) (and the always awesome Sandi Metz also gave a great [talk](https://www.youtube.com/watch?v=URSWYvyc42M) which clarified the difference for me - mocks are a tool to test "outgoing command messages"; see her chart at 27:45 in the talk).

Technically the only mock in this test is the `mock_pghook` because this is the one you use to test the expectation that a particular method call that should have occurred. The other variables just contain just drop-in replacements for real objects to provide scaffolding that is needed to set up the test and make it work.

So to make the test slightly easier to read I'd personally rename the first two variables (as well as `MockDagRun` above) so that they don't contain the term "mock" (because if they do, as the reader of the test I would expect them to play a part in the assert statement below, which isn't the case and confused me a little at first).

I guess technically they are stub or fake objects, but I'd be happy to name them "dummy" as well. So perhaps `DummyDagRun`/`dummy_dag`/`dummy_task_op` or `FakeDagRun`/`fake_dag`/`fake_task_op` or similar?

flowetl/etl/tests/test_render_sql_callable.py
```diff
@@ -0,0 +1,55 @@
...
+    # callable runs in and a mock of the db_hook which will run the sql
+    mock_dag = DAG(dag_id="testing", default_args={"start_date": "2016-01-01"})
+    mock_task_op = BaseOperator(task_id="testing", dag=mock_dag)
+    mock_pghook = MagicMock()
+
+    # Mock of the config passed to the dag_run
+    conf = {"number": 23, "template_path": Path("etl/voice")}
+    mock_dag_run = MockDagRun(conf=conf)
+
+    render_sql_callable(
```
maxalbert: Just to be sure I'd also check that no calls have happened before the the call to `render_sql_callable`:
```suggestion
    assert mock_pghook.mock_calls == []
   
    render_sql_callable(
```

flowetl/etl/tests/test_render_sql_callable.py
```diff
@@ -0,0 +1,55 @@
...
+        dag_run=mock_dag_run,
+        db_hook=mock_pghook,
+        task=mock_task_op,
+        config_path=Path(config_dir),
+        template_name=template_name,
+    )
+
+    _, _, kwargs = mock_pghook.mock_calls[0]
```
maxalbert: ...and maybe that it the mock was called only once:
```suggestion
    assert len(mock_pghook.mock_calls) == 1
    _, _, kwargs = mock_pghook.mock_calls[0]
```

flowetl/etl/etl/production_task_callables.py
```diff
@@ -8,10 +8,63 @@
...
+    *,
+    dag_run: DagRun,
+    task: BaseOperator,
+    db_hook: DbApiHook,
+    config_path: Path,
+    template_name: str,
+    **kwargs,
+):
+    """
+    This function takes information from the DagRun conf to locate
```
maxalbert: Thanks for the detailed docstring by the way, and the additional information on the context of what the parameters mean and where they come from / how they are meant to be used. Goes a long way to help understand how the pieces work together. 👌

flowetl/etl/tests/test_render_sql_callable.py
josh-gree: Ok I get why we do this but not sure why we check no calls have happened?

maxalbert: Just my anal way to be absolutely sure that the call on the mock happened because we invoked `render_sql_callable` and not for some other reason. I agree it's unnecessary, just makes the test a tiny bit more readable imho (but I'm also happy to leave it out if you prefer).

PR #784, opened 2019-05-15
Add query objects to dependency graph

### LINKED ISSUE #767: calculate_dependency_graph should include query objects in the node's attributes
The query dependency graph would be more useful if the nodes included the query object as an attribute.

greenape:
Closes #767

### I have:

### Description

Adds the query object to the node attributes of the query dependency graph.


greenape:
Won't change the graph at all, but would mean you could traverse up a graph from the leaves and store as you went, for example.

maxalbert:
Am I correct that the only thing blocking a merge here is the alleged tiny drop in coverage - which is probably due to #789?

PR #790, opened 2019-05-15
Add ADRs

### LINKED ISSUE #780: ADRs are missing
We used to have a directory in the docs for architectural decision records (ADRs), but this is not currently present. It should be added again.

maxalbert:
Closes #780 

### I have:

- [X] Added an Architectural Decision Record (ADR), if appropriate

### Description

This PR adds a directory `docs/source/developer/adr/` for architectural decision records, and includes the ADRs that we used to have. They are also included in the docs, under `Developer > Architectural Decision Records`. While I was at it, I also moved the other two developer-related pages under the "Developer" heading.

As a side note, I'm not super happy with the term "ADR" because the purpose of these little documents feels a bit broader than just _architectural_ (and also the name is a bit clumsy in general). I saw that the Prefect folks call the equivalent documents ["PINs"](https://docs.prefect.io/guide/PINs/PIN-1-Introduce-PINs.html) ("Prefect Improvement Notices"). Maybe we can find a similar term for ours ("FINs"?). The name should also reflect whether we want to think of these mainly as records of decisions that have been taken, or as documents that play a more "active" role in actually arriving at those decisions (similar to Python PEPs).


docs/mkdocs.yml
```diff
@@ -22,10 +22,19 @@ nav:
...
+    - developer/developer.md
+    - developer/dev_environment_setup.md
+    - Architectural Decision Records (ADR):
+       - developer/adr/README.md
+       - developer/adr/0001-pipenv-for-package-and-dependency-management.md
+       - developer/adr/0002-pytest-for-testing.md
+       - developer/adr/0003-http-api.md
+       - developer/adr/0004-http-framework.md
+       - developer/adr/0005-IPC-methods.md
+       - developer/adr/0006-JWTs.md
```
jch1g10: Is it possible to add all files in `developer/adr/` here without having to specify each file individually? (I have a feeling the answer is no, but it would be convenient to be able to add new ADRs without having to remember to also add them to the nav).

greenape: No cake + eating, sadly. You can manually dictate the nav, or you can let it be autogenerated. Can't combine the two.

PR #792, opened 2019-05-15
Wait for pre-cached queries to finish before building docs

### LINKED ISSUE #787: Docs build failing on master branch
e.g. https://circleci.com/workflow-run/20bcdd20-605b-4b26-a093-09ff7e149257

I suspect this is going to be a sporadic issue, and that the solution is going to be arranging for more things to be pre-cached before the notebooks run on CI.

jch1g10:
Closes #787 

### I have:

### Description

The sporadic docs build failures noted in #787 are due to queries in the worked examples executing without some of the sub-queries having been pre-cached. This appears to be happening because the pre-cache script exits before the queries finish, and queries which have not started executing when the script finishes will go missing because there is no python interpreter to keep track of the state.

This PR modifies `cache_queries.py` so that it waits for all queries to finish before exiting, which should resolve the CI failures.


maxalbert:
This looks like a useful change in any case as far as I can tell, so we should probably merge it regardless, but the docs failure seems to be cause by something else in addition to this. 🤔

jch1g10:
> This looks like a useful change in any case as far as I can tell, so we should probably merge it regardless, but the docs failure seems to be cause by something else in addition to this. 🤔

The issue now appears to be that large queries in the pre-cache script are running before their dependencies are cached.

maxalbert:
Out of interest, how do you go about debugging this issue and where do you see that this is what's causing the failures? Are you checking the postgres logs, or can you see this from the flowmachine/flowdb ones?

jch1g10:
> Out of interest, how do you go about debugging this issue and where do you see that this is what's causing the failures? Are you checking the postgres logs, or can you see this from the flowmachine/flowdb ones?

This is from looking at the flowmachine logs, to see what SQL was being executed when the error occurred and which cache tables it was (or wasn't) referencing.

I haven't yet worked out a nice way to get the postgres logs on Circle.

maxalbert:
Yeah, I think currently the only way to get the postgres logs is to re-run the job on CircleCI with ssh enabled and actually view them inside the flowdb container (or `docker cp && scp` them to your local machine if you want to more conveniently inspect an entire postgres log file).

maxalbert:
The CI runs pass now, but given that no further commits were added I assume the underlying issue you mentioned is still present?

jch1g10:
> The CI runs pass now, but given that no further commits were added I assume the underlying issue you mentioned is still present?

Yes, I suspect the CI runs would continue to fail intermittently.

docs/cache_queries.py
```diff
@@ -170,10 +108,26 @@
...
+for query in reversed(list(nx.topological_sort(full_graph))):
+    try:
+        all_query_stores.append(full_graph.nodes[query]["query_object"].store())
+    except ValueError:
+        # Some dependencies cannot be stored
+        pass
+
+
+print("Waiting for queries to finish")
+concurrent.futures.wait(all_query_stores)
```
maxalbert: I'm wondering whether this still allows for queries to be stored in the "wrong" order (i.e., before all their dependencies have been stored), because the futures can potentially run in any order, no?

maxalbert: Also, once this works we should probably add support for optionally storing all dependencies in `Query.store()` directly.

jch1g10: The futures can run in any order, but the query state will be changed to `queued` before creating a future to store the query so queries should still be aware that their dependencies are going to be cached.

maxalbert: Ah of course, excellent point. 👍

.circleci/config.yml
```diff
@@ -654,6 +657,41 @@ jobs:
             - store_artifacts:
                 path: /home/circleci/project/docs/flowkit-docs.zip
                 destination: docs
+      - run:
+          name: Copy postgres logs
```
greenape:
```suggestion
          name: Copy postgres logs
          when: on_fail
```

greenape: Think you'll need to install psql as well btw.

jch1g10:
Right, I think I've got this working now. I've made the following changes:

All queries used in the example notebooks, and all of their dependencies, are now pre-cached before starting the `mkdocs` build.

The FlowDB script `configurate.py` uses `psutil` to get the total memory and CPU count, but this is giving incorrect values on CircleCI (36 CPUs and 72GB memory, when the actual resources are 2 CPUs and 4GB memory - see [here](https://circleci.com/docs/2.0/configuration-reference/#resource_class)). I suspect this - rather than insufficient pre-caching - was the root cause of the issue.

To get around the CPUs issue, I have set `MAX_CPUS=2`, `MAX_WORKERS=2` and `MAX_WORKERS_PER_GATHER=2` in the CircleCI config. The total memory is used to set the `shared_buffers` and `effective_cache_size` parameters, which could not be set using env vars as for the cores and workers, so I have added two new env vars `SHARED_BUFFERS_SIZE` and `EFFECTIVE_CACHE_SIZE` to allow manually setting these parameters. I'm completely open to renaming these env vars, or a different approach altogether, but this at least seems to work.

I've added a step to the `build_docs` CI job to extract the Postgres logs from FlowDB if the build fails, by reading the logs into a table and then copying them to `stdout`. This seems a little convoluted, but without direct access to the FlowDB container I think this is the only way to access the logs.

I've also made a few small tweaks to the worked examples, which didn't seem big enough to deserve PRs of their own (title, filename and colourmap changes).

PR #796, opened 2019-05-16
Add flags to allow showing which test sets yield which coverage

greenape:
Changes the codecov upload command to add a flag, which allows one to see which test suites are responsible for which test coverage when viewing reports.


.circleci/config.yml
```diff
@@ -333,7 +333,7 @@ jobs:
           path: test_results
       - run:
           name: Upload coverage
-          command: bash <(curl -s https://codecov.io/bash)
+          command: bash <(curl -s https://codecov.io/bash) -F flowmachine_Unit_Tests
```
maxalbert: Lowercase?
```suggestion
          command: bash <(curl -s https://codecov.io/bash) -F flowmachine_unit_tests
```

PR #798, opened 2019-05-16
Flowetl bookkeeping model

josh-gree:
Closes N/A

### I have:

### Description

This PR adds a sqlalchemy model for recording the process of ingestion. The model is used in `move_file_and_record_ingestion_state__callable` to record the state as files are moved from location to location. Have added tests of the model against an in memory sqlite DB.


flowetl/etl/tests/conftest.py
```diff
@@ -1,14 +1,73 @@
...
 @pytest.fixture(scope="function")
 def create_fake_dag_run():
-    def fake_dag_run(*, conf):
+    """
+    Returns a function that can be used to generate a fake DagRun with
+    specific config
+    """
+
+    def fake_dag_run(*, conf=None):
         return FakeDagRun(conf=conf)
```
greenape: I'm not advocating anything here, but any special reason this doesn't just return `Mock(conf=conf)`?

flowetl/etl/tests/test_dummy_callables.py
```diff
@@ -3,52 +3,57 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 # -*- coding: utf-8 -*-
+"""
+Tests that the testing dummy calables behave as expected
```
greenape:
```suggestion
Tests that the testing dummy callables behave as expected.
```

flowetl/etl/tests/test_dummy_callables.py
```diff
@@ -3,52 +3,57 @@
...
+"""
+from unittest.mock import patch
```
greenape: Why `patch` and not pytest's `monkeypatch`, out of interest?

flowetl/etl/tests/test_etl_record_model.py
```diff
@@ -0,0 +1,89 @@
...
+    Test that we get an exception raised when we try
+    to add a new row with an invalid cdr_type.
+    """
+    file_data = {
+        "file_name": "bob.csv",
+        "cdr_type": "spaghetti",
+        "cdr_date": pendulum.parse("2016-01-01").date(),
+        "state": "ingest",
+    }
+    with pytest.raises(Exception):
```
greenape: It just gives a generic Exception?

flowetl/etl/etl/model.py
```diff
@@ -0,0 +1,87 @@
...
+            Name of file being processed
+        cdr_type : str
+            CDR type of file being processed ("calls", "sms", "mds" or "topups")
+        cdr_date : Date
+            The date with which the files data is associated
+        state : str
+            The state in the ingestion process the file currently
+            is ("ingest", "quarantine" or "archive")
+        session : Session
+            A sqlalchmy session for a DB in which this model exists.
```
greenape:
```suggestion
            A sqlalchemy session for a DB in which this model exists.
```

flowetl/etl/etl/model.py
```diff
@@ -0,0 +1,87 @@
...
+        Add new row to the etl book-keeping table.
+
+        Parameters
+        ----------
+        file_name : str
+            Name of file being processed
+        cdr_type : str
+            CDR type of file being processed ("calls", "sms", "mds" or "topups")
+        cdr_date : Date
+            The date with which the files data is associated
```
greenape:
```suggestion
            The date with which the file's data is associated
```

flowetl/etl/etl/model.py
```diff
@@ -0,0 +1,87 @@
...
+        self, *, file_name: str, cdr_type: str, cdr_date: pendulumDate, state: str
+    ):
+
+        # This should be set more globally - not using enums
+        # because don't have a migration strategy in place
+        allowed_states = ["ingest", "archive", "quarantine"]
+        allowed_cdr_types = ["calls", "sms", "mds", "topups"]
+        # pylint: disable=no-else-raise
+        if state not in allowed_states or cdr_type not in allowed_cdr_types:
+            raise Exception(
```
greenape:
```suggestion
            raise ValueError(
```

flowetl/etl/etl/model.py
```diff
@@ -0,0 +1,87 @@
...
+    ) -> None:
+        """
+        Add new row to the etl book-keeping table.
+
+        Parameters
+        ----------
+        file_name : str
+            Name of file being processed
+        cdr_type : str
+            CDR type of file being processed ("calls", "sms", "mds" or "topups")
```
greenape: Kinda wondering if this should be an enum.

flowetl/etl/etl/model.py
```diff
@@ -0,0 +1,87 @@
...
+        cdr_date : Date
+            The date with which the files data is associated
+        state : str
+            The state in the ingestion process the file currently
+            is ("ingest", "quarantine" or "archive")
```
greenape: And also an enum?

flowetl/etl/etl/production_task_callables.py
```diff
@@ -98,13 +95,23 @@ def move_file_and_record_ingestion_state__callable(
     to_path = mount_paths[to_dir]
 
     file_name = dag_run.conf["file_name"]
+    cdr_type = dag_run.conf["cdr_type"]
```
greenape: I keep seeing these and wondering why they're hidden in the dag run conf, and not top level parameters?

flowetl/etl/tests/conftest.py
josh-gree: It did but max pointed out that this was overkill for something so simple...

flowetl/etl/tests/test_dummy_callables.py
josh-gree: No reason - I know how to use that one

flowetl/etl/tests/test_etl_record_model.py
josh-gree: Yeah with some text...

flowetl/etl/etl/model.py
josh-gree: Ok

flowetl/etl/etl/model.py
josh-gree: It should be but you have no way of doing a DB migration so it can't be...When you can migrate tables when you get a new CDR type then yeah but until then it needs to be flexible enough to add new types

flowetl/etl/etl/model.py
josh-gree: As above - although could perhaps argue that these are less likely to change...

flowetl/etl/etl/production_task_callables.py
josh-gree: Cos thats where they live - they do not exist at compile time...

josh-gree: you trigger a DAG with `conf=conf` and then that conf is in the `dag_run` for each task. When the DAG is constructed they do not exist....

flowetl/etl/tests/conftest.py
maxalbert: I suggested removing the mock in a previous PR (can't find the link right now), although this was primarily because the way in which the attribute was set on the mock was a bit roundabout and make the code hard to read.

I had no idea you can just pass values for attributes which you want set during initialisation of the mock. 😮 That makes it much more readable, so I'd be fine with that solution too. 👌 

Another reason (although I'll readily admit a much more pernickety one) is that technically this isn't a mock (see my comment [here](https://github.com/Flowminder/FlowKit/pull/774#discussion_r283762082)). But I agree for pragmatic and convenience reasons it would be totally fine to use a `Mock` here if we want to.

flowetl/etl/etl/model.py
```diff
@@ -0,0 +1,87 @@
...
+    cdr_date = Column(Date)
+    state = Column(String)
+    timestamp = Column(DateTime)
+
+    def __init__(
+        self, *, file_name: str, cdr_type: str, cdr_date: pendulumDate, state: str
+    ):
+
+        # This should be set more globally - not using enums
+        # because don't have a migration strategy in place
```
maxalbert: Out of interest, what do you mean here? Why does a missing migration strategy prevent using enums?

flowetl/etl/etl/model.py
maxalbert: Are we talking about the same kind of enum here? I assumed @greenape was talking about Python enums rather than Postgres ones?

flowetl/etl/etl/model.py
josh-gree: Because you have no way of keeping the data in this table and adding a new type very easily...

flowetl/etl/etl/model.py
josh-gree: I am talking about doing something like;

```
class Status(enum.Enum):
    """
    possible states a found file can be in.
    """

    in_process = 1
    done = 2
    quarantine = 3


class CDRType(enum.Enum):
    """
    possible types of cdr a found file could be
    """

    voice = 1
    sms = 2
    mds = 3
    topups = 4


class etl(Base):
    """
    Table for storing the state of a found file.
    """

    __tablename__ = "etl"
    __table_args__ = {"schema": "etl"}

    id = Column(Integer, primary_key=True)
    file_name = Column(String)
    cdr_type = Column(Enum(CDRType))
    cdr_date = Column(Date)
    status = Column(Enum(Status))
    time_stamp = Column(DateTime)
```

Which is both at the same time...

flowetl/etl/etl/production_task_callables.py
greenape: Ah right. So this is to enable having a single etl dag, rather than a dag per cdr type?

flowetl/etl/etl/model.py
```diff
@@ -0,0 +1,87 @@
...
+        allowed_states = ["ingest", "archive", "quarantine"]
+        allowed_cdr_types = ["calls", "sms", "mds", "topups"]
```
maxalbert: If you use @greenape's ingenius trick of inheriting from `str` when defining the enum, you can still have all the convenience of accepting strings as arguments but nicer validation without having to explicitly check the types:
```python
from enum import Enum

class CDRType(str, Enum):
    CALLS = "calls"
    SMS = "sms"
    MDS = "mds"
    TOPUPS = "topups"
```
And then you can replace the validation below (`cdr_type not in allowed_cdr_types`) simply with `self.cdr_type = CDRType(cdr_type)`. The latter will raise an error if `cdr_type` is not one of the supported ones:
```python
>>> CDRType("foo")
ValueError: 'foo' is not a valid CDRType
```
And you also still get string comparison if you want to:
```python
>>> "sms" == CDRType("sms") == CDRType.SMS
True
```

josh-gree: But why! That's just an overly complex way of doing exactly the same thing...

flowetl/etl/etl/model.py
```diff
@@ -0,0 +1,87 @@
...
+
+    __tablename__ = "etl_records"
+    __table_args__ = {"schema": "etl"}
+
+    id = Column(Integer, primary_key=True)
+    file_name = Column(String)
+    cdr_type = Column(String)
+    cdr_date = Column(Date)
+    state = Column(String)
+    timestamp = Column(DateTime)
```
maxalbert: This may not apply, but just wondering if you want an additional column for additional free-text metadata such as a comment or debug message? It's quite possible that this is a bad idea though because we want to keep debugging information in a more structured form, so just throwing it out there for consideration.

flowetl/etl/etl/production_task_callables.py
josh-gree: So the cdr_type could probably be removed (could be a default arg to all callables) but the file name cannot...

maxalbert: This is an airflow-specific limitation, right? I agree with @greenape that I'm a little confused every time I see this. I'm sure it makes sense when you think it through, but I'm still slightly hazy on the flow of information and which information is available at which time. So we should remember to document this well because I'm sure it'll catch people out otherwise.

flowetl/etl/etl/model.py
josh-gree: Could put location of the logs for the DAG run in case of a failure? But as it stands I have no envisioned use so will leave it...

flowetl/etl/tests/conftest.py
josh-gree: I am just going to leave it as it is...both do the same thing...It's not difficult to understand what is going on.

flowetl/etl/etl/production_task_callables.py
greenape: I feel like I'm being obtuse here, but _why_ can't the filename go? For this to work it is based on the date + Cdr type, right?

maxalbert: Opened #801 for this.

flowetl/etl/tests/conftest.py
```diff
@@ -1,14 +1,73 @@
...
+from sqlalchemy import create_engine
+from sqlalchemy.orm import sessionmaker
+
+from etl.model import Base, ETLRecord  # pylint: disable=unused-import
 
+# pylint: disable=too-few-public-methods
 class FakeDagRun:
-    def __init__(self, conf):
+    """
+    a fake DagRun object used for faking dag config
```
maxalbert:
```suggestion
    A fake DagRun object used for faking dag config
```

flowetl/etl/tests/conftest.py
greenape: Could also use `types.SimpleNamespace(conf=conf)`?

flowetl/etl/tests/conftest.py
```diff
@@ -1,14 +1,73 @@
...
+    """
+
+    def __init__(self, conf=None):
         self.conf = conf
 
 
+class FakeTaskInstance:
+    """
+    a fake TaskInstance object
```
maxalbert:
```suggestion
    A fake TaskInstance object
```

flowetl/etl/etl/production_task_callables.py
josh-gree: DAGs get triggered with a dictionary `conf` this is then accessible from all `TaskInstances` (running tasks) - some information is only known at runtime date of the data, type of the cdr for this file, where to store temp tables - all this goes in the `conf` dict and is used at *runtime*...

josh-gree: Things needed at *runtime* cannot be passed to the DAG as args to the python callables...they must be extracted from the `conf` dict.

flowetl/etl/tests/conftest.py
```diff
@@ -1,14 +1,73 @@
...
+
+@pytest.fixture(scope="function")
+def session():
+    """
+    Fixture to yield a session to an in memory DB with
+    correct model in place.
+    """
+    engine = create_engine("sqlite:///:memory:")
+    engine.execute(f"ATTACH DATABASE ':memory:' AS etl;")
+    Base.metadata.create_all(bind=engine)
```
maxalbert: Just for my understanding: this automatically creates the `etl_records` table referred to by the `ETLRecord` model, right?

flowetl/etl/etl/production_task_callables.py
greenape: Sure, but isn't the date stuff exactly what airflow's execution date concept about?

flowetl/etl/tests/conftest.py
josh-gree: Yeah - this is why there needs to be the unused import - `Base.metadata.create_all` creates all sqlalchemy tables in scope. Could pass the actual table to make it explicit...

flowetl/etl/etl/model.py
greenape: It is a more visible and checkable way of doing the same thing, and moves the responsibility for checking the validity of the input out of the function, no?

flowetl/etl/tests/test_dummy_callables.py
maxalbert: @greenape Out of interest, do you know what the difference is between the two? When is `monkeypatch` preferable, or are they mostly equivalent?

flowetl/etl/etl/production_task_callables.py
josh-gree: You are asking me to completely redesign what I have spent time and energy on! I have no intention of doing that....

flowetl/etl/tests/test_etl_record_model.py
maxalbert: If you know what that text is (or at least part of it), would be good to add a `match` argument, so that we're sure it's actually the exception we expect and not some other one:
```python
with pytest.raises(Exception, match="<some_text>"):
    ...
```

flowetl/etl/etl/production_task_callables.py
josh-gree: Also it does not make sense because we do not receive data at regular occurrences. Data may arrive 2 days after the date of the data - it may come 5 days at a time on a single day...the airflow's execution date concept works if you control the other side and know things are present but we do not know that!

greenape: I get the irregular data flow thing, but aren't you explicitly _providing_ the execution date when calling `trigger_dag`? Or am I misunderstanding how `trigger_dag` works? (Very possible, the docs are a bit terse on that).

josh-gree: No the execution date will be the time it was executed which is `now()` - if it was set to the date of the data then you could never retry the ingestion.

josh-gree: Then to fall back on airflow's retry machinery you get back into the issue of the irregular data flow....

flowetl/etl/tests/conftest.py
josh-gree: I could but that's less understandable...

flowetl/etl/etl/production_task_callables.py
josh-gree: The filename could go...

josh-gree: It can go from the model but it can't go from the config because I need to know it to move it - only other way is to keep the pattern that we matched hanging around and substitute the date back into it but I feel it's easier to just do that once and pass it to the conf.

josh-gree:
@greenape ? You haven't actually asked me to change anything...as far as I am concerned this is done?

flowetl/etl/etl/production_task_callables.py
greenape: That's helpful to know, thanks.

My next question then, is why not just clear the DagRun if it already exists, as part of the sense dag?

The wider purpose here is that 'retry' is intended to be a single 'move the file back to the ingest folder' thing, right?

josh-gree: Could do that but this is not part of this PR - haven't even implemented the trigger logic properly.

josh-gree: Also the retry is not automatic - it needs a person to look at the file that failed and to manually move it back to the dump location.

flowetl/etl/tests/test_dummy_callables.py
greenape: I think they're pretty interchangeable to be honest - with monkeypatch, you don't need the context handler because it'll undo everything once the test finishes, and there's a specific `monkeypatch.setenv` method. I was just curious.

flowetl/etl/etl/production_task_callables.py
greenape: @josh-gree tells me that a future PR which adds a little more meat to this section will enable a better discussion, so I'm alright to move on.

PR #826, opened 2019-05-20
Add libpq runtime requirement for psycopg2 #825

### LINKED ISSUE #825: FlowAuth docker container doesn't work with Postgres
Using a Postgres backend with Flowauth's docker container will not currently work because of a missing runtime dependency, libpq.

greenape:
Closes #825

### I have:

### Description

Installs libpq in the FlowAuth docker container to allow psycopg2 to actually work.


maxalbert:
Would it be easy to add a test for this in CI, so that we're sure it will continue to work?

greenape:
Good point @maxalbert 

Not trivially, as it turns out. Keeping this open while I do it in a non-trivial way.

greenape:
Right. I've made a few more changes to this to enable testing against Postgres on CI.

- Cypress tests now all work _without_ resetting the state of the backing db before each test
- FlowAuth's config is now done on first request and behaves much more like FlowAPI's
- The size limits for fields have been removed, because they are ignored for sqlite anyway, and don't make any difference to Postgres aside from raising an error
- The cypress tests now run against the built docker container
- Supplying an admin username and password is now mandatory
- Added a missing .coveragerc
- CLI commands are now tested

greenape:
I should also mention, I had to replace the use of `cy.visit` in the cypress tests with the workaround here: https://github.com/cypress-io/cypress/issues/3438#issuecomment-478660605

To deal with cypress not clearing cookies when using `cy.visit`.

CHANGELOG.md
```diff
@@ -10,9 +10,14 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Changed
 - The `FLOWDB_DEBUG` environment variable has been renamed to `FLOWDB_ENABLE_POSTGRES_DEBUG_MODE`.
+- FlowAuth will now automatically set up the database when started without needing to trigger via the cli.
+- FlowAuth now requires that at least one administrator account is created by providing env vars or secrets for:
+    - `FLOWAUTH_ADMIN_PASSWORD`
+	- `FLOWAUTH_ADMIN_USERNAME`
```
maxalbert:
```suggestion
    - `FLOWAUTH_ADMIN_USERNAME`
```

docs/source/install.md
```diff
@@ -173,11 +173,9 @@ FlowAuth is designed to be deployed as a single Docker container working in coop
 
 FlowAuth supports any database supported by [SQLAlchemy](https://sqlalche.me), and to connect you will only need to supply a correct URI for the database either using the `DB_URI` environment variable, or by setting the `DB_URI` secret. If `DB_URI` is not set, a temporary sqlite database will be created.
 
-On first use, you will need to create the necessary tables and an administrator account. 
+FlowAuth will attempt to create all necessary tables when first accessed, but will not overwrite any existing tables. To wipe any existing data, you can either set the `INIT_DB` environment variable to `true`, or run `flask init-db` from inside the container (`docker exec <container-id> flask init-db`).
```
maxalbert: Should this be `RESET_FLOWAUTH_DB` instead of `INIT_DB`? (Haven't tried it, just from looking at the changes in `flowauth/config.py`.)

flowauth/backend/flowauth/config.py
```diff
@@ -0,0 +1,60 @@
...
+    try:
+        with open(Path("/run/secrets") / key, "r") as fin:
+            return fin.read().strip()
+    except FileNotFoundError:
+        return default
+
+
+def get_config():
+    FLOWAUTH_FERNET_KEY = getsecret(
+        "FLOWAUTH_FERNET_KEY", os.environ["FLOWAUTH_FERNET_KEY"]
```
maxalbert: One issue I found when working on #838 was that this invocation will raise an error when using secrets only (even if the secret is defined), because Python will try to access `os.environ["FLOWAUTH_FERNET_KEY"]` _before_ the call to `getsecret()`. I fixed this by pushing the fallback to the environment variable into the helper function (see [here](https://github.com/Flowminder/FlowKit/blob/b26c54c45eeda952e58b2991f222d69802dcc974/flowapi/flowapi/config.py#L15)). Might be worth using a similar pattern here.

flowauth/backend/flowauth/config.py
```diff
@@ -0,0 +1,60 @@
...
+    ).encode()
+    Fernet(FLOWAUTH_FERNET_KEY)  # Error if fernet key is bad
```
maxalbert: Is it worth wrapping this call to `Fernet()` into a try/except block and essentially re-raise the error, just to make the error message a bit more explicit? Possibly not, just mentioning it because I always need to do a double take when I see a line of code that _looks_ like its purpose is to do something, but the only reason it exists is to throw an error.

If you don't want to go the explicit try/except route I suggest to at least explicitly ignore the return value so that it's apparent to the reader that this doesn't have any invisible side effects:
```python
_ = Fernet(FLOWAUTH_FERNET_KEY)  # Error if fernet key is bad
```

flowauth/backend/flowauth/models.py
```diff
@@ -73,8 +73,8 @@ class User(db.Model):
     """
 
     id = db.Column(db.Integer, primary_key=True, autoincrement=True)
-    username = db.Column(db.String(64), unique=True, nullable=False)
-    _password = db.Column(db.String(128), nullable=False)
+    username = db.Column(db.String(), unique=True, nullable=False)
```
maxalbert: Out of interest, are we abandoning these limits on the string length, or are they checked/enforced elsewhere?

maxalbert: Ignore me, should have read your PR description more closely. ;)

greenape: They're enforced in the frontend for the username, server names, and group names, but not elsewhere.

flowauth/backend/flowauth/config.py
greenape: Ahh yes, I spotted that. It was making me think that the `get_secret` function is a perfect example of something that should live in a common module.

flowauth/backend/tests/test_commands.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+def test_init_db_force(app):
+    """
+    DB should be wiped clean if force is true.
+    """
+    with app.app_context():
+        runner = app.test_cli_runner()
+        result = runner.invoke(init_db_command, ["--force"])
+        assert len(User.query.all()) == 0
```
maxalbert: See comment above, this test depends on the app having been initialised already. Can we check this is the case by adding `assert len(User.query.all()) > 0` at the beginning of this test?

flowauth/backend/tests/conftest.py
```diff
@@ -69,40 +75,31 @@ def test_user(app):
...
+
+
+@pytest.fixture
+def test_group(app):
+    with app.app_context():
+        group = Group(name="TEST_GROUP")
+        db.session.add(group)
         db.session.commit()
-        return user.id, user.username, "TEST_ADMIN_PASSWORD"
+        return group.id, group.name
```
maxalbert: Can we return a namedtuple here (or something similar)? Otherwise the tests below are a bit cryptic (what does `test_group[0]` mean?)

flowauth/frontend/cypress/integration/user_details_spec.js
```diff
@@ -1,20 +1,24 @@
 /* This Source Code Form is subject to the terms of the Mozilla Public
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
-
+Cypress.Cookies.debug(true);
```
maxalbert: Did you want to keep this?

maxalbert:
(I have removed the `ready-to-merge` label just in case you want to make final tweaks. Feel free to add it back once you're done.)

PR #831, opened 2019-05-21
Remove FLOWDB_SERVICES env var

### LINKED ISSUE #827: FLOWDB_HOST not what I expected
**Describe the bug**

If I run the following command to spin up flowdb with other containers;

```
 DOCKER_SERVICES="flowdb flowetl flowetl_db" make up-no_build
```

Then the `DOCKER_FLOWDB_HOST` is not set to `flowdb` as I would expect but to `flowdb_testdata` as this is the default in the `Makefile`. In order to specify starting a subset of services with the correct `DOCKER_FLOWDB_HOST` I have to run;

```
FLOWDB_SERVICES=flowdb DOCKER_SERVICES="flowdb flowetl flowetl_db" make up-no_build
```

Specifying `flowdb` twice...And possibly allowing for somebody to do something like;

```
FLOWDB_SERVICES=flowdb DOCKER_SERVICES="flowdb_synthetic_data flowetl flowetl_db" make up-no_build
```

Which would again lead `DOCKER_FLOWDB_HOST` being incorrect. 

Is there not a way of removing the `FLOWDB_SERVICES` env var completely and just finding the name of the `flowdb` service from `DOCKER_SERVICES` if it exists? Also not completely clear why you ever want to have multiple versions of `flowdb` at the same time!?

maxalbert:
Closes #827

### I have:

### Description

This PR removes the `FLOWDB_SERVICES` env var from the toplevel Makefile, so that now `DOCKER_SERVICES` is the only variable that controls which services are spun up upon `make up`.

This isn't the only possible solution. I actually like @greenape's [suggestions](https://github.com/Flowminder/FlowKit/issues/827#issuecomment-494430581) for an alternative tweak. The reason why I chose to remove this env var is because now that we also have `flowetl` and the `worked_examples` in the `DOCKER_SERVICES` there are even more possible combinations that the user may want (previously it was basically just about which version of flowdb to cheose), so it feels clunky and somewhat error-prone to support the different use cases via different env vars in the Makefile. The cleaner solution seems to me to only have `DOCKER_SERVICES` in the Makefile, and if we want more flexibility we should be moving towards proper build and startup scripts (which is probably a good idea anyway, because I think we're duplicating this functionality in a few places in various build scripts).

Anyway, just my current view on it. Happy to discuss, and also to change this PR if a different solution is more desirable.


Makefile
```diff
@@ -9,21 +9,18 @@
...
+# you'd like to use when running `make up`. Examples:
 #
-#     FLOWDB_SERVICES=flowdb_testdata make up
-#     FLOWDB_SERVICES=flowdb_synthetic_data make up
-#     FLOWDB_SERVICES="flowdb_testdata flowdb_synthetic_data" make up
+#     DOCKER_SERVICES="flowdb_synthetic_data flowapi flowmachine flowauth flowmachine_query_locker" make up
+#     DOCKER_SERVICES="flowdb_synthetic_data" make up
+#     DOCKER_SERVICES="flowdb_testdata flowetl flowetl_db" make up
 #
 # flowmachine and flowapi will connected to the first flowdb service in the list.
```
jch1g10: This will fail if there's more than one flowdb service in the list, because it will attempt to allocate them all to the same port. If they were on different localhost ports then it would spin up fine, but I'm not sure which one flowmachine and flowapi would connect to since they'll connect to FlowDB on `flowdb:5432`, which will be the same for all flowdb services. But as @josh-gree mentioned in the issue, it's not clear whether we would ever want to spin up multiple FlowDB services at the same time, so it might be fine to leave things as they are and change this comment to say that the list can contain at most one of "flowdb", "flowdb_testdata" or "flowdb_synthetic_data".

docs/README.md
```diff
@@ -15,7 +15,11 @@ This will create any necessary Docker containers, run all notebooks, generate ma
 
 For local development, you can also run `pipenv run serve`, which will build the docs, and then start a live reloading web server. You can view the docs at http://localhost:8000, and any changes you make to markdown files in the `build` directory will be reflected immediately.
 
-Alternatively, you can build the docs by sourcing `development_environment` and  running `FLOWDB_SERVICES="flowdb_synthetic_data" make up` in the project root to bring up the required containers, and then running `pipenv run mkdocs` directly.
+Alternatively, you can build the docs by sourcing `development_environment` and  running
+```
+DOCKER_SERVICES="flowdb_testdata flowapi flowmachine flowmachine_query_locker" make up
```
jch1g10:
```suggestion
DOCKER_SERVICES="flowdb_synthetic_data flowapi flowmachine flowmachine_query_locker" make up
```

Makefile
greenape: Could use some gnu make voodoo to error there using something like.. `$(words $(filter flowdb%, $(DOCKER_SERVICES))` and `$(error..`?

maxalbert: Thanks! Following @greenape's suggestion I added an explicit check for multiple flowdb services in the Makefile so that it now throws an error.

Issue #789, opened 2019-05-15
codecov reports wrong for flowmachine.utils


greenape:
For reasons I can’t riddle out, flowmachine.utils does not have _any_ test coverage from the flowmachine unit tests on codecov.

There is ~45% coverage from the integration tests, but zero from the unit tests.

As far as I can see, test_utils.py should give this file 90+% coverage.

greenape:
I’ve dropped codecov an email, because it honestly looks to me like this is something wrong at their end.

PR #838, opened 2019-05-22
Fix secrets quickstart

### LINKED ISSUE #836: FlowAPI fails to start up when using secrets quickstart instructions
**Describe the bug**
FlowAPI fails to start up when using secrets quickstart instructions.

**Product**
FlowAPI

**Version**
Current master.

**To Reproduce**
Follow the instructions in the [Installation > Secrets Quickstart](https://flowminder.github.io/FlowKit/install/#secrets-quickstart) section of the docs (note that they currently need tweaking to work, which will be fixed in an upcoming PR). While most services start up successfully, FlowAPI fails to start up. The reason for this is that some config options are only defined via secrets and the environment variables are undefined, but [`get_config`](https://github.com/Flowminder/FlowKit/blob/master/flowapi/flowapi/config.py#L33) tries to access them anyway.

**Expected behavior**
All services should start up successfully.

### LINKED ISSUE #837: OpenSSL certificate generation command in secrets quickstart docs is broken
The instructions [Installation > Secrets Quickstart](https://flowminder.github.io/FlowKit/install/#secrets-quickstart) are slightly buggy and currently don't always allow to spin up a working system. In particular, the command which creates an OpenSSL certificate can fail.

There are two reasons for this. Firstly, it contains the instruction `[[ "Darwin" -eq "$(uname -s)" ]]`. The `eq` does a numeric comparison rather than a string comparison, and due to some bash weirdness this leads to the command always succeeding. On a non-Darwin system this means that the configuration file can't be found and as a result the `cert.pem` file won't be generated. (Which leads to an SSL error when hypercorn attempts to start up inside the container.)

Another issue is that on CentOS the config file does not exist at the location`/etc/ssl/openssl.cnf` but rather at `/etc/pki/tls/openssl.cnf`, which equally results in a broken certificate and an SSL error when hypercorn starts up.

maxalbert:
Closes #836
Closes #837

### I have:

- [] Formatted any Python files with [black](https://github.com/ambv/black)

### Description

- Fixes `secrets-quickstart.sh` so that the SSL generation works more robustly now. I also added a few more safeguards which abort if something goes wrong (e.g. the networks of an existing docker stack can't be removed within the timeout).

- Updates the docs instructions in `install.md` with similar fixes. These could do with some more brushing up, but I don't want to spend too much time on this now, partly because I expect them to change further and partly because we may want to fully wrap them up in the script and not explicitly mention them in the docs.

- Fixes `flowapi/config.py` so that all config options are now read by checking the secret first and falling back to the environment variable if defined. If neither is defined, an error is raised. This makes the startup more robust, or at least makes errors explicit when they occur. It's still not ideal that these will only happen during startup of the container and only be visible in the docker logs, this needs further tweaking once we fully decide how to handle these config settings.


maxalbert:
I should mention that I haven't added a test for the docker swarm startup with secrets yet. We definitely want this, but it may need some thinking how to best get this working on CircleCI, and I'd prefer to do this in a separate PR once I have a better understanding on how to iron out some remaining wrinkles with the production deployment. Let me know if this ok (I'll open an issue for it in any case).

maxalbert:
Strange. Codecov [reports](https://codecov.io/gh/Flowminder/FlowKit/compare/bc32e37c295d0068191e4d2f2d84fee8f6f701af...b26c54c45eeda952e58b2991f222d69802dcc974/src/flowapi/flowapi/config.py?before=flowapi/flowapi/config.py) reduced coverage again in `flowapi/config.py`, but the terminal [output](https://circleci.com/gh/Flowminder/FlowKit/58657) reports 100% coverage. Presumably another instance of #789?

greenape:
I think the terminal is wrong, because the .coveragerc doesn't have branch coverage on for some reason: https://github.com/Flowminder/FlowKit/blob/fix-secrets-quickstart/flowapi/.coveragerc

greenape:
Looks like a couple of monkeypatch tests will sort the coverage - one which unsets at least one required env var, and one which replaces `open` with [`mock_open`](https://docs.python.org/3/library/unittest.mock.html#mock-open) so that one can emulate reading a secrets file?

maxalbert:
These tests already exist, no? See [flowapi/tests/unit/test_secrets.py](https://github.com/Flowminder/FlowKit/blob/175be255d1f294faedb49ad83caa90d97d4e34fe/flowapi/tests/unit/test_secrets.py). Also, when I run the tests locally with your PR #841 merged (so that branch coverage is enabled) I still get 100% coverage in the terminal. And adding a debugging message in the line before `return fin.read().strip()` (which supposedly is not getting executed) prints this debugging message, so it must reach that branch. ¯\_(ツ)_/¯ Am I missing something?

greenape:
Ah you’re right, sorry! Yeah, looks like #789 again.

flowapi/flowapi/config.py
```diff
@@ -6,46 +6,57 @@
...
-        Name of the secret.
-    default: str
-        Default value to return if the file does not exist
+        Name of the secret / environment variable.
 
     Returns
     -------
     str
-        Value in the file, or default
+        Value in the file, or value of the environment variable, or default if defined.
```
greenape: Default’s gone now

PR #845, opened 2019-05-23
Fix sporadic flowauth fails

### LINKED ISSUE #844: Sporadic flowauth test fails
FlowAuth tests are periodically failing on the first test, e.g. https://circleci.com/gh/Flowminder/FlowKit/59539 or see https://dashboard.cypress.io/#/projects/67obxt/runs/2933/failures/9f4670b4-f665-4edb-afa0-5f4410600139

This seems to be because cypress is _sometimes_ not sending the cookie.

greenape:
Closes #844

### I have:

### Description

Added two multiprocessing `Event`s, which prevent more than one forked flowauth process from trying to initialise the database, or create demo data, and will cause others to wait until this is done.

### DIFF
--- a/flowauth/backend/tests/test_commands.py
+++ b/flowauth/backend/tests/test_commands.py
@@ -64,6 +64,7 @@ def test_init_db_force(app):
             add_admin_command, ["DUMMY_ADMINISTRATOR", "DUMMY_ADMINISTATOR_PASSWORD"]
         )
         assert len(User.query.all()) > 0
+        app.config["DB_IS_SET_UP"].clear()
         result = runner.invoke(init_db_command, ["--force"])
         assert len(User.query.all()) == 0
 
@@ -74,6 +75,7 @@ def test_demo_data(app):
     """
     with app.app_context():
         runner = app.test_cli_runner()
+        app.config["DB_IS_SET_UP"].clear()
         result = runner.invoke(demodata)
         assert len(User.query.all()) == 2
         assert len(Group.query.all()) == 3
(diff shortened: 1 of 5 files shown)


flowauth/backend/tests/test_commands.py
```diff
@@ -64,16 +64,73 @@ def test_init_db_force(app):
...
 
 
+def test_demo_data_only_sets_up_once(app, caplog):
+    """
+    Demo data should only be created once.
+    """
+    with app.app_context():
+        runner = app.test_cli_runner()
+        app.config["DB_IS_SET_UP"].clear()
+        result = runner.invoke(demodata)
```
maxalbert: Just to double-check: these two instances of `runner.invoke(demodata)` will run in serial, i.e. the first one will definitely finish before the second one is run? Just want to make sure this test can't accidentally fail if they run concurrently.

flowauth/backend/tests/test_commands.py
```diff
@@ -64,16 +64,73 @@ def test_init_db_force(app):
...
+
+
+def test_demo_data_skips_when_in_progress(app, caplog):
+    """
+    Shouldn't attempt to create demo data if something else is.
+    """
+    with app.app_context():
+        runner = app.test_cli_runner()
+        app.config["DB_IS_SET_UP"].clear()
+        app.config["DB_IS_SETTING_UP"].set()
```
maxalbert: Technically the use of `DB_IS_SETTING_UP` as a flag is an internal implementation detail, but I can't think of an easy way to reliably trigger the test condition in a different way so I guess it's fine to use it in the test too.

greenape: Actually that's a good point - these should probably be on `g` instead really. I'll change it.

flowauth/backend/tests/test_commands.py
greenape: Definitely serial.

PR #854, opened 2019-05-28
Flowauth token ui change

### LINKED ISSUE #834: Cleaner flowauth ui for permissions
The current FlowAuth UI for choosing API and spatial aggregation permissions is _very_ granular and includes a _lot_ of information, which will grow further as more API methods are added.

We should look to hide more of this information by default, on the basis that the most common usage flow is likely to be 'give me the maximum access I'm allowed'.

I'd like to preserve the level of granularity possible, but to provide a more user friendly default flow, and hide the majority of the checkboxes by default.

BhavinPanch:
Closes #834 

### I have:

### Description

I have added Material UI Expansion Panels to declutter flowauth UI


greenape:
Cool. Can you add some screenshots to show what the revised UI looks like?

BhavinPanch:
Here is flowauth token permission UI.

Screenshot 2019-05-28 at 10 04 44
Screenshot 2019-05-28 at 10 04 57
Screenshot 2019-05-28 at 10 05 05
Screenshot 2019-05-28 at 10 05 17

greenape:
That looks pretty promising to me @jch1g10 and @maxalbert can you take a look as well?

maxalbert:
Yes, this looks promising. I like the idea of only showing only a couple of "global" / "all permissions" checkboxes by default.

First impressions from a quick test run:

- The two "global" checkboxes (for the full permissions) work and toggle all the individual ones. However, I can't actually toggle any of the individual checkboxes any more (clicking them has no effect), so this needs to be fixed.

- With the demo data, the test user only has partial permissions for certain aggregation units (whereas the ones for "cell" and "site" are not allowed and are shown as greyed out when the panel is expanded). However, when I toggle the global checkbox for "Aggregation Units" the cell/site ones also get selected, which should not happen.

- Need to add tests for a few workflows, including the above ones.

- Lower priority (just a minor usability issue): When one of the panels (say, for "API permissions") is expanded so that all the individual checkboxes are visible, but the global checkbox unchecked, once I toggle this global checkbox the panel collapses so that the individual checkboxes are now hidden. I'd prefer to keep the panel open in this case so that I can see which permissions I'm now granted. (However, I do like the way the reverse case works where the panel expands if the global permissions box is checked and it was previously collapsed.)

flowauth/frontend/cypress/integration/add_new_token.js
```diff
@@ -54,6 +54,165 @@ describe("Token generation", function() {
...
+      });
+    //uncheck first sub-level checkbox
+    cy.get("#unit")
+      .first()
+      .click({
+        force: true
+      });
+    cy.get("#units").should("not.be.checked");
+  });
+  it("Top-level Aggrigation unit checkbox intermidiate", function() {
```
greenape:
```suggestion
  it("Top-level Aggregation unit checkbox indeterminate", function() {
```

Can these be fixed everywhere? I know it is a bit anal, but I find the typos really distracting! See also: checkboxs when it should be checkboxes.

PR #864, opened 2019-05-29
Public key linkage between FlowAuth and FlowAPI

### LINKED ISSUE #89: Public key auth for JWTs
Per https://blog.miguelgrinberg.com/post/json-web-tokens-with-public-key-signatures via @maxalbert, we'd be better off using a public key solution for the JWTs as opposed to the current shared secrets.

Should - 
- allow supplying public/private key to auth manager
- Maybe support generating a key pair inside the auth manager
- Provide a mechanism to get the public key from the auth manager
- Make use of the aud key in the JWT to ensure that tokens are valid only on specific servers

### LINKED ISSUE #735: Have generate-jwt accept named arguments instead of positional ones
Spinning out of a comment on PR #733:

Currently the `generate-jwt` CLI utility accepts positional arguments:
```
generate-jwt TEST_USER SECRET 1 TEST_SERVER --all-access http://localhost:9090
```
For readability (and to avoid accidentally mixing up the order), it would be nice to have these as named arguments instead:
```
generate-jwt --all-access \
    --username TEST_USER \
    --secret-key SECRET \
    --lifetime 1 \
    --flowapi-server-name TEST_SERVER \
    http://localhost:9090
```

### LINKED ISSUE #779: Default secret (e.g. for worked examples) is inconsistent with the one in `generate-jwt` help string
I wanted to use `generate-jwt` to generate a token to run the Jupyter notebook for the worked examples.

Running `generate-jwt --help` shows a useful example invocation:
```
generate-jwt TEST_USER SECRET 1 TEST_SERVER --all-access http://localhost:9090
```
However, the secret (`SECRET`) is uppercase, which is inconsistent with the default secrets in the worked examples container. It would be helpful if those were consistent with each other, so that the user doesn't need to look up or guess the correct value of `JWT_SECRET_KEY`.

So we should either change `SECRET` to `secret` in the example invocation in the help string, or alternatively change the default secret key from `secret` to `SECRET` (the latter is perhaps the slightly more consistent option?).

greenape:
Closes #89 closes #779 closes #735

### I have:

### Description

This changes the way JWTs are verified to use RSA public keys instead of a shared secret.

FlowAuth now expects to get a private key, with which it will sign all the JWTs it generates. FlowAPI now expects to get the corresponding _public_ key, with which it can verify the JWTs. This means there are no longer any shared secrets, only a single key pair. The public key can be safely distributed, because it allows you only to verify JWTs.

Adds: 

- Two new env vars/secrets:
  - `PRIVATE_JWT_SIGNING_KEY` environment variable/secret added to FlowAuth, which should be a PEM encoded RSA private key, optionally base64 encoded if supplied as an environment variable.
  - `PUBLIC_JWT_SIGNING_KEY` environment variable/secret added to FlowAPI, which should be a PEM encoded RSA public key, optionally base64 encoded if supplied as an environment variable.
- A new admin page to FlowAuth, which allows you to view/download the _public_ key.


docs/source/install.md
```diff
@@ -152,7 +152,17 @@ FlowAuth will attempt to create all necessary tables when first accessed, but wi
...
+
+`PRIVATE_JWT_SIGNING_KEY` is used to sign the tokens generated by FlowAuth, which ensures that they can be trusted by FlowAPI. When deploying instances of FlowAPI, you will need to supply them with the corresponding _public_ key, to allow them to verify the tokens were produced by this instance of FlowAuth.
+
+You can use `openssl` to generate a private key:
+
+```bash
+openssl genrsa -out flowauth-private-key.key 4096
+```
+
+And then create a public key from the key file (`openssl rsa -pubout -in  flowauth-private-key.key -out flowapi-public-key.pub`), or download it from FlowAuth once started. Should you need to supply the key using environment variables, rather than secrets (not recommended), you should base64 encode the key - FlowAuth and FlowAPI will automatically decode it for use.
```
maxalbert: If it's easy to give an example command to show how to base64-encode the key this would be useful.

.circleci/config.yml
```diff
@@ -83,26 +82,27 @@ executors:
...
...L1dPQVFnMjJJeHI3Rg0KSDRWVis4OWpYN0hNajhZVXgrWk54bkc3RUM2TFh0QkUzT3ZVSUpXQWxQOTFPeHVKVkhMQUZSM3B3MWJTTHBROQ0KdGFpeXhhTUdrTVVpa3QySVdKZ0VZckJQS0lMOHZqR0RSNVJxTWpld2JrNUhoOWdsL2x5dUthQ2xlTEc4SThBSA0KSVFJUEFvSUJBRzVQd1MvK3E2VDFLTXVUUVNnQ0J3VlBjcWVLTDFQa2VCWlFWRDJKMGlzQVFzQ2VFaUhHVGJDMQ0KR0ZVSjZsUEg5S0taeS9JU1dJL3YxMEM1TzJCa0QrZmYvRnBqdFRyci9vWVphSTlWSEhPSW04TWNXd2ZvTjErOQ0KcTBhNm1HRklPZzdMTzhBdkdyemRTZlA2elE3aTFHMEdNaHBta0VRMk9iQWNZSTN2UkRzdHQzMlIwUlB1akljdw0KaWNEWUlhVGJ4TVFRN2k0ZzhBWjdrTkRvdVR4U2hyNFRWbGRuTXdXdEdRUHlHSE96UUJqckhXVUFQeWRCOXptQQ0Kc1V4aHVmTDRmalNKNDAyY21kR2prUW9LRXF1NlFMQVNXU1JoS0dWUDNIMHdQR1Z0cWFNNVVOVGU3SUU4UlpiTw0KSHlJSTNJU3dDOVl6WE1HallXellFOXI2YmdpZzFoa0NnZ0VBY09WbnpBRlJNQWhyc3JBTjVjTmdabS9VWUthVA0KVXozeFBKNStLVXp2b0w4a2t0VHgyVTVRempkenVlZmxiOVVDT1AveWp6TElnUlRhTnRKdjZVODRURUNWSVQyRA0KVUtIVkdlQ0hGUE5MZU9UbE5PTVVjOFo0SGw1Tzd4MjcwQXhXTTJndXBncUVNUGIxbnJnWGF2ekNDZ2lSU2lnZQ0KbUx6a0FpOElEYzduZkxPUXBjaGVNSG02TkNBcDRRZ1pFYkRaYmp6VUdrUTlRMGNlT0wwLy9JYzIwdmQ1a2pWbw0KTjI5dnJuYStCTW0rejJ3Q2FJSFZmNWlWbzhFdGxwN2E2OEJnQkpEUVVxTzh4M1hpMWhCWkxuWlNJSEFWYjZXLw0KaGd1UmVkRlJucXQ5QWhVdVN0ZlRzUExoUDlXNGN5bDd2QzcyZ3k4K0lkbTgwTkFCaGpaUE0zelF1QT09DQotLS0tLUVORCBSU0EgUFJJVkFURSBLRVktLS0tLQ==
```
maxalbert: The `JWT_SIGNING_KEY` env vars are pretty lengthy, and we repeat them multiple times in the CircleCI config. Is there a way to only define them once and then re-use them in the different environments that need them?

greenape: Actually I think we can take them out completely- they’re available as env vars from the org context.

docs/source/install.md
```diff
@@ -252,64 +275,92 @@ until [ -z "$(docker service ls --filter label=com.docker.stack.namespace=secret
...
+
 # Remove existing secrets
 echo "Removing existing secrets"
+
 docker secret rm FLOWMACHINE_FLOWDB_PASSWORD || true
 docker secret rm FLOWMACHINE_FLOWDB_USER || true
 docker secret rm FLOWAPI_FLOWDB_PASSWORD || true
 docker secret rm FLOWAPI_FLOWDB_USER || true
-docker secret rm POSTGRES_PASSWORD || true
+docker secret rm FLOWDB_POSTGRES_PASSWORD || true
```
maxalbert: I'm definitely happy to rename `POSTGRES_PASSWORD` to `FLOWDB_POSTGRES_PASSWORD` (or `FLOWDB_ADMIN_PASSWORD` or similar), but I can't see any changes in the flowdb code that would reflect this renaming. Does it pick up the correct secret file? (Haven't tried it yet.)

maxalbert: Never mind, I assume the fact that `POSTGRES_PASSWORD_FILE` is set to the correct value will suffice.

CHANGELOG.md
```diff
@@ -8,18 +8,25 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
 
+- `PRIVATE_JWT_SIGNING_KEY` environment variable/secret added to FlowAuth, which should be a PEM encoded RSA private key, optionally base64 encoded if supplied as an environment variable.
+- `PUBLIC_JWT_SIGNING_KEY` environment variable/secret added to FlowAPI, which should be a PEM encoded RSA public key, optionally base64 encoded if supplied as an environment variable.
 - The dev provisioning Ansible playbook now automatically generates an SSH key pair for the `flowkit` user. [#892](https://github.com/Flowminder/FlowKit/issues/892)
 
 ### Changed
 
 - The quick-start script now only pulls the docker images for the services that are actually started up. [#898](https://github.com/Flowminder/FlowKit/issues/898)
+- FlowAuth and FlowAPI are now linked using an RSA keypair, instead of per-server shared secrets. [#89](https://github.com/Flowminder/FlowKit/issues/89)
+  - `FLOWAUTH_ADMIN_PASSWORD`
```
maxalbert: These two env vars are not related to the RSA keypair, right? (But from the indentation they appear to be.) Is there a line missing which explains their meaning?

docs/source/install.md
```diff
@@ -252,64 +270,92 @@ until [ -z "$(docker service ls --filter label=com.docker.stack.namespace=secret
...
 openssl req -newkey rsa:4096 -days 3650 -nodes -x509 -subj "/CN=flow.api" \
     -extensions SAN \
     -config <( \
-        cat $(   ( [[ -e /System/Library/OpenSSL/openssl.cnf ]] \
-                   && echo "/System/Library/OpenSSL/openssl.cnf" ) \
-              || ( [[ -e /etc/ssl/openssl.cnf ]] \
-                   && echo "/etc/ssl/openssl.cnf" ) \
-              || ( [[ -e /etc/pki/tls/openssl.cnf ]] \
-                   && echo "/etc/pki/tls/openssl.cnf" ) ) \
+        cat $(   ( [[ -e /System/Library/OpenSSL/openssl.cnf ]] && echo "/System/Library/OpenSSL/openssl.cnf" ) \
```
maxalbert: I agree it would be nicer to have these all on a single line. The reason I broke this up in a previous PR is because the docs page is fairly narrow, so the complete line isn't visible if it's too long, which I personally find irritating because you need to keep scrolling right and left if you want to follow the script. Would you be happy to keep these lines split so that they are fully readable in the docs without horizontal scrolling?

docs/source/install.md
```diff
@@ -252,64 +270,92 @@ until [ -z "$(docker service ls --filter label=com.docker.stack.namespace=secret
...
+
+# Deploy the stack
 
 echo "Deploying stack"
-docker stack deploy --with-registry-auth -c docker-stack.yml secrets_test
+docker stack deploy -docker-stack.yml secrets_test
 ```
 
-This will bring up a single node swarm, create random 16 character passwords for the database users, generate a certificate valid for the `flowkit.api` domain (and point that to `localhost` using `/etc/hosts`), pull all necessary containers, and bring up the API with `secret` as the JWT secret key.
+This will bring up a single node swarm, create random 16 character passwords for the database users, generate a certificate valid for the `flowkit.api` domain (and point that to `localhost` using `/etc/hosts`), pull all necessary containers, and bring up the API with a newly generated JWT signing key pair.
```
maxalbert: Isn't the statement "This will [...] bring up the API with a newly generated JWT signing key pair" a bit confusing, because the key pair isn't generated in the API but rather in FlowAuth?

flowapi/flowapi/config.py
```diff
@@ -45,8 +51,37 @@ def get_secret_or_env_var(key: str) -> str:
...
+    """
+    try:
+        return serialization.load_pem_public_key(
+            key_string.encode(), backend=default_backend()
+        )
+    except ValueError:
+        try:
+            return load_public_key(base64.b64decode(key_string).decode())
+        except (binascii.Error, ValueError):
+            raise ValueError("Failed to load key.")
```
maxalbert:
```suggestion
            raise ValueError("Failed to load public key.")
```

flowauth/backend/flowauth/admin.py
```diff
@@ -868,3 +866,25 @@ def list_all_tokens():
...
+@blueprint.route("/public_key")
+@login_required
+@admin_permission.require(http_exception=401)
+def get_public_key():
+    """
+    Get the public key which can be used to verify tokens for this
+    flowauth server.
+    """
+    key = (
+        current_app.config["PRIVATE_JWT_SIGNING_KEY"]
```
maxalbert: Not a big deal, but it's slightly non-intuitive that you need to load the private key from the config in order to return the public one. Would it make sense to define `current_app.config["PUBLIC_JWT_SIGNING_KEY"]` in the config so that we can retrieve it directly where needed?

flowauth/backend/flowauth/token_management.py
```diff
@@ -244,8 +253,8 @@ def generate_token(
     ----------
     username : str
         Username for the token
-    secret : str
-        Shared secret to sign the token with
+    private_key : str
+        Private key to use to sign the token
```
maxalbert: Can we mention the valid formats here? (E.g. do base64-encoded keys work?)

flowauth/backend/flowauth/token_management.py
```diff
@@ -255,7 +264,7 @@ def generate_token(
 
     Examples
     --------
-    >>> generate_token(flowapi_identifier="TEST_SERVER",username="TEST_USER",secret="SECRET",lifetime=datetime.timedelta(5),claims={"daily_location":{"permissions": {"run":True},)
+    >>> generate_token(flowapi_identifier="TEST_SERVER",username="TEST_USER",private_key="SECRET",lifetime=datetime.timedelta(5),claims={"daily_location":{"permissions": {"run":True},)
```
maxalbert: The example `private_key="SECRET"` gives the impression that the private key can be an arbitrary secret string. Instead, I'd prefer to give some indication about valid formats if possible, or at least make it clear that arbitrary strings won't work.

flowauth/backend/tests/test_token_admin.py
```diff
@@ -13,3 +16,20 @@ def test_list_tokens_for_server(client, auth, test_admin):
     response = client.get("/admin/tokens", headers={"X-CSRF-Token": csrf_cookie})
     assert 200 == response.status_code
     assert 1 == len(response.get_json())
+
+
+@pytest.mark.usefixtures("test_data")
+def test_get_server_public_key(app, client, auth):
+    """Test that admin can get the flowauth server's public key"""
```
maxalbert: Should we add a test that regular users can't retrieve the public, or is this already covered somewhere?

Thinking about it, it actually shouldn't be a problem if they _can_ retrieve it, although there is no need to do so.

flowauth/backend/tests/test_token_admin.py
```diff
@@ -13,3 +16,20 @@ def test_list_tokens_for_server(client, auth, test_admin):
...
+    response, csrf_cookie = auth.login("TEST_ADMIN", "DUMMY_PASSWORD")
+
+    response = client.get("/admin/public_key", headers={"X-CSRF-Token": csrf_cookie})
+    key_json = response.get_json()
+    expected_key = load_public_key(environ["PUBLIC_JWT_SIGNING_KEY"]).public_bytes(
```
maxalbert: Since we use the `public_bytes(
        encoding=serialization.Encoding.PEM,
        format=serialization.PublicFormat.SubjectPublicKeyInfo,
    )` mantra in a bunch of places, is it worth pushing this into the definition of `load_public_key` itself? (Or are we making use of the return value of `load_public_key` in different ways elsewhere?)

flowauth/frontend/cypress/integration/add_new_server.js
```diff
@@ -14,95 +14,95 @@ describe("Server management", function() {
...
+    });
     cy.contains("#name-helper-text").should("not.exist");
   });
   it("Add server name more than 120 characters", function() {
     cy.get("#new").click();
     //adding username
-    cy.get("#name").type(
-      "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
-    );
+    cy.get("#name").type("a".repeat(121), {
```
maxalbert: Nice!

flowauth/backend/flowauth/admin.py
greenape: I had it set up like that originally, but it seemed weird to have a config item which was a. only used in one place, and b. was a function of another config item.

maxalbert: Personally, I find readability and clarity sufficient as a reason to have a separate definition even if it's only used in a single place. (After all, it's also useful to wrap up code in a function if it hides unnecessary details even if that function is only called in a single place.) But not a big deal in this case, I'll leave it to you what you prefer.

docs/source/install.md
```diff
@@ -152,57 +149,69 @@ FlowAuth will attempt to create all necessary tables when first accessed, but wi
 
 FlowAuth requires you to create at least one admin user by setting the `FLOWAUTH_ADMIN_USER` and `FLOWAUTH_ADMIN_PASSWORD` environment variables or providing them as secrets. You can combine these environment variables with the `INIT_DB` environment variable.
 
-You _must_ also provide two additional environment variables or secrets: `FLOWAUTH_FERNET_KEY`, and `SECRET_KEY`. `FLOWAUTH_FERNET_KEY` is used to encrypt server secret keys, and tokens while 'at rest' in the database, and decrypt them for use. `SECRET_KEY` is used to secure session, and CSRF protection cookies.
+You _must_ also provide three additional environment variables or secrets: `FLOWAUTH_FERNET_KEY`, `SECRET_KEY`, and `PRIVATE_JWT_SIGNING_KEY`. `FLOWAUTH_FERNET_KEY` is used to encrypt tokens while 'at rest' in the database, and decrypt them for use. `SECRET_KEY` is used to secure session, and CSRF protection cookies.
```
maxalbert:
```suggestion
You _must_ also provide three additional environment variables or secrets: `FLOWAUTH_FERNET_KEY`, `SECRET_KEY`, and `PRIVATE_JWT_SIGNING_KEY`. `FLOWAUTH_FERNET_KEY` is used to encrypt tokens while 'at rest' in the database, and decrypt them for use. `SECRET_KEY` is used to secure session and CSRF protection cookies.
```
(My understanding is the end of the sentence refers to "session [cookies] and CSRF protection cookies", so the comma after session is misleading?)

docs/source/install.md
```diff
@@ -152,57 +149,69 @@ FlowAuth will attempt to create all necessary tables when first accessed, but wi
...
-While `SECRET_KEY` can be any arbitrary string, `FLOWAUTH_FERNET_KEY` should be a valid Fernet key. A convenience command is provided to generate one - `flask get-fernet`.  
+You can use `openssl` to generate a private key:
 
+```bash
+openssl genrsa -out flowauth-private-key.key 4096
+```
+
+And then create a public key from the key file (`openssl rsa -pubout -in flowauth-private-key.key -out flowapi-public-key.pub`), or download it from FlowAuth once started. Should you need to supply the key using environment variables, rather than secrets (not recommended), you should base64 encode the key (e.g. `base64 -i flowauth-private-key.key`). FlowAuth and FlowAPI will automatically decode base64 encoded keys for use.
+
+By default, `SECRET_KEY` will be set to `secret`. You should supply this to ensure a secure system.
```
maxalbert:
```suggestion
By default, `SECRET_KEY` will be set to `secret`. You should supply a different value to ensure a secure system.
```

maxalbert: Is there a particular reason we have a default, rather than force the user to choose a value?

docs/source/install.md
maxalbert: I'm still a little confused around this though - it's a bit difficult to follow the various env vars and/or secrets being passed around via different names for the variables or secrets.

Also, it just occurred to me that the changelog mentions that `POSTGRES_PASSWORD_FILE` was renamed to `POSTGRE_PASSWORD`. So I assume something is broken somewhere?

flowauth/frontend/cypress/integration/public_key.js
```diff
@@ -0,0 +1,31 @@
...
+    );
+    // Log in and navigate to user details screen
+    cy.login_admin();
+    cy.goto("/");
+    cy.get("#public_key").click();
+    cy.wait(50);
+  });
+  it("Public key is shown", function() {
+    cy.get("#public_key_body").should(
+      "to.have.text",
```
maxalbert:
```suggestion
      "have.text",
```

docs/source/install.md
greenape: Ah it is - with the secrets in the Postgres image, the name of the secret is the name of a file, and you then set `POSTGRES_PASSWORD_FILE` to the path to that file. Iirc, we (I) had incorrectly interpreted this as 'you should call the secret `POSTGRES_PASSWORD_FILE`', and fixing that is what the changelog is alluding too.

flowauth/frontend/cypress/plugins/index.js
```diff
@@ -12,6 +12,9 @@
 // the project's config changing)
 
 module.exports = (on, config) => {
-  // `on` is used to hook into various events Cypress emits
-  // `config` is the resolved Cypress config
+  // Add in the public key from the parent environmen to cypress'
```
maxalbert:
```suggestion
  // Add in the public key from the parent environment to cypress'
```

docs/source/install.md
maxalbert: Ah, I forgot/missed that this is (i) a thing that is already present in the Postgres docker image, so it won't show up in _our_ source code, and (ii) is they add the `_FILE` suffix to the env var name via a generic function in their `docker-entrypoint.sh` script, so you can't even search for `POSTGRES_PASSWORD_FILE` in their docker setup. 😂

docs/source/install.md
greenape: Took this out - we don't actually have a default anymore!

flowkit_jwt_generator/flowkit_jwt_generator/fixtures.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+@pytest.fixture
+def access_token_builder(audience: Optional[str] = None) -> Callable:
+    """
+    Fixture which builds short-life access tokens.
+
+    Parameters
+    ----------
+    audience : str, default None
```
maxalbert: What's an example for the `audience` string? Also, all the usages I can find in the code seem to pass a dict instead of a string?

maxalbert: Never mind, the dict of course contains the claims passed to the `token_maker` closure below. Would still be useful to have an example of the `audience` string here.

flowkit_jwt_generator/flowkit_jwt_generator/fixtures.py
```diff
@@ -0,0 +1,120 @@
...
+        Optionally specify an audience for the token
+
+    Returns
+    -------
+    function
+        Function which returns a token encoding the specified claims.
+    """
+
+    if audience is None and "FLOWAPI_IDENTIFIER" in os.environ:
+        audience = os.environ["FLOWAPI_IDENTIFIER"]
```
maxalbert: Does this mean if audience is `None` and `FLOWAPI_IDENTIFIER` is not set, then `audience=None` will be passed to `generate_token()` below? Is this an issue at all?

flowkit_jwt_generator/flowkit_jwt_generator/jwt.py
```diff
@@ -273,32 +237,39 @@ def print_token(username, secret_key, lifetime, audience):
 
     For example:
 
+
+    generate-jwt --username TEST_USER --private-key $PRIVATE_JWT_SIGNING_KEY --lifetime 1 --audience TEST_SERVER all-access http://localhost:9090
```
maxalbert: This is missing the `--flowapi-url` option:
```suggestion
    generate-jwt --username TEST_USER --private-key $PRIVATE_JWT_SIGNING_KEY --lifetime 1 --audience TEST_SERVER all-access --flowapi-url http://localhost:9090
```

maxalbert: Strangely, the `--flowapi-url` doesn't show up though when running `generate-jwt --help`?

flowkit_jwt_generator/flowkit_jwt_generator/fixtures.py
greenape: Nah. Totally valid thing to do, and will just result in no audience key in the JWT - the resulting token wouldn't be usable for flowapi, but would by no means be invalid.

flowkit_jwt_generator/flowkit_jwt_generator/jwt.py
```diff
@@ -273,32 +237,39 @@ def print_token(username, secret_key, lifetime, audience):
...
+
+    Or,
+
+    generate-jwt --username TEST_USER --private-key $PRIVATE_JWT_SIGNING_KEY --lifetime 1 --audience TEST_SERVER query -a admin0 -a admin1 -p run -p get_result daily_location --query -a admin0 -p get_result flows
```
maxalbert: This example command also doesn't work (`Error: Missing option "--query-name").

flowkit_jwt_generator/flowkit_jwt_generator/fixtures.py
maxalbert: So what's a scenario where this may be useful?

flowkit_jwt_generator/flowkit_jwt_generator/jwt.py
greenape: Ah yes. Little convoluted that - it'll only show up in the help for the subcommand (`all-access`). Annoyingly, `click` won't let you get to that bit unless you've successfully completed all the required named args, so you need to do something like `enerate-jwt --username TEST_USER --private-key $PRIVATE_JWT_SIGNING_KEY --lifetime 1 --audience TEST_SERVER all-access --help` to see it.

PR #865, opened 2019-05-29
Fix the badges

greenape:
Because some having rounded edges and some not is driving me nuts


jch1g10:
The gitter badge still doesn't match the others

PR #896, opened 2019-06-04
Display year, seconds, tz when creating tokens

### LINKED ISSUE #260: Flowauth should display year (and also seconds and timezone?) in the token expiry date
**Describe the bug**
Flowauth does not display the year in the token expiry date, making it impossible to know when exactly a token will expire. Similarly, it would be helpful if it displayed token expiry up to second resolution, and also displayed the timezone (or made it clear that it refers to UTC if that's the case?)

**Product**
Flowauth

**Version**
Latest master (e03de7b0f0a3d)

**To Reproduce**

1. Follow the steps in #259 until step 6.
2. Notice that the token expiry date reads "January 1st 12:00 am".

**Expected behavior**

The token expiry date should read "1st January 2019, 12:00 am" or similar.

In fact, I'd prefer if the information was as comprehensive as possible so I'd like to see second-resolution and timezone information as well.

**Screenshots**

screen shot 2019-01-19 at 11 07 05

greenape:
Closes #260

### I have:

### Description

Switches the display of token expiry date time in the create token screen from:

Screenshot 2019-06-04 at 16 12 26

to:

Screenshot 2019-06-04 at 16 12 11

### DIFF
--- a/flowauth/frontend/src/TokenDetails.js
+++ b/flowauth/frontend/src/TokenDetails.js
@@ -25,12 +25,16 @@ const styles = theme => ({
 });
 
 class TokenDetails extends React.Component {
+  constructor(props) {
+    super(props);
+    this.nameRef = React.createRef();
+  }
   state = {
     nickName: {},
     rights: {},
     expiry: new Date(),
     latest_expiry: new Date(),
-    username_helper_text: ""
+    name_helper_text: ""
   };
 
   handleSubmit = () => {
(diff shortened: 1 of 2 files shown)


flowauth/frontend/src/TokenDetails.js
```diff
@@ -201,16 +212,22 @@ class TokenDetails extends React.Component {
...
           <MuiPickersUtilsProvider utils={DateFnsUtils}>
             <DateTimePicker
               value={expiry}
               onChange={this.handleDateChange}
               disablePast={true}
               maxDate={latest_expiry}
+              format="yyyy/MM/dd HH:mm:ss"
+              ampm={false}
+              margin="normal"
+              helperText={new Date().toTimeString().slice(9)}
```
maxalbert: Can we add a comment here that says this displays the timezone (which is not obvious otherwise)?

PR #903, opened 2019-06-06
Fix quickstart script to work with branches

### LINKED ISSUE #902: Quick-start script does not work with branches (when run locally)
When run locally (as opposed to CircleCI), the quick-start script does not work correctly with branches. The reason is that the `docker-compose.yml` file relies on the `CONTAINER_TAG` variable to be set to the correct value, but this only [happens](https://github.com/Flowminder/FlowKit/blob/master/quick_start.sh#L50) in CI. Also, the `BRANCH` and `CONTAINER_TAG` variables basically have the same meaning but locally one needs to set them independently for the script to work correctly.

As a related observation, when just reading the `quick-start.sh` script it is not obvious at all that `CONTAINER_TAG` is relevant. I nearly deleted it because I couldn't see it being used anywhere (and only when I ran the script did it become apparent that `docker-compose.yml` depends on it).

I suggest to use the same name for the variable in both `quick-start.sh` and `docker-compose.yml` so that only one of them need to be set (and I'd prefer something slightly more explicit such as `GIT_REV_OR_BRANCH` over `CONTAINER_TAG` because then it's obvious what the supported values are).

maxalbert:
Closes #902 

### I have:

### Description

- Renamed `CONTAINER_TAG` to `GIT_REVISION` in the docker-compose and docker-stack files.

  Rationale is that this makes it clear what the allowed values are and is more immediately intuitive for developers (otherwise, how do you find out the correct container tag?).

- Renamed `BRANCH` to `GIT_REVISION` in `quick_start.sh`. Same rationale as above, and it is good to use the same name in this script as in the docker-compose files so that we don't need to set an additional variable in `quick_start.sh` that is otherwise never used inside the script itself.

- Renamed `BRANCH` to `DOCS_BRANCH` in `docs/build.sh` and `docs/mkdocs.yml`, to distinguish it from the other variables above and make it clear that it is independent of them.

-

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -624,7 +624,7 @@ jobs:
       image: circleci/classic:201808-01
     working_directory: /home/circleci/project
     environment:
-      GIT_REVISION_OR_BRANCH: $CIRCLE_BRANCH
+      GIT_REVISION: $CIRCLE_SHA1
     steps:
       - checkout:
           path: /home/circleci/project/
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -624,7 +624,7 @@ jobs:
       image: circleci/classic:201808-01
     working_directory: /home/circleci/project
     environment:
-      BRANCH: $CIRCLE_BRANCH
+      GIT_REV_OR_BRANCH: $CIRCLE_BRANCH
     steps:
       - checkout:
           path: /home/circleci/project/
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -624,7 +624,7 @@ jobs:
       image: circleci/classic:201808-01
     working_directory: /home/circleci/project
     environment:
-      GIT_REV_OR_BRANCH: $CIRCLE_BRANCH
+      GIT_REVISION_OR_BRANCH: $CIRCLE_BRANCH
     steps:
       - checkout:
           path: /home/circleci/project/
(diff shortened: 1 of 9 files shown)


greenape:
I'm not immediately sold on this, I think because what it does is change the name of env var from describing what it does, to where it is derived from, which to me seems less informative? Like, there _happens_ to be a quite tight mapping from commit hash to docker tag, but that's almost coincidental, and doesn't hold for master vs latest.

maxalbert:
Yeah, I take your point and I was pondering that too.

The question is, when you need to set the value as a developer, what do you care about? For example, if you want to deploy a certain version of the system, where do you look up the information about which container tag to use?

I never go to docker cloud and search for a tag because it gives you no context about what that tag refers to. Instead, I look at the git history, find the revision I want to deploy and use _that_. Which is why I figured the name `GIT_REVISION` is far more practically useful, especially to someone who doesn't know how the docker container tags are derived. (And even for me who I'm familiar with our setup, every time I see the variable `CONTAINER_TAG` in the docker-compose file it takes me a second to remember that this _actually_ refers to the git revision).

Having said that, I could be convinced to revert the name back to `CONTAINER_TAG` but then I think we should clearly document its meaning in the docker-compose file and the install docs. Let me know if you prefer that.

maxalbert:
Closing this for now. Will open a separate PR which hopefully better addresses @greenape's comments.

Issue #390, opened 2019-02-15
Should be able to instantiate Query without a connection


maxalbert:
Currently the code in `Query.__init__()` raises an error if no valid connection to flowdb exists.

This has various disadvantages. For example, it makes it necessary to create connections in all tests that use a flowmachine query, even if they don't require running it. A more subtle downside is that it encourages to write tests in a more integrationy style (typically running against the database and checking the result, which slows down the test suite) rather than encouraging a design which allows for more unit-style tests.

It would be beneficial if Query objects could be instantiated without requiring an active connection to flowdb (e.g. by passing a dummy connection).

greenape:
Definitely. Historical reason for this is that in some cases, the query needs to check against FlowDB to know if there's any chance it can be successfully run (e.g. are there missing dates), or in a few cases, to require another query to be written as a table (e.g. `Random`). Obviously one can either pass in the connection to init for those cases, or add a secondary pre-flight checks method that does this.

greenape:
Closed via #1999

PR #928, opened 2019-06-12
 Add spatial unit classes

### LINKED ISSUE #926: Refactor spatial levels/aggregation
Many FlowMachine queries take a `level` parameter (e.g. `"versioned-site"` or `"admin3"`) to indicate the spatial locations/regions to which results should be joined or aggregated. In cases where a level has additional parameters (e.g. `size` for `"grid"` level), these must also be passed, resulting in a large number of parameters being passed around.

Additionally, there are many places (e.g. `JoinToLocation`, `GeoDataMixin`, `DistanceMatrix`, `subscriber_locations`, `TotalNetworkObjects`) where if/else statements are required for handling different levels in different ways.

We should define classes to represent spatial levels, to encapsulate the various parameters, and handle differences in logic using methods of these classes instead of if/else statements.

This is a sub-issue of #373

jch1g10:
Closes #926 

### I have:

### Description

The aim of this PR is to encapsulate all parameters and logic handling related to spatial levels / aggregation units in a set of new classes.

In `flowmachine/core/spatial_unit.py` I have defined the following new classes:
- `SpatialUnitMixin`: Provides a `location_id_columns` property (which replaces `get_columns_for_level`) and a `location_subset_clause` method which returns a SQL `WHERE` clause to subset a query to one or more locations (used by `FirstLocation`), along with methods for checking whether a spatial unit has certain criteria (e.g. represents a join to geographic data, or has longitude/latitude columns).
- `CellSpatialUnit`: Represents "cell" level (i.e. where no location join is required).
- `GeomSpatialUnit`: Inherits from `Query`. This is an abstract base class which represents a query that maps cell location IDs to geographic locations. Has a `get_geom_query` method which returns a SQL query to select the `location_id_columns` along with a `geom` column containing the geometry (this is used by `GeoDataMixin` and `DistanceMatrix`, among others).
- `LonLatSpatialUnit`: Derived from `GeomSpatialUnit` - represents the point-like levels ("lat-lon", "versioned-site", "versioned-cell").
- `PolygonSpatialUnit`: Derived from `GeomSpatialUnit` - represents the polygon levels ("polygon", "admin*", "grid").

`spatial_unit.py` also contains functions `versioned_cell_spatial_unit`, `versioned_site_spatial_unit`, `admin_spatial_unit`, `grid_spatial_unit` and `make_spatial_unit`, to help in creating spatial unit objects.

All classes which previously took a `level` parameter now accept a `spatial_unit` parameter instead, which should be a spatial unit object.

I have removed `cell_mappings.py`, as the `CellTo*` classes are superseded by `PolygonSpatialUnit`.

I've added a `Geography` class, which wraps the spatial unit `get_geom_query` method as a `Query` object with `GeoDataMixin`. This class is used by the FlowMachine server "get_geography" action.

`Grid` has moved from `flowmachine/features/spatial` to `flowmachine/core` so that it can be imported in `spatial_unit.py`.

Now that spatial units are themselves `Query` objects, these have replaced the `CustomQuery`/`CellTo*` objects which were created internally in `JoinToLocation`. I've also added a `location_joined_query` helper function which returns a `JoinToLocation` when a location join is required, or the original query object when no join is required (i.e. cell spatial unit), to avoid replicating this if/else logic in several other places.

I've removed the `_get_location_join` method of `GeoDataMixin`, which did a recursive search through the dependency tree to find a `JoinToLocation` object. Instead, any query which uses `GeoDataMixin` without redefining `_geo_augmented_query` must now have a `spatial_unit` attribute, whose `get_geom_query` method is used to add a "geom" column.

Previously, the `_SubscriberCells` class represented cell-level subscriber locations, and the `subscriber_locations` function returned either a `_SubscriberCells` or `JoinToLocation` object depending on the level. Now that I've added `location_joined_query` to handle this if/else case, I've combined `_SubscriberCells` and `subscriber_locations` into a `SubscriberLocations` class. Similarly, I've combined `_TotalCellEvents` into `TotalLocationEvents`.

I've moved the marshmallow `AggregationUnit` custom field out of `query_schemas/custom_fields.py` into `query_schemas/aggregation_unit.py`, and added a function `get_spatial_unit_obj` for getting a spatial unit object from an `aggregation_unit` string.

A couple of related thoughts which I think fall outside the scope of this PR:
- We should decide whether we want to extend `DistanceMatrix` to non-point locations. If so, this will hopefully be an easier task after this refactoring. If not, and we want `DistanceMatrix` to continue to support only versioned-cell and versioned-site spatial units, it may make more sense to calculate and store a distance matrix permanently in FlowDB.
- It would be useful in some cases to be able to create spatial unit objects without a DB connection. This links in with #390
- `Geography` is currently exposed as a query kind through the FlowAPI `run`/`poll`/`get` route, as well as `geography`. This simplifies adding permissions for the geography route, but doesn't make a lot of sense for users because the `get` route won't format the response as GeoJSON bere returning it.


flowmachine/flowmachine/core/join_to_location.py
```diff
@@ -275,3 +106,34 @@ def _make_query(self):
         """
 
         return sql
+
+
+def location_joined_query(left, *, spatial_unit, time_col="time"):
```
greenape: Can we type annotate the spatial_unit params?

flowmachine/flowmachine/core/spatial_unit.py
```diff
@@ -0,0 +1,820 @@
...
+        return isinstance(self, PolygonSpatialUnit)
+
+    def verify_criterion(self, criterion, negate=False):
+        """
+        Check whether this spatial unit meets a criterion, and raise an
+        InvalidSpatialUnitError if not.
+
+        Parameters
+        ----------
+        criterion : str
```
greenape: Make it an enum?

greenape:
Man, and I thought I had a tendency towards giant PRs ;)

I think the overall thrust of this is good (and it is a marathon effort), more once I've looked in detail.

flowmachine/flowmachine/core/server/query_schemas/aggregation_unit.py
```diff
@@ -0,0 +1,33 @@
...
+for getting the corresponding SpatialUnit object.
+"""
+
+from flowmachine.core import make_spatial_unit
+
+from marshmallow.fields import String
+from marshmallow.validate import OneOf
+
+
+class AggregationUnit(String):
```
greenape: (Maybe not now, but) this is a prime target for shifting to a JSON object which maps to the spatial objects.

jch1g10: Yes, I agree. I haven't attempted to implement this shift, partly because I've attempted to avoid changing any behaviour in this PR and partly because I think this change would be best as part of a wider consideration of how we want to expose aggregation units (and the associated permissions). But anticipation of this change was one of the reasons for moving this definition out of `custom_fields.py`.

greenape:
With the GeoDataMixin change - can we just straightforwardly mandate, e.g. abstract method or similar, that the `spatial_unit` needs to exist?

greenape:
Ok, I think what I'd like to do is get the conflicts resolved, then merge this and open a couple of issues for any post-hoc cleanup.

jch1g10:
> With the GeoDataMixin change - can we just straightforwardly mandate, e.g. abstract method or similar, that the `spatial_unit` needs to exist?

Hmm. The `spatial_unit` attribute is only needed in `_geo_augmented_query`, and there are several classes (e.g. `Grid`) which redefine this method, so `self.spatial_unit` is then not required. We could access `spatial_unit` in `_geo_augmented_query` through an abstract method (e.g. `_get_spatial_unit`), but then classes like `Grid` would also need to define this method but never use it.

I suppose we could make `_geo_augmented_query` an abstract method, and derive a new class (something like `GeoDataUsingSpatialUnitMixin`, but with a better name) which implements `_geo_augmented_query` and adds a new abstract method `_get_spatial_unit`?

PR #929, opened 2019-06-13
Minor flowetl tweaks

maxalbert:
Just a few tweaks from my trying out #858, mostly just renamings to make some things a little more obvious at a glance.

### I have:

### Description

- Renamed `TAG` to `FLOWETL_TESTS_CONTAINERS_TAG`
- Renamed the `tag` fixture to `container_tag`
- Added a structlog logger (which also fixes a minor issue with logging messages from tests not showing up before)
- Fixed a duplicate test name which shadowed one of the tests (the second one `test_find_files_non_default_filter`).
- Currently `os.getcwd()` is used to determine the `mounts` folder, which relies on the tests being run in the correct directory. I have introduced a fixture `flowetl_mounts_dir` to make things slightly more robust.
- Switched to non-standard ports for running the tests. @greenape has pointed out that it would be better to not assign any explicit ports but I'll leave this for a separate PR.
- Renamed fixture `data_dir` to `postgres_data_dir_for_tests`
- Added a separate fixture to pull docker images (mainly so that we can print some logging messages and the tests don't appear to be hanging)
- Added a fixture to ensure that the `AIRFLOW_HOME` and `TESTING` env vars are set when running the tests.
- Renamed `test_airflow_home_dir` to `airflow_home_dir_for_tests` because the former sounds like it's a pytest test _function_.
- Removed some code duplication in `wait_for_completion `

### DIFF
--- a/flowetl/tests/integration/conftest.py
+++ b/flowetl/tests/integration/conftest.py
@@ -307,9 +307,9 @@ def airflow_local_setup():
     }
     env = {**os.environ, **extra_env}
     # make test Airflow home dir
-    test_airflow_home_dir = os.environ["AIRFLOW_HOME"]
-    if not os.path.exists(test_airflow_home_dir):
-        os.makedirs(test_airflow_home_dir)
+    airflow_home_dir_for_tests = os.environ["AIRFLOW_HOME"]
+    if not os.path.exists(airflow_home_dir_for_tests):
+        os.makedirs(airflow_home_dir_for_tests)
 
     initdb = Popen(
         ["airflow", "initdb"], shell=False, stdout=DEVNULL, stderr=DEVNULL, env=env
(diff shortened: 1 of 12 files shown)


flowetl/tests/integration/conftest.py
```diff
@@ -18,34 +19,41 @@
...
 from docker.types import Mount
 from shutil import rmtree
 
+here = os.path.dirname(os.path.abspath(__file__))
+logger = logging.getLogger("flowetl-tests")
+
+
+@pytest.fixture(scope="session")
+def flowetl_mounts_dir():
+    return os.path.abspath(os.path.join(here, "..", "..", "mounts"))
```
greenape: Just a note for the future that it would be preferable if this stuff used tmpdirs instead of relying on empty directories we've tricked git into keeping around.

PR #930, opened 2019-06-13
FlowETL CI tweaks

greenape:
Slight simplification of the flowetl test config by setting the working directory to flowetl, meaning you don't need to cd for every command, and adds a save cache step so the restore cache step doesn't feel lonely.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,7 +42,6 @@ defaults:
       tags:
         only: /.*/
 
-
 executors:
   python_with_flowdb: # Synthetic data generation execution environment
     parameters:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -901,28 +898,26 @@ workflows:
...
           <<: *run_always_org_context
       - retag_images:
           name: retag_master_branch
-          requires:
-            &retag_requirements
+          requires: &retag_requirements
             - test_quickstart
             - test_quickstart_examples_with_smaller_data
-#            - test_quickstart_with_larger_data
+            #            - test_quickstart_with_larger_data
```
maxalbert:
```suggestion
            # - test_quickstart_with_larger_data
```

PR #931, opened 2019-06-13
Stop pretty-quick getting carried away and autoformatting everything

greenape:
Restrict Flowauth's code formatting pre-commit hook to auto formatting javascript.

### DIFF
--- a/flowauth/frontend/package.json
+++ b/flowauth/frontend/package.json
@@ -38,7 +38,7 @@
   ],
   "husky": {
     "hooks": {
-      "pre-commit": "pretty-quick --staged"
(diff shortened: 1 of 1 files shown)


flowauth/frontend/package.json
```diff
@@ -38,7 +38,7 @@
   ],
   "husky": {
     "hooks": {
-      "pre-commit": "pretty-quick --staged"
+      "pre-commit": "pretty-quick --staged --pattern \"**/*.*(js|jsx|json)\""
```
maxalbert: Purely out of curiositoy, why the last `*` after the dot and before the `js/jsx/json` suffix?

PR #932, opened 2019-06-13
Add flowetl's etl module to codecov

greenape:
Includes FlowETL's ETL module in the codecov reports.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -590,9 +590,12 @@ jobs:
       - run:
           name: run etl module unit tests
           command: |
-            pipenv run pytest --junit-xml=test_results/pytest/results_unit_etl_module.xml ./etl/tests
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -590,9 +590,12 @@ jobs:
       - run:
           name: run etl module unit tests
           command: |
-            pipenv run pytest --junit-xml=test_results/pytest/results_unit_etl_module.xml ./etl/tests
+            pipenv run pytest --junit-xml=test_results/pytest/results_unit_etl_module.xml --cov=./etl/etl \
+            --cov-report xml \
+            --cov-report term ./etl/tests
```
maxalbert: Do we want `term-missing` instead of `term`?

greenape:
```suggestion
            --cov-report term-missing ./etl/tests
```

PR #933, opened 2019-06-13
Remove MOUNT_HOME env var

maxalbert:
### I have:

### Description

This PR removes the `MOUNT_HOME` env var from flowetl. The only reason I can see we might want this is if it needs to be configured to a different value during testing, but AFAICT the variable is only relevant for paths inside the docker container where the value is hard-coded to `/mounts/` anyway.


maxalbert:
(For the record, the main reason I see for removing this is to avoid the user having to set it as a mandatory env var during deployment when it isn't relevant to the actual host file system.)

PR #953, opened 2019-06-18
Remove pygraphviz as dev dependency (only load on demand)

### LINKED ISSUE #952: Pipenv install error due to pygraphviz if graphviz is missing
If `graphviz` is not installed on the system then running `pipenv install --dev` in the `flowmachine` folder will fail because the installation of `pygraphviz` (which is a dev dependency of `flowmachine`) will fail.

Since `graphviz` needs to be installed on the system the best we can do is to raise a meaningful error message. Given that `pygraphviz` is only needed in `plot_dependency_graph()` in `flowmachine/utils.py`, I'm tempted to remove `pygraphviz` as a dependency from the Pipfile and raise a meaningful error when the user tries to call that function.

maxalbert:
Closes #952 

### I have:

### Description

As suggested in #952, this PR removed `pygraphviz` as a dev dependency and raises a meaningful error (which also mentions the `graphviz` system package) if it isn't installed but needed to run `plot_dependency_graph()`. Happy to go with a different solution if there is a better way.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -294,15 +294,15 @@ jobs:
     steps:
       - checkout
       - restore_cache:
-          key: flowmachine-deps-1-{{ checksum "flowmachine/Pipfile.lock" }}
+          key: flowmachine-deps-2-{{ checksum "flowmachine/Pipfile.lock" }}
       # Need to install graphviz and pygraphviz manually because it was removed from the Pipfile
       # (see https://github.com/Flowminder/FlowKit/issues/952)
+      - run: cd flowmachine && sudo apt-get install -y graphviz && pipenv run pip install pygraphviz
       - run: cd flowmachine && pipenv install --dev --deploy && pipenv run pip install -e .
       - save_cache:
-          key: flowmachine-deps-1-{{ checksum "flowmachine/Pipfile.lock" }}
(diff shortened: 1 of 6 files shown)


.circleci/config.yml
```diff
@@ -293,8 +293,11 @@ jobs:
     working_directory: /home/circleci/project
     steps:
       - checkout
+      - run: cd flowmachine && sudo apt-get install -y graphviz && pipenv run pip install pygraphviz
       - restore_cache:
           key: flowmachine-deps-1-{{ checksum "flowmachine/Pipfile.lock" }}
+      # Need to install graphviz and pygraphviz manually because it was removed from the Pipfile
+      # (see https://github.com/Flowminder/FlowKit/issues/952)
       - run: cd flowmachine && pipenv install --dev --deploy && pipenv run pip install -e .
```
greenape: I think this is why the test is failing - there's already a cache for this iteration of the lock file, so the install of pygraphviz isn't getting cached.

maxalbert: Sure, that's why I moved the installation command outside of the save_cache/restore_cache bracket. But I clearly don't understand how CircleCI caching works. ;) Should I have moved it _after_ the save_cache command?

greenape: You could bump the deps version, or wait until #954 is merged which will change the lock file so the cache gets made again

PR #955, opened 2019-06-19
Fix token ui tests

greenape:
Wait for permissions to be loaded before finishing processing click events on top level checkboxes - fixes sporadic fail on CI.

(Also fixes permissions boxes not showing up when first adding a new server)

### DIFF
--- a/flowauth/frontend/cypress/integration/add_new_server.js
+++ b/flowauth/frontend/cypress/integration/add_new_server.js
@@ -14,37 +14,38 @@ describe("Server management", function() {
   it("Add server name with space", function() {
     cy.get("#new").click();
     // adding username with space
-    cy.get("#name").type("Server ");
+    cy.get("#name").type("Server ", { force: true });
     //checking validation text
     cy.get("#name-helper-text").should(
       "have.text",
       "Server name may only contain letters, numbers and underscores."
     );
     cy.get("#name")
-      .type(" ")
-      .clear();
-    cy.get("#name").type("SERVER_TEST01");
+      .type(" ", { force: true })
+      .clear({ force: true });
+    cy.get("#name").type("SERVER_TEST01", { force: true });
     cy.contains("#name-helper-text").should("not.exist");
   });
   it("Add blank server name", function() {
     cy.get("#new").click();
     //adding blank username
     cy.get("#name")
-      .type(" ")
-      .clear();
+      .type(" ", { force: true })
+      .clear({ force: true });
     //checking validation text
     cy.get("#name-helper-text").should(
       "have.text",
       "Server name can not be blank."
     );
-    cy.get("#name").type("SERVER_TEST01");
+    cy.get("#name").type("SERVER_TEST01", { force: true });
     cy.contains("#name-helper-text").should("not.exist");
   });
   it("Add server name more than 120 characters", function() {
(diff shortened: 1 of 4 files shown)


flowauth/frontend/cypress/integration/add_new_server.js
```diff
@@ -14,95 +14,96 @@ describe("Server management", function() {
   it("Add server name with space", function() {
     cy.get("#new").click();
     // adding username with space
-    cy.get("#name").type("Server ");
+    cy.get("#name").type("Server ", { force: true });
```
maxalbert: What's the purpose of `force: true` here (and below)?

flowauth/frontend/cypress/integration/add_new_server.js
```diff
@@ -14,95 +14,96 @@ describe("Server management", function() {
...
-    cy.get("#name").type("SERVER_TEST01");
+    cy.get("#name").type("SERVER_TEST01", { force: true });
     cy.contains("#name-helper-text").should("not.exist");
   });
   it("Add server name more than 120 characters", function() {
     cy.get("#new").click();
     //adding username
     cy.get("#name").type(
-      "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa"
+      "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
```
maxalbert: In a recent PR you had a nice concise way of writing this (something like `"a".repeat(121)` or similar). Do you want to change it here too?

flowauth/frontend/cypress/integration/add_new_server.js
greenape: Works around cypress trying to scroll the thing into view, but it ending up hidden because their scrolling algorithm is bad. This types (or whatever) into the box even if it isn't visible - I'd add a comment to this effect, but we end up having to do it pretty much universally across the tests, so is already mentioned elsewhere.

flowauth/frontend/cypress/integration/add_new_server.js
greenape: Ahh yes, in #864 
Yeah, might as well throw that in here as well.

flowauth/frontend/cypress/integration/add_new_server.js
maxalbert: Yeah I was going to say the same thing. Would be nice to have a comment but since it's everywhere it's already fairly visible (just not entirely clear why if you don't see that comment). A bit annoying that we have to do this everywhere but seems to be necessary. Out of interest, is this a known bug that they are aware of and that might be fixed some time in the future?

greenape: Yeah, see https://github.com/cypress-io/cypress/issues/871

They are apparently working on it (or were in December last year).

PR #956, opened 2019-06-20
MFA for FlowAuth

### LINKED ISSUE #121: Support 2factor auth in FlowAuth
Because FlowAuth is potentially much more publicly deployed than other components, and is the gatekeeper for access to the rest of any given FlowKit deployment, it should of course be as secure as possible.

Supporting 2factor auth would be a clear boost here.

greenape:
Closes #121

### I have:

### Description

This adds optional MFA support to FlowAuth.

MFA can be enabled in two ways: by the user, or by the admin.

If an admin enables it for a user, the user will be taken through a mandatory MFA setup when they next log in. They will first be shown 16 backup codes, each of which grants a one time access if their authenticator device is unavailable. They must either copy, or download the backup codes to proceed.
They will next be shown a QR code, which can be scanned by an authenticator app (e.g. Google Authenticator) to set up 2fa. They must enter a code generated by the app to proceed.

After running through that, they are logged in and 2fa is enabled - the next time they log in, they will be required to enter a 2fa code in addition to their username and password. mfa codes are single use - if the user successfully logs in, then immediately logs out, the code they just used is not valid and they will need to wait for a new one.

Users can optionally enable mfa themselves, from the user settings page. The process is identical, but can be cancelled at any time.

If an admin has required that a user have mfa, the user cannot disable it. An admin _can_ disable it for a user, but they'll be required to set it up again through the login flow at next login.

A user may re-set up mfa at any time - their existing mfa config will not be replaced unless they complete the process.

A user may regenerate a new set of backup codes at any time, this obviously renders the previous backups invalid.


jch1g10:
One minor point: If a user decides to enable 2fa themselves, then once they complete the setup process they will be taken back to the password-change screen with the "enable" button still present (and the other two greyed-out), which gives the impression the setup was unsuccessful. Simply refreshing the page will change the "enable" button to "disable", so it's not a major problem.

hprobotic:
hi @greenape im trying to use your mfa code in my personal project. But i got issue with `@hybrid_property`. `key.decode()` function 
```python
    @hybrid_property
    def secret_key(self):
        import pdb; pdb.set_trace()
        key = self._secret_key
        # key = 'gAAAAABdWAGGpY4vU1yiHmXwZxxVklymH5W30AoFpGdThuFf-33PxCAvKJ7hyljOxye-y2cqk4iTmb4ZzHXNf5LhyqYuf4irnjAB_CP9W4t5ftZ1Oj37nOA='
        try:
            key.decode()
            # *** AttributeError: 'str' object has no attribute 'decode'
        except AttributeError:
            return key
            # return 'gAAAAABdWAGGpY4vU1yiHmXwZxxVklymH5W30AoFpGdThuFf-33PxCAvKJ7hyljOxye-y2cqk4iTmb4ZzHXNf5LhyqYuf4irnjAB_CP9W4t5ftZ1Oj37nOA='
        return get_fernet().decrypt(key).decode()
```

and in another place

```python
    def validate(self, code: str) -> bool:
        current_app.logger.debug(f"Verifying {code} using {self.secret_key}")
        is_valid = pyotp.totp.TOTP(self.secret_key).verify(code)
        # -> Error: binascii.Error: Non-base32 digit found
        if is_valid:
            if (
                self.last_used_two_factor_code == code
            ):  # Reject if the code is being reused
                raise Unauthorized("Code not valid.")
            else:
                self.last_used_two_factor_code = code
            db.session.add(self)
            db.session.commit()
            return True
        else:
            raise Unauthorized("Code not valid.")
```

PR #977, opened 2019-06-25
Fix versions of python modules not being set in docker images

### LINKED ISSUE #818: Version 0+unknown
The current docker build process leads to all installed libs in the docker containers being version 0+unknown. This is _largely_ a non issue, because the container itself is versioned, but does mean that the spec generated by flowapi will have that as the version as well.

Presumably the best solution is to have build scripts, which build a wheel for the module and copy that rather than the source files as we do currently.

greenape:
Closes #818

### I have:

### Description

Uses versioneer's ability to infer a version number from a source tree to set the version numbers for python modules inside docker containers.


Dockerfile
```diff
@@ -9,11 +9,19 @@
 FROM jupyter/scipy-notebook
 
 RUN rm -rf /home/$NB_USER/work
-
+ARG SOURCE_VERSION=0+uknown
```
maxalbert:
```suggestion
ARG SOURCE_VERSION=0+unknown
```

.circleci/config.yml
```diff
@@ -547,15 +547,17 @@ jobs:
           path: /home/circleci/project/
       - attach_workspace:
           at: /home/circleci
-      - setup_remote_docker
+      - setup_remote_docker:
+          version: 18.09.3
```
maxalbert: Out of interest, why do we need to pin the version here?

greenape: Uses an ancient version if we don't - specifically, we need a newer engine so we can use env vars in combination with `COPY`

flowapi/flowapi/_version.py
```diff
@@ -382,7 +382,7 @@ def render_pep440_post(pieces):
 
     The ".dev0" means dirty. Note that .dev0 sorts backwards
     (a dirty tree will appear "older" than the corresponding clean one),
-    but you shouldn't be releasing software with -dirty anyway.
+    but you shouldn't be releasing software with -dirty anyways.
```
maxalbert: 😄

flowauth/Dockerfile
```diff
@@ -24,6 +24,11 @@ ENV FLASK_APP flowauth
...
+ARG SOURCE_VERSION=0+unknown
+ENV SOURCE_VERSION=${SOURCE_VERSION}
+ENV SOURCE_TREE=FlowKit-${SOURCE_VERSION}
+WORKDIR /${SOURCE_TREE}/flowauth
 COPY backend .
-RUN pip install --no-cache-dir .
\ No newline at end of file
+RUN python setup.py bdist_wheel && pip install dist/*.whl
+WORKDIR /app
+COPY backend/uwsgi.ini .
```
maxalbert: What exactly is the purpose of this `uwsgi.ini` file? Should this be included in the distribution instead (e.g. via `MANIFEST.in`)?

greenape: It configures uwsgi, which is serving the app in combination with nginx. So is deployment-side rather than distribution side.

PR #978, opened 2019-06-25

### LINKED ISSUE #947: Use execution_date in dags

danwilliams:
Closes #947 

### I have:

### Description

This updates the execution_date to use the one parsed from the file instead of always using now, and replacing the tasks that used PythonOperator with the custom render_and_run_sql__callable (now deleted) to use the PostgresOperator instead.


flowetl/etl/etl/etl_utils.py
```diff
@@ -107,42 +111,25 @@ def construct_etl_dag(
...
+        dag_id=f"etl_{cdr_type}",
+        schedule_interval=None,
+        default_args=default_args,
+        template_searchpath=config_path,
     ) as dag:
 
-        init = PythonOperator(
-            task_id="init", python_callable=init, provide_context=True
+        init = init(task_id="init")
+        extract = extract(task_id="extract", sql=f"/etl/{cdr_type}/extract.sql")
```
maxalbert: Is the `sql` path relative to the `template_searchpath`/`config_path` defined above? I.e. does this path refer to `/mounts/config/etl/{cdr_type}/extract.sql`?

If so, is it possible to remove the leading slash to avoid it being interpreted as an absolute path (or does Airflow require the leading slash)?
```suggestion
        extract = extract(task_id="extract", sql=f"etl/{cdr_type}/extract.sql")
```

flowetl/etl/tests/test_construct_etl_dag.py
```diff
@@ -35,8 +39,18 @@ def test_construct_etl_dag_with_test_callables():
 
     assert dag.dag_id == f"etl_{cdr_type}"
 
-    dag_task_callable_mapping = {t.task_id: t.python_callable for t in dag.tasks}
-    assert dag_task_callable_mapping == TEST_ETL_TASK_CALLABLES
+    dag_task_callable_mapping = {
+        t.task_id: t.python_callable
+        for t in dag.tasks
+        if not isinstance(t, PostgresOperator)
```
maxalbert: Why is this check needed? Won't all tasks used in the test run be instances of `PythonOperator`?

(If we do need to filter them, might be useful to be explicit that we require them to be python operators: `if isinstance(t, PythonOperator)`?)

flowetl/etl/tests/test_construct_etl_dag.py
```diff
@@ -35,8 +39,18 @@ def test_construct_etl_dag_with_test_callables():
...
+    }
+    expected_dag_task_callable_mapping = {
+        t.task_id: TEST_ETL_TASK_CALLABLES[t.task_id](task_id=t.task_id).python_callable
+        for t in dag.tasks
+        if not isinstance(t, PostgresOperator)
+    }
+    assert dag_task_callable_mapping == expected_dag_task_callable_mapping
+    [t.assert_called_once() for _, t in task_callable_mapping.items()]
```
maxalbert: Nice! Good idea to check that the callables are actually called. 👍

flowetl/etl/tests/test_construct_etl_dag.py
```diff
@@ -55,8 +71,20 @@ def test_construct_etl_dag_with_production_callables():
 
     assert dag.dag_id == f"etl_{cdr_type}"
 
-    dag_task_callable_mapping = {t.task_id: t.python_callable for t in dag.tasks}
-    assert dag_task_callable_mapping == PRODUCTION_ETL_TASK_CALLABLES
+    dag_task_callable_mapping = {
+        t.task_id: t.python_callable
+        for t in dag.tasks
+        if not isinstance(t, PostgresOperator)
```
maxalbert: See above:
```suggestion
        if not instance(t, PythonOperator)
```

And similarly in L84 below. (Need to add `from airflow.operators.python_operator import PythonOperator` in the import section, too.)

flowetl/mounts/config/etl/calls/extract.sql
```diff
@@ -1,2 +1,2 @@
-DROP TABLE IF EXISTS {{ extract_table }};
-CREATE TABLE {{ extract_table }} (LIKE events.{{cdr_type.name.lower()}} INCLUDING ALL)
+DROP TABLE IF EXISTS {{ dag_run.conf.extract_table }};
```
maxalbert: Out of interest, is the purpose of adding the `dag_run.conf` prefix to make the code more explicit / readable?

greenape: I think I'd prefer using the [built in macros](https://airflow.apache.org/macros.html) or at a push, user defined macros. Most of the names are (or should be) variations on the type of thing being etl'd combined with the date, no?

flowetl/etl/etl/etl_utils.py
danwilliams: Yup, it's relative to the `template_searchpath` from above, added a short comment to reflect this. The leading slash wasn't needed indeed, good point - removed from the paths.

flowetl/etl/tests/test_construct_etl_dag.py
danwilliams: They will all be `PythonOperator` in this specific scenario, but the one below will have the "production" ones; these tests check that they are properly set for a DAG - this was needed for the second one so it seemed consistent to have both of them like that. Although, they will be `PythonOperator` or `BranchPythonOperator`, but it makes sense to make it inclusive rather than exclusive. Updated the checks to check for instances of `PythonOperator` and `BranchPythonOperator`.

flowetl/mounts/config/etl/calls/extract.sql
danwilliams: The `dag_run.conf` prefix was needed to get the vars from the template config - will look into using the macros, the extract and transform tables have an uuid (the same one) appended to them, but this _should_ be possible using `macros.uuid` - there might be some things need adding for the `clean` task to then work correctly.

danwilliams:
> Strangely, the latest CircleCI run for this branch failed (`TimeoutError: DAG 'etl_mds' did not reach desired state success.`), but locally the tests pass for me so I hope it was a spurious error. I just re-started the build, let's see whether the tests pass this time.

Weird, it looks like the timeout might not have been enough, or as you said, a spurious error?

flowetl/mounts/config/etl/calls/extract.sql
maxalbert: Hmm, I'm in two minds about whether I agree with this or not. Are you suggesting to replace e.g. `{{ extract_table }}` with (some variation of) `etl.x{{ uuid_sans_underscore }}`? I find the former _much_ more meaningful and easy to understand at a glance.

Granted, if you need to debug something you need to look up the definition of `extract_table`, but 99% of the time you don't care about what the extract table is called but you _do_ want to know that you're dealing with the `extract_table` rather than, say, `transform_table`, no? Aren't `{{ extract_table }}` and `{{ transform_table }}` much more readable than `etl.x{{ uuid_sans_underscore }}` and `etl.t{{ uuid_sans_underscore }}`?

greenape: Finding an alternative to the uuid suffix would also seem fine to me tbh.

greenape: I see your point there @maxalbert - I was thinking something like `etl.{{ cdr_type }}_{{ ds_nodash }}`, but I'd also be happy with `extract_table` being a custom macro that does that or similar.

I just really dislike stuff being hidden in the `conf` ;)

maxalbert: Sure, but no matter how you determine the names of these tables, as a reader of the code (and/or someone who needs to drop in their own ETL snippets), wouldn't you rather know what their _purpose_ is rather than what they're called? I feel it creates less mental overhead because you don't need to worry about coming up with a good name / temporary table location yourself and can just use the one that flowetl has already helpfully provided you with. Or do you feel differently?

maxalbert: (Sorry, that was a response to your previous comment - somehow Github doesn't refresh the page while I'm writing a reply. 😂)

And yes, I take your point about stuff being hidden in the conf, although I don't mind too much because at least it makes it obvious where to look if you need to look up the value. (Incidentally, needing to use the conf to look up those values is yet another reason for me to not like Airflow too much. ;))

flowetl/etl/tests/test_construct_etl_dag.py
maxalbert: I just realised that `BranchPythonOperator` is actually a subclass of `PythonOperator` so we only need a check for `isinstance(..., PythonOperator)`.

maxalbert:
Looks like `flowetl/etl/etl_utils.py` needs reformatting with black (the addition of the comment made the lint job [fail](https://circleci.com/gh/Flowminder/FlowKit/72365)).

(If you [set up](https://flowkit.xyz/developer/dev_environment_setup/#pre-commit-hook-for-python-code-formatting-with-black) the pre-commit hook git should run this automatically for you when you commit.)

flowetl/etl/etl/etl_utils.py
```diff
@@ -22,6 +22,7 @@
 from airflow import DAG
 from airflow.hooks.postgres_hook import PostgresHook
 from airflow.operators.python_operator import BranchPythonOperator, PythonOperator
+from airflow.operators.postgres_operator import PostgresOperator
```
maxalbert: This import of `PostgresOperator` can now be removed since we're not using it any more (and similarly for `BranchPythonOperator` once we've removed the associated isinstance check below).

maxalbert:
Thanks for the changes. Looks good to me now. 👍

I tweaked a couple of very minor things (see inline comments), one of which (the failing lint job) prevented the merge. Happy to merge this now when the tests pass. :)

flowetl/mounts/config/etl/calls/extract.sql
maxalbert: I suggest to leave this out of the current PR and _if_ we want to change something around this address it in a separate PR.

maxalbert:
(Looks like we'll need to merge #990 first to make the tests pass.)

flowetl/etl/tests/test_construct_etl_dag.py
danwilliams: Because the user-defined macros don't render as templates, it wasn't as simple to write one as `"extract_table": "etl.x{{ cdr_type }}_{{ ds_nodash }}"`. Instead, they have been made into lambdas and pass the `ds_nodash` through the templates (`"extract_table": lambda execution_time: f"etl.x{ cdr_type }_{ execution_time }",` and `{{ extract_table(ds_nodash) }}`).

What do you think of that? It can be either that, or making the extract_table as an extract_table_prefix instead and using `{{ extract_table_prefix }}_{{ ds_nodash }}` in the templates, OR passing the entire dag_run instead of ds_nodash and using dag_run.conf inside the lambda.

PR #990, opened 2019-06-27
Bump pinned marshmallow version, unpin marshmallow-oneofschema

greenape:
+ subsumes #987 #986 #982


docs/Pipfile
```diff
@@ -19,8 +19,7 @@ numpydoc = "*"
 black = "==19.3b0"
 click = "*"
 flowmachine = {editable = true,path = "./../flowmachine"}
-marshmallow = {editable = true,ref = "3.0.0rc6",git = "https://github.com/marshmallow-code/marshmallow.git"} # Temporarily required while there's a marshmallow versions conflict
-marshmallow-oneofschema = {editable = true,ref = "8195dc3e595fa33a5f007f10ba7cfc08f2177515",git = "https://github.com/marshmallow-code/marshmallow-oneofschema.git"} # Temporarily required while there's a marshmallow versions conflict
```
maxalbert: Why can we _remove_ `marshmallow-oneofschema` here (rather than simply unpin it)? I assume if the tests pass it's fine, just wondering.

greenape: Because there's now a pypi release that works with marshmallow 3, it'll get properly resolved from the Flowmachine setup.py.

Issue #640, opened 2019-04-17
Schema and data model redesign


greenape:
We should revisit the schema of FlowDB, and the data model of FlowKit in general.

This is a long standing desire - the schema as it exists is nigh nonexistent, and we strongly suspect that a much more efficient approach is possible. The data model is ad hoc, implicit, and kind of 'on demand'. That's partly because of the evolved nature of FlowKit, and partly as a result of the origins of the system and a need to be very careful with disk space and the like.

We're now in a position where we can look at alternative approaches to both, and should take advantage of that.

greenape:
This subsumes #181.

greenape:
Closed via #1500

PR #993, opened 2019-06-27
Add auxiliary etl query table

### LINKED ISSUE #988: Add auxiliary table to record results of regular queries (e.g. run as part of daily ingestion)
As part of the daily ingestion process one typically wants to run some initial quality checks or calculate informative metrics. Simple examples: how many call events happened on that date? Are there any `cell_id` values in this date's CDR data which are not present in the `infrastructure.cells` table (and if so how many)? Etc.

In the longer run this will be supported by a set of supplementary tables (to be added as part of #640) that record the results of standard queries (or standard pre-aggregations) which are run as part of the ingestion process.

In the immediate short term it will be useful to have a single auxiliary table where the results of semi-ad-hoc queries can be recorded in a slightly less structured and more informal way but which still allows quick querying of results. Suggestion for the structure of this table:
```
|   cdr_date | cdr_type | type_of_query_or_check |  outcome | optional_comment_or_description |
|------------+----------+------------------------+----------+---------------------------------|
| 2019-01-01 | calls    | num_total_calls        | 42000000 | NULL                            |
| 2019-01-01 | calls    | num_missing_cell_ids   |     4242 | needs investigation             |
```

Such a table should be added to flowdb, e.g. as `etl.daily_queries` (any suggestions for a better name?).

maxalbert:
Closes #988 

### I have:

### Description

This PR adds a table `etl.post_etl_queries` to record the outcome of queries run as part of the regular ETL process. I'm happy to tweak the name and structure of the table if we think something else is more appropriate, suggestions are welcome.

For the time being I have kept the data types for all columns (except `cdr_date`) as `TEXT`. Going forward one could imagine using enums for `cdr_type` and in particular for `type_of_query_or_check` (so that we enforce consistency in this column, which would help with querying the table). In the future we may also be more specific with the data type of the `outcome` column (e.g. use int/float/boolean/text depending on the actual query it records) but until the details become clearer and we split this auxiliary table into separate ones (probably as part of #640) this seemed to be the most generically useful approach.


flowdb/sql/04_schema_other.sql
```diff
@@ -78,6 +78,18 @@ CREATE TABLE etl.etl_records (
 	PRIMARY KEY (id)
 );
 
+CREATE TABLE etl.post_etl_queries (
+    cdr_date DATE,
```
greenape: Can we record the date of the check as well?

maxalbert: Good point. 👍

maxalbert: I've made it a `TIMESTAMPTZ` rather than a `DATE` column because there is no harm in having the more precise info about when the check was run.

PR #996, opened 2019-06-28

### LINKED ISSUE #947: Use execution_date in dags

danwilliams:
Closes #947 

### I have:

### Description

Because the user-defined macros don't render as templates, it wasn't as simple to write one as "extract_table": "etl.x{{ cdr_type }}_{{ ds_nodash }}". Instead, they have been made into lambdas and pass the ds_nodash through the templates ("extract_table": lambda execution_time: f"etl.x{ cdr_type }_{ execution_time }", and {{ extract_table(ds_nodash) }}).


maxalbert:
I like this approach! It's a nice compromise between having a meaningful macro name and making use of the built-in macro `ds_nodash` so that it's obvious which date is being operated on. :)

I'll add a couple of very minor inline comments, and then this should be ready to merge.

There is a small merge conflict with master at the moment which needs to be resolved (apologies, I think this is caused by my already merging #978 last night). The CI build is currently failing but this is due to an upgrade in dependencies (see #990) and should be resolved by merging master into this branch.

flowetl/etl/etl/etl_utils.py
```diff
@@ -114,7 +114,13 @@ def construct_etl_dag(
         dag_id=f"etl_{cdr_type}",
         schedule_interval=None,
         default_args=default_args,
-        template_searchpath=config_path, # template paths will be relative to this
+        template_searchpath=config_path,  # template paths will be relative to this
+        user_defined_macros={
+            "extract_table": lambda execution_time: f"etl.x{ cdr_type }_{ execution_time }",
```
maxalbert: I think I'd prefer if we had a verb in the lambdas, e.g. `get_extract_table`, `get_transform_table`, `get_load_table` (or similar). This makes the code where they are being used slightly more intuitive to read and understand imho.

Also, can we rename the argument from `execution_time` to `execution_date` to be consistent with both Airflow's terminology and the value that's actually parsed to them?

maxalbert:
It just occurred to me that previously the table names contained a uuid (which guaranteed to make them unique) and we now rely solely on the `cdr_date` for the table name. Is there a risk that this will lead to conflicts if a second run of the same ingestion DAG is started while the first one is still running? 

In most cases this should not be a problem because once the `init` step has been run the data for the date in question is markes as "being ingested" which should prevent the second run from getting past the `init` step (I hope?).

However, there may be a small time window before the first `init` step has been completed where a race condition might occur? For robustness, would be good to either convince ourselves that this is not an issue, or to ensure that the table names are truly unique for each DAG run (in which case we should also make sure that this doesn't lead to a conflict in a later step of the ETL process).

danwilliams:
> It just occurred to me that previously the table names contained a uuid (which guaranteed to make them unique) and we now rely solely on the cdr_date for the table name. Is there a risk that this will lead to conflicts if a second run of the same ingestion DAG is started while the first one is still running?

Hmm, that's a good point - there may be issues if two DAGs are running in the same time for files with the same cdr date (continuing discussion below in the last paragraph).

> In most cases this should not be a problem because once the init step has been run the data for the date in question is markes as "being ingested" which should prevent the second run from getting past the init step (I hope?).

That's correct, the pair (file, cdr_date) won't be ingested again.

> However, there may be a small time window before the first init step has been completed where a race condition might occur? For robustness, would be good to either convince ourselves that this is not an issue, or to ensure that the table names are truly unique for each DAG run (in which case we should also make sure that this doesn't lead to a conflict in a later step of the ETL process).

Following from the above - since the execution_date is not `now` anymore but the date parsed from the file, there may still be issues if running more than one DAG in the same time for a file that starts on the same date... An alternative would be to use `ts_nodash` instead of `ds_nodash` - but if the files would have the same cdr_type - we could still run into issues. If we expect to parse files that have the same timestamp in parallel, maybe having an uuid in the conf would still be the best alternative for us.

maxalbert:
> If we expect to parse files that have the same timestamp in parallel, maybe having an uuid in the conf would still be the best alternative for us.

My understanding is that unless a second DAG run is triggered manually, each file will only be parsed once, so ingestion should run fine. (If concurrency is set then this will ingest multiple files in parallel, but each individual file should still be parsed only once.)

So I think the only case where problems _may_ occur is if multiple DAG runs are triggered manually in quick succession and those multiple runs interfere with each other (which shoudn't happen in practice because it'll be the Airflow scheduler that triggers runs at defined intervals, typically once a day).

It's probably fine to consider this scenario a fringe case which we don't need to deal with (at least not for now). I was just thinking that if there is an easy way to avoid any race conditions (even for this fringe case) then we may as well do that.

danwilliams:
> So I think the only case where problems may occur is if multiple DAG runs are triggered manually in quick succession and those multiple runs interfere with each other (which shoudn't happen in practice because it'll be the Airflow scheduler that triggers runs at defined intervals, typically once a day).

Unless they are run for the same file (which shouldn't work - because it'll be marked as ingested in the first run anyway) - that shouldn't be a problem. It creates the tables with { cdr_type }_{ ds_nodash } which means it will only try to use the same table for the same file `CALLS_date.csv.gz` - which shouldn't work because of the above. This means this should be fine then, unless I'm missing anything? Or do you think the issue will be with manually triggering a build right before the init task changes the ingestion status for a file?

Updated the lambdas to have a verb in the name, and the execution_time to execution_date for consistency.

flowetl/etl/etl/etl_utils.py
```diff
@@ -114,6 +114,12 @@ def construct_etl_dag(
         schedule_interval=None,
         default_args=default_args,
         template_searchpath=config_path,  # template paths will be relative to this
+        user_defined_macros={
+            "get_extract_table": lambda execution_date: f"etl.x{ cdr_type }_{ execution_date }",
+            "get_transform_table": lambda execution_date: f"etl.t{ cdr_type }_{ execution_date }",
+            "get_load_table": lambda execution_date: f"events.{ cdr_type }_{ execution_date }",
+            "cdr_type": cdr_type.lower(),
```
maxalbert: I see why you added the `.lower()` here but I think it should be fine to omit it and require the user to provide the correct capitalisation.
```suggestion
            "cdr_type": cdr_type,
```
(This will also avoid a discrepancy between the `cdr_type` macro and the value of `cdr_type` in the table macros above which don't use `.lower()`, although admittedly this is mostly a cosmetic detail).

maxalbert:
> Or do you think the issue will be with manually triggering a build right before the init task changes the ingestion status for a file?

Yes, that was the corner case I was pondering about, but I agree this is most likely a non-issue in practice so I'm happy to keep the current approach. 👍

maxalbert:
Turns out the integration tests were failing because of a wrong interpolation of the `CDRType` enums (where the raw value `CDRType.CALLS` was passed instead of the string `"calls"`, for example). I just pushed a commit to fix this, so hopefully this should be ready to merge now.

PR #997, opened 2019-06-28
Updated airflow container to use gosu for permissions dropdown (#716)

### LINKED ISSUE #716: Setting the GID/UID for flowetl container
Currently using [fixuid](https://github.com/boxboat/fixuid) to set user/group at runtime for the airflow user. The maintainers say the following about production use -> boxboat/fixuid/issues/1.

@greenape suggests using same method as PG does but my understanding is that has exactly the same issue of dropping down to root during startup of the container...

Of course other solution is to rebuild image with a given UID/GID for each possible server it is deployed on...

danwilliams:
Closes #716 

### I have:

### Description

Gosu is a similar binary that drops privileges to the expected used but does not do any permissions change itself so those are handled outside.

Previous approach was still running all processes as root, ignoring the USER airflow line.


greenape:
Definitely like this approach - useful beyond the development scope I think (e.g. anytime you want to be able to bind-mount to load dags without needing to worry about having a user on the host which matches the uid and GID of the airflow one in-container.)

Can we test it on CI tho? I'm aware that the dev team are primarily on Mac, and that docker behaves quite differently when it comes to permissions and so on between Mac and Linux.

flowetl/Dockerfile
```diff
@@ -32,6 +38,11 @@ RUN pip install /opt/etl
 
 USER airflow
 COPY --chown=airflow:airflow ./dags/* ${AIRFLOW_HOME}/src/
+
+# Needed on $AIRFLOW_HOME so that different users passed through --user
+# have the permission to create and use their files for Airflow
+RUN chmod -R 777 ${AIRFLOW_HOME}
```
greenape: PG image does this, and then replaces with 700 when the container actually runs. Any specific reason not to do similarly?

danwilliams: In PG's case, it does the chmod when run as root, and `chmod 700 "$PGDATA" 2>/dev/null || :` which allows it to fail if chown also fails. In our case, if it's not started with uid 0, the chown and chmod would fail because that uid/gid won't be the owner or root, but I agree that's a good idea to do it _when possible_. Added that.

danwilliams:
> Can we test it on CI tho? I'm aware that the dev team are primarily on Mac, and that docker behaves quite differently when it comes to permissions and so on between Mac and Linux.

It has been tested on Linux, and on CI... do you mean adding some more tests that may be missing?

PR #1016, opened 2019-07-01
Fix location_event_counts not working with subsets of event types

### LINKED ISSUE #1015: EventTypes custom field needs to deserialise with schema
The `EventTypes` custom field validates as a list containing any combination of the events tables, and hence gets used to populate the `table`/`tables` argument. This won't actually work though, because `table`/`tables` are expected to be schema-qualified with `events.TABLE_NAME`.

So, until the machinery inside is changed to be a bit more sensible - this needs to deserialise with the `events.` prefix.

greenape:
Closes #1015

### I have:

### Description

Added a custom deserialiser to `EventTypes` field which dedupes the list, and appends the `events.` prefix required to use this as a `table` argument.


flowmachine/flowmachine/core/server/query_schemas/custom_fields.py
```diff
@@ -11,7 +11,9 @@
 
 class EventTypes(fields.List):
     """
-    A string representing an event type, for example "calls", "sms", "mds", "topups".
+    A list of strings representing an event type, for example "calls", "sms", "mds", "topups".
+
+    When deserialised, will be uniqueified, and prefixed with "events."
```
maxalbert: What do you mean by "uniqueified"?

(And where is this actually happening - in the call to `super()`?)

greenape:
```suggestion
    When deserialised, will be deduped, and prefixed with "events."
```

flowmachine/flowmachine/core/server/query_schemas/custom_fields.py
```diff
@@ -29,6 +31,12 @@ def __init__(self, required=False, validate=None, **kwargs):
             **kwargs,
         )
 
+    def _deserialize(self, value, attr, data, **kwargs):
```
maxalbert: Can we stick a comment in here to say that this is a temporary workaround until the underlying issue mentioned in #1015 is solved and then we can hopefully remove this again?

greenape:
```suggestion
    def _deserialize(self, value, attr, data, **kwargs):
        # Temporary workaround for https://github.com/Flowminder/FlowKit/issues/1015 until underlying issue resolved
```

PR #1018, opened 2019-07-01
Tasks/968 expose top up balance api

### LINKED ISSUE #968: Expose TopUpBalance via api
- [x] Add marshmallow spec for `TopUpBalance`
- [x] Add to allowable joined aggregate queries
- [x] Add flowclient function `topup_balance(*, start, stop, level, subscriber_subset)`
- [x] Add test to `integration_tests/tests/test_queries.py`
- [ ] Update acceptance tests
  - [ ] integration_tests/tests/flowapi_tests/test_api_spec.py
  - [ ] integration_tests/tests/flowmachine_server_tests/test_server.py
- [x] Update `docs/source/4-developer.md` with newly added query
- [x] Update Changelog

gzagatti:
Closes #968

### I have:

### Description

FlowAPI's 'joined_spatial_aggregate' endpoint now exposes top-up balance.


gzagatti:
I would like to add to PR description, that I added the `mode` statistic to `TopUpBalance` because one of the exposed statistics was mode which was not available in `TopUpBalance`. I have added the relevant test for this statistics.

In addition to that, I have added a test to see if `TopUpBalance` can be joined with a spatial aggregate. But it would be probably a better idea if we added a general test that tested if any objected derived from `SubscriberFeature` can be joined with a spatial aggregate.

flowmachine/flowmachine/core/server/query_schemas/flowmachine_query.py
```diff
@@ -25,7 +25,6 @@
 from .unique_subscriber_counts import UniqueSubscriberCountsSchema
 from .location_introversion import LocationIntroversionSchema
 from .total_network_objects import TotalNetworkObjectsSchema
-from .radius_of_gyration import RadiusOfGyrationSchema
```
greenape: 👍

flowmachine/flowmachine/features/subscriber/topup_balance.py
```diff
@@ -215,6 +225,17 @@ def _make_query(self):
             """
             return sql
 
+        if self.statistic in {"mode"}:
+            sql = f"""
+            SELECT DISTINCT ON (subscriber) subscriber, balance AS value
```
greenape: Can't one just do `mode() WITHIN GROUP(ORDER BY balance) as value`?

gzagatti: The `TopUpBalance` class weights the statistics it computes by the time that the subscriber held a given balance. All the other stats in this class are weighted in that way, so I think the mode should also be weighted to ensure consistency. In this case, the mode balance is the balance that was held for most of the time. That is, the balance that shows the highest weight. The postgres built-in `mode` function will only compute the most common balance without regards to the weight.

We can contrast the above with the mode for top up recharge, which will follow exactly what you suggested here.

gzagatti:
I agree with your concerns regarding the stat combos. 

This also kept coming into my mind when computing CDR covariates for our projects. Should we compute a mean of the covariates for all subscribers who are assigned a home location to a certain site? Or should we compute the mean of the site?

I think both are valid and depend on what you are capturing. If we look to a business district, then the mean of the site and the mean of the mean of its residents can be very different.

All in all, I think it should be the choice of the analyst on what he wants and Flowmachine should support both. Ideally, a covariates should be defined with regards to CDR events (eg. top-up balance, call duration, SMS volume, etc) and then aggregated according to whichever dimension we are interested in (eg. subscriber, temporal, geographic, etc).

With this in mind, I think this whole discussion would deserve a whole issue on its own. For now, I would just keep it consistent with the other API calls, which is to take the mean of the mean.

CHANGELOG.md
```diff
@@ -29,6 +29,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - FlowAPI's 'joined_spatial_aggregate' endpoint now exposes unique location counts.[#949](https://github.com/Flowminder/FlowKit/issues/949)
 - FlowAPI's 'joined_spatial_aggregate' endpoint now exposes subscriber degree.[#969](https://github.com/Flowminder/FlowKit/issues/969)
 - Flowdb now contains an auxiliary table to record outcomes of queries that can be run as part of the regular ETL process [#988](https://github.com/Flowminder/FlowKit/issues/988)
+- FlowAPI's 'joined_spatial_aggregate' endpoint now exposes top-up balance.[#968](https://github.com/Flowminder/FlowKit/issues/968)
```
```suggestion
```

PR #1022, opened 2019-07-02

### LINKED ISSUE #947: Use execution_date in dags

danwilliams:
Closes #947 

### I have:

### Description

Fixes some test failures.


maxalbert:
Closing this PR as this change has already been included in #996 (see https://github.com/Flowminder/FlowKit/pull/996/commits/83b153293035a89890951f5e21929eb6a3db5dda).

PR #1028, opened 2019-07-03
Allow ingestion from postgres table in addition to csv files

### LINKED ISSUE #1024: Modify extract.sql and transform.sql templates for more consistent treatment of ingestion from CSV files and external postgres table
The `extract.sql` and `transform.sql` templates should be slightly modified in terms of what they do, in order to better fit with the conceptual model of the ETL steps and to allow us to treat ingestion from CSV and an external Postgres db more similarly.

Currently the example template `extract.sql` creates an empty table in the same format as `events.{event_type}`, and `transform.sql` simply does a `SELECT *` from this into a new table, which is then hooked as a partition into `events.{event_type}`.

However, it would be a better fit to the E-T-L steps if `extract.sql` simply made the raw data visible and it was the responsibility of `transform.sql` to ensures that it is transformed into the correct shape (i.e., consistent with `events.{event_type}`).

Let's think this through more concretely for the two main raw data sources we currently support: external CSV files and external PostgreSQL tables. Both of these will be made visible to FlowDB via foreign data wrappers during the ETL process.

I think`extract.sql` should simply create a VIEW into the foreign data wrapper which selects the appropriate data for the subsequent transform and load steps. For example, in the case of ingestion from CSV files this might simply be a view around the full fdw-wrapped CSV file (or a union-ed SELECT from multiple external CSV files if the data for a single date is split up over multiple files). Similarly, for ingestion from another postgres database it might be a simple VIEW along the lines of `SELECT * FROM {fdw_wrapped_table} WHERE {today} <= event_time AND event_time < {tomorrow}`.

Then the `transform.sql` template can read the data from the extract VIEW and transform it into a table that has the same shape as `events.{event_type}`, so that it can subsequently just be hooked as a partition into `events.{event_type}`.

### LINKED ISSUE #1027: Flowetl should allow ingesting from both csv files and an external postgres db
Currently FlowETL assumes that ingestion will happen from CSV files but it is possible that raw data is provided via an external PostgreSQL database. FlowETL should support both options.

maxalbert:
Closes #1027 
Closes #1024 

### I have:

### Description

Changes (in rough order of appearance in the diff on Github):

- I have slightly re-rigged the logic which chooses between `testing` and `production` (in `flowetl/dags/etl.py` and `flowetl/dags/etl_sensor.py`). Previously one would set `TESTING="true"` to enable the testing DAG callables, and otherwise it would use `production`. This is fine if those are the only options, but for local development it can be useful to (temporarily) add third option so that one can play with alternative DAGs and callables. I have replaced the boolean env var `TESTING` with an env var called `FLOWETL_RUNTIME_CONFIG` which can take any value, so one can set `FLOWETL_RUNTIME_CONFIG={"testing"|"production"|"dev"}` and more easily experiment with alternative DAG setups.

- The file `flowetl/etl/etl/config_constant.py` was removed because it is basically a duplicate (pure-Python) definition of the config in `flowetl/mounts/config/config.yml`, so when tweaking that config one must remember to tweak both files. Instead, I added a pytest fixture which loads the config from `config.yml` so that it can be used in the tests (which is the only place where `config_constant.py` was referenced previously).

- The structure of the `config.yml` file has been changed. Previously, it assumed that ingestion would happen from csv files and used the following structure to configure this:
```
etl:
  calls:
    concurrency: 4
    pattern: CALLS_(\d{4}[01]\d[0123]\d).csv.gz
  sms:
    concurrency: 4
    pattern: SMS_(\d{4}[01]\d[0123]\d).csv.gz
```
Now it is possible to choose between ingestion from csv or another postgres table and the config now allows configuring both as follows:
```
etl:
  calls:
    concurrency: 4
    source:
      source_type: csv
      filename_pattern: CALLS_(\d{4}[01]\d[0123]\d).csv.gz
  sms:
    concurrency: 4
    source:
      source_type: sql
      table_name: "sms_raw_data_dump"
```

- The sql templates `extract.sql` and `transform.sql` have been modified accoording to the suggestions in #1024. Now `extract.sql` simply creates a view instead of a table (note that in production this means there needs to be a pre-step to set up the underlying foreign data wrapper that this view refers to; for file-based ingestion this should probably be FlowETL's responsibility? - if so it should be added in a subsequent PR).

- The loop logic in `production_trigger__callable` has been inverted. Previously it would first find all the unprocessed files and then trigger the DAGs for all CDR types. This now happens the other way around: the outer loop iterates over all CDR types, then finds all unprocessed dates (either based on the filename or based on the dates present in the postgres table containing the raw data, depending on the source type (csv/sql) configured for this CDR type) and triggers a DAG run for each unprocessed date.

- Added an env var called `FLOWETL_INTEGRATION_TESTS_DISABLE_PULLING_DOCKER_IMAGES` to avoid pulling docker images from docker cloud before running the tests. This is needed when running the integration tests locally (in addition to running `make flowetl-build`) because otherwise the locally built images won't be used to spin up the docker containers for the tests.

Minor or unrelated changes:

- Renamed the folder where DAG definitions are stored from `src` to `dags` (this is only visible inside the container so doesn't really matter, but I found it a bit more intuitive when reading the source code).

- Renamed `trigger__callable` to `production_trigger__callable` for clarity.

- I added an explicit install of `psycopg2-binary` to the Pipfile to avoid a warning about upcoming change to the psycopg2 wheel.


flowetl/etl/etl/production_task_callables.py
```diff
@@ -84,36 +108,75 @@ def trigger__callable(
...
-            execution_date=cdr_date,
-            run_id=f"{file.name}-{str(uuid)}",
-            conf=config,
-            replace_microseconds=False,
-        )
+    for cdr_type, cfg in cdr_type_config.items():
+        cdr_type = CDRType(cdr_type)
+
+        source_type = cfg["source"]["source_type"]
+        print(f"{cdr_type!r} ({source_type}): {cfg}")
```
danwilliams: Should this use logger instead?

flowetl/etl/etl/production_task_callables.py
```diff
@@ -84,36 +108,75 @@ def trigger__callable(
...
+
+        if source_type == "csv":
+            filename_pattern = cfg["source"]["filename_pattern"]
+            print(f"Filename pattern: {filename_pattern!r}")
```
danwilliams: Should this use logger instead?

flowetl/etl/etl/production_task_callables.py
```diff
@@ -84,36 +108,75 @@ def trigger__callable(
...
+                if ETLRecord.can_process(
+                    cdr_type=cdr_type, cdr_date=date, session=session
+                )
+            ]
+            logger.info(f"Dates found: {dates_present}")
+            logger.info(f"Unprocessed dates: {unprocessed_dates}")
+
+            for cdr_date in unprocessed_dates:
+                config = {
+                    "cdr_type": cdr_type,
```
danwilliams: For consistency, can we use the `cdr_type` here, in the `run_id` as above (@138) too?

flowetl/tests/integration/conftest.py
```diff
@@ -173,11 +172,23 @@ def mounts(postgres_data_dir_for_tests, flowetl_mounts_dir):
 
 @pytest.fixture(scope="session", autouse=True)
 def pull_docker_images(docker_client, container_tag):
-    logger.info(f"Pulling docker images (tag='{container_tag}')...")
-    docker_client.images.pull("postgres", tag="11.0")
-    docker_client.images.pull("flowminder/flowdb", tag=container_tag)
-    docker_client.images.pull("flowminder/flowetl", tag=container_tag)
-    logger.info("Done pulling docker images.")
+    disable_pulling_docker_images = (
```
danwilliams: 🎖 👍

maxalbert:
Ok, the tests finally pass and have full test coverage again. :)

I needed to make a few more changes to the code. I tweaked/replaced some helper functions so that they would be slightly easier to test, and in particular I needed to replace the in-memory SQLite database with a postgres one (via [testing.postgresql](https://pypi.org/project/testing.postgresql/)) because the SQL templates now use syntax that SQLite doesn't support. This is not ideal because it increases the unit test runtime by ~2s and ideally I'd like to keep them as fast as possible so that they can be run frequently during development but it seems an ok compromise for now until we can refactor things further (e.g. it could be useful to allow providing the extract/transform SQL templates as sqlalchemy code snippets rather than just raw SQL).

@danwilliams It would be great if you could take another look and let me know if you have any suggestions for modifications, or approve the PR if you're happy with the changes (the approval is needed to be able to merge this into master).

PR #1057, opened 2019-07-09
Tasks/843 flowdb compose duplicates

### LINKED ISSUE #843: Base compose file won't work with docker-compose up
Via @josh-gree 

The current compose file will just error if used directly with `docker-compose up`, which isn't ideal.

What would work as an alternative, is to split out the three flowdb containers into two additional compose files, so you can override the flowdb container.

e.g. `docker-compose up` gives you a bare flowdb, `docker-compose up -f docker-compose.yml -f docker-compose-testdata.yml` gives you testdata etc.

danwilliams:
Closes #843 

### I have:

### Description

This splits out the `flowdb_testdata` and `flowdb_syntheticdata` into two additional compose files that can be used in combination with the base one so that the base one also works by itself. One downside of this is that we can't make use of the YAML alias anymore, and the flowdb env vars needed to be duplicated in the additional compose files.

The `quick_start.sh` file needed changing since we have now more than one docker-compose file. Logic has been added to create a temporary directory using `mktemp` in which to temporarily download the docker-compose files, use a common prefix for the docker instance names (`flowkit_qs` to not conflict with other docker instances not started using the quick start file locally), and auto-cleanup when the `quick_start.sh` script finishes by removing the temporary folder. Also, the Makefile has been updated to extract the duplicated docker-compose commands to the top (so that common flags / paths can be changed in one place).


maxalbert:
Very nice! I like the additional Makefile variables, which make the individual targets a lot clearer to read. 👍 

There is a tiny bit of duplication in these because the flags `-f $(DOCKER_COMPOSE_FILE) -f $(DOCKER_COMPOSE_SYNTHETICDATA_FILE) -f $(DOCKER_COMPOSE_TESTDATA_FILE)` are repeated in each, but not a big deal. Happy to merge this as is.

Issue #1029, opened 2019-07-03
Last and most-common handset type via api


greenape:
Basically identical with #1011

gzagatti:
Solved with #1059.

Issue #421, opened 2019-02-26
Column names across queries should be unified and more generic (where appropriate)


maxalbert:
A lot of queries return tables where at least one column name is specific to the metric being calculated (or to the aggregation level, etc.). For example, in `DistanceCounterparts` the column names are `subscriber` and `distance_{statistic}` (where `statistic` can be count/sum/min/max/avg/...). This is a particularly obvious example but even if the column name is fixed, it differs across different queries.

We should work towards unifying this and giving the column a generic name (e.g. `value`). After all, the caller _knows_ which query kind or statistic they requested. This would both simplify a lot of code in flowmachine itself because we don't need to keep track of column names all over the place and also make life easier for any calling code (and ultimately the user) because code which further wrangles the data can assume a consistent column name and doesn't need a case distinction based on some input parameter.

PR #1059, opened 2019-07-09
Tasks/1011 expose last and modal handset api

### LINKED ISSUE #1011: Enable last & modal handset via api
Note: will only work at spatial level with modal average. May need special handling.

gzagatti:
Closes #1011 and #1029

### I have:

### Description

This PR augments FlowAPI's 'joined_spatial_aggregate' endpoint by exposing handset.[#1011](https://github.com/Flowminder/FlowKit/issues/1011).

Handset characteristics are qualitative metrics such as handset type, brand, OS, etc. Therefore, traditional aggregation methods like "avg", "std", "median", etc do not make sense. For qualitative metrics, it is most appropriate to return a frequency table with relative counts. In order to do that, this PR introduces a new aggregation method to `JoinedSpatialAggregate` called `dist` which computes the frequency table.

Given the different type of aggregation methods, `JoinedSpatialAggregateSchema` conducts another set of validation of the method parameter which ensures that only the correct methods are called with quantitative and qualitative metrics.

This PR also introduces `SubscriberHandsetCharacteristic` to FlowMachine. This query class is built on top of `SubscriberHandset`. It selects the characteristic of interest out of all characteristics made available by `SubscriberHandset`. This comes in handy when exposing the API. To avoid redundancy, `SubscriberPhoneType` was removed.


flowclient/flowclient/client.py
```diff
@@ -1475,3 +1475,49 @@ def nocturnal_events(
...
+        "brand",
+        "model",
+        "software_os_name",
+        "software_os_vendor",
+    ],
+    method: str = ["last", "most-common"],
+    subscriber_subset: Union[dict, None] = None,
+) -> dict:
+    """
+    Return query spec for nocturnal events
```
maxalbert:
```suggestion
    Return query spec for handset
```

CHANGELOG.md
```diff
@@ -16,6 +16,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - FlowETL now supports ingesting from a postgres table in addition to CSV files. [#1027](https://github.com/Flowminder/FlowKit/issues/1027)
 - `FLOWETL_RUNTIME_CONFIG` environment variable added to control which DAG definitions the FlowETL integration tests should use (valid values: "testing", "production").
 - `FLOWETL_INTEGRATION_TESTS_DISABLE_PULLING_DOCKER_IMAGES` environment variable added to allow running the FlowETL integration tests against locally built docker images during development.
+- FlowAPI's 'joined_spatial_aggregate' endpoint now exposes handset.[#1011](https://github.com/Flowminder/FlowKit/issues/1011) and [#1029](https://github.com/Flowminder/FlowKit/issues/1029)
+- `JoinedSpatialAggregate` now supports "dist" stats which computes outputs the relative distribution of the passed metrics.
```
maxalbert: Can we rename the parameter `dist` to `distr`, to avoid confusion with "distance"?

flowclient/flowclient/client.py
```diff
@@ -1207,8 +1207,8 @@ def joined_spatial_aggregate(
         Modal or daily location query to use to localise the metric
     metric: dict
         Metric to calculate and aggregate
-    method: {"avg", "max", "min", "median", "mode", "stddev", "variance"}, default "avg"
-        Method of aggregation one of "avg", "max", "min", "median", "mode", "stddev" or "variance".
+    method: {"avg", "max", "min", "median", "mode", "stddev", "variance", "dist"}, default "avg".
+        Method of aggregation one of "avg", "max", "min", "median", "mode", "stddev", "variance" or "dist". If the metric is qualitative it will only accept the "dist" method which yields the distribution of the metric in relative terms. All of the other methods will be rejected. On the other hand, the "dist" method will be rejected for all quantitative metrics.
```
maxalbert: It wasn't quite clear to me what "qualitative" means in this context until I looked at the code/examples. This is about the distinction between continuous and categorical variables, right? Would it be useful to rephrase the docstring to make this clearer, e.g. along the following lines?
```suggestion
       Method of aggregation; one of "avg", "max", "min", "median", "mode", "stddev", "variance" or "dist". If the metric refers to a categorical variable (e.g. a subscriber handset type) it will only accept the "dist" method which yields the relative distribution of possible values. All of the other methods will be rejected. On the other hand, the "dist" method will be rejected for all continuous variables.
```

flowclient/flowclient/client.py
```diff
@@ -1475,3 +1475,49 @@ def nocturnal_events(
...
+    method: str = ["last", "most-common"],
+    subscriber_subset: Union[dict, None] = None,
+) -> dict:
+    """
+    Return query spec for nocturnal events
+
+    Parameters
+    ----------
+    start : str
+        ISO format date of the first day for which to count nocturnal events, e.g. "2016-01-01"
```
maxalbert:
```suggestion
        ISO format date of the first day for which to count handsets, e.g. "2016-01-01"
```

flowclient/flowclient/client.py
```diff
@@ -1475,3 +1475,49 @@ def nocturnal_events(
...
+    stop : str
+        ISO format date of the day _after_ the final date for which to count nocturnal events, e.g. "2016-01-08"
```
maxalbert:
```suggestion
        ISO format date of the day _after_ the final date for which to count handsets, e.g. "2016-01-08"
```

flowmachine/flowmachine/core/server/query_schemas/joined_spatial_aggregate.py
```diff
@@ -47,9 +49,32 @@ class JoinedSpatialAggregateSchema(Schema):
...
+        quant_metrics = [
+            "radius_of_gyration",
+            "unique_location_counts",
+            "topup_balance",
+            "subscriber_degree",
+            "topup_amount",
+            "event_count",
+            "nocturnal_events",
+        ]
+        qual_metrics = ["handset"]
```
maxalbert: See comment above. I feel it would help to rename these variables to reflect the continuous vs. categorical distinction as opposed to qualitative/quantitative (because even the categorical metrics still return a quantitative result).

flowmachine/flowmachine/core/server/query_schemas/joined_spatial_aggregate.py
```diff
@@ -47,9 +49,32 @@ class JoinedSpatialAggregateSchema(Schema):
...
+        if data["metric"]["query_kind"] in quant_metrics:
+            validate = OneOf(
+                ["avg", "max", "min", "median", "mode", "stddev", "variance"]
+            )
+        elif data["metric"]["query_kind"] in qual_metrics:
+            validate = OneOf(["dist"])
+        else:
+            raise ValidationError(
+                "{data['metric']['query_kind']} does not have a valid metric type."
```
maxalbert:
```suggestion
                f"{data['metric']['query_kind']} does not have a valid metric type."
```

flowmachine/flowmachine/features/subscriber/subscriber_tacs.py
```diff
@@ -17,6 +17,62 @@
 from .metaclasses import SubscriberFeature
 from ...core import Table
 
+valid_characteristics = {
```
maxalbert: Can we sort these alphabetically so that they are easier to parse at a glance?

For convenience, here is the sorted list:
```
valid_characteristics = {
    "brand",
    "depth",
    "display_colors",
    "display_height",
    "display_type",
    "display_width",
    "hardware_bluetooth",
    "hardware_edge",
    "hardware_gprs",
    "hardware_gps",
    "hardware_umts",
    "hardware_wifi",
    "height",
    "hnd_type",
    "j2me_cldc_10",
    "j2me_cldc_11",
    "j2me_cldc_20",
    "j2me_midp_10",
    "j2me_midp_20",
    "j2me_midp_21",
    "mms_built_in_camera",
    "mms_receiver",
    "model",
    "software_os_name",
    "software_os_vendor",
    "software_os_version",
    "syncml_dm_acc_gprs",
    "syncml_dm_app_bookmark",
    "syncml_dm_app_browser",
    "syncml_dm_app_internet",
    "syncml_dm_app_java",
    "syncml_dm_app_mms",
    "syncml_dm_settings",
    "tac",
    "wap_1_2_1",
    "wap_2_0",
    "wap_push_oma_app_browser",
    "wap_push_oma_app_ims",
    "wap_push_oma_app_internet",
    "wap_push_oma_app_mms",
    "wap_push_oma_app_poc",
    "wap_push_oma_cp_bookmarks",
    "wap_push_oma_settings",
    "wap_push_ota_app_browser",
    "wap_push_ota_app_internet",
    "wap_push_ota_app_mms",
    "wap_push_ota_bookmarks",
    "wap_push_ota_multi_shot",
    "wap_push_ota_settings",
    "wap_push_ota_single_shot",
    "wap_push_ota_support",
    "weight",
    "width",
}
```

flowmachine/flowmachine/features/subscriber/subscriber_tacs.py
```diff
@@ -357,15 +413,31 @@ def _make_query(self):
...
-    Class representing which type of phone (Basic, Feature, or Smart) a subscriber
-    is associated with.
+    Class extracting a single characteristic from the handset.
 
     Parameters
-    ----------
+    ---------
     start, stop : str
          iso-format start and stop datetimes
+    characteristic: {'tac', 'syncml_dm_app_browser', 'height',
```
maxalbert: See above, would be useful to sort these alphabetically please.

flowmachine/flowmachine/features/subscriber/subscriber_tacs.py
```diff
@@ -380,9 +452,10 @@ class SubscriberPhoneType(SubscriberFeature):
     Examples
     -------------
 
-    >>> subscriber_smart = SubscriberPhoneType('2016-01-01', '2016-01-07')
+    >>> subscriber_smart = SubscriberHandsetCharacteristic('2016-01-01', '2016-01-07',
+        'hnd_type')
     >>> subscriber_smart.get_dataframe()
-             subscriber handset_type
+             subscriber     hnd_type
```
maxalbert: This should use `value` as the generic column name please. (See https://github.com/Flowminder/FlowKit/issues/421)

maxalbert: Ah, I just saw that this is already implemented below so this was just a leftover in the docstring. 👍

flowmachine/tests/test_spatial_aggregate.py
```diff
@@ -36,3 +41,22 @@ def test_can_be_aggregated_lon_lat(get_dataframe):
...
+        spatial_unit=make_spatial_unit("admin", level=3),
+        method="most-common",
+    )
+    metric = SubscriberHandsetCharacteristic(
+        "2016-01-01", "2016-01-02", characteristic="hnd_type"
+    )
+    agg = JoinedSpatialAggregate(metric=metric, locations=locations, method="dist")
+    df = get_dataframe(agg)
+    assert ["pcod", "metric", "key", "value"] == list(df.columns)
+    assert df[df.metric == "value"].groupby("pcod").sum().value.unique() == 1
```
maxalbert:
```suggestion
    assert all(df[df.metric == "value"].groupby("pcod").sum() == 1.0)
```

flowmachine/tests/test_subscriber_tacs.py
```diff
@@ -106,20 +99,27 @@ def test_subscriber_handset(get_dataframe):
     assert tc.loc["1p4MYbA1Y4bZzBQa"].model == "LB-01"
 
 
-def test_subscriber_phonetype(get_dataframe):
-    """Check that correct smart/feature label is returned."""
+def test_subscriber_handset_characteristic(get_dataframe):
+    """Check that correct handset characteristic is returned."""
```
maxalbert:
```suggestion
    """Check that correct handset characteristic is returned for selected subscribers."""
```

integration_tests/tests/query_tests/test_queries.py
```diff
@@ -490,6 +490,40 @@
...
+                    "query_kind": "handset",
+                    "start_date": "2016-01-01",
+                    "end_date": "2016-01-02",
+                    "characteristic": "hnd_type",
+                    "method": "last",
+                },
+                "method": "dist",
+            },
+        ),
+        (
```
maxalbert: If I am not mistaken this set of test parameters is an exact duplicate of the previous ones (just written in a different way), so it's sufficient to only keep one of them to avoid running the exact same test twice.

integration_tests/tests/query_tests/test_queries.py
```diff
@@ -503,6 +537,60 @@ def test_run_query(query_kind, params, universal_access_token, flowapi_url):
...
+                    "end_date": "2016-01-02",
+                    "characteristic": "hnd_type",
+                    "method": "last",
+                },
+                "method": "avg",
+            },
+        ),
+    ],
+)
+def test_fail_query_incorret_parameters(
```
maxalbert:
```suggestion
def test_fail_query_incorrect_parameters(
```

integration_tests/tests/query_tests/test_queries.py
```diff
@@ -503,6 +537,60 @@ def test_run_query(query_kind, params, universal_access_token, flowapi_url):
...
+    query_kind, params, universal_access_token, flowapi_url
+):
+    """
+    Test that queries fail with incorrect parameters.
+    """
+    query_spec = getattr(flowclient, query_kind)(**params)
+    con = flowclient.Connection(url=flowapi_url, token=universal_access_token)
+    with pytest.raises(Exception):
```
maxalbert: Great to have this test for failure with incorrect parameters, but I suggest making the assertion a bit more specific:
- Capture the specific type of exception you expect (e.g. `FlowclientConnectionError`?) rather than a generic one.
- Add a `match="..."` argument to `pytest.raises(...)` so that we know the error message contains the expected test.

Otherwise it's quite possible that the test passes because _some_ exception is raised, but it's actually hiding a different error which we don't get to see. (Which isn't the case here but it's definitely happened to me in the past.)

I suggest something along the lines of:
```
with pytest.raises(FlowclientConnectionError, match="Must be one of: avg, max, min, median, mode, stddev, variance."):
    ...
```

maxalbert: FWIW, the value of the `match` parameter may (or may not) need tweaking now that #1056 is merged.

gzagatti:
@maxalbert, thanks for the review.

I agree with all the inline comments. I will make the necessary changes tomorrow.

Regarding your comments at the end, I had the same thought as you. It felt unnecessary to copy the whole TAC table to retrieve only a few columns. My thought in this case was that if the analyst (or multiple analysts) required multiple characteristics, it would be a bit of a waste of time to keep doing the merge of the subscriber table with the TAC table for each single requested characteristic (especially if in the limit an analyst would request all characteristics one by one). 

I think that if we had a separate subscriber table that kept track of the subscriber's TACs (along with other useful things), then I would be all in favor of having the query class selecting only the relevant TAC column.

Therefore and apart from the limited time available, I would suggest to make these modifications after discussing the above issue in a separate PR.

flowmachine/flowmachine/features/subscriber/subscriber_tacs.py
```diff
@@ -357,15 +413,70 @@ def _make_query(self):
...
-class SubscriberPhoneType(SubscriberFeature):
+class SubscriberHandsetCharacteristic(SubscriberFeature):
     """
-    Class representing which type of phone (Basic, Feature, or Smart) a subscriber
-    is associated with.
+    Class extracting a single characteristic from the handset.
 
     Parameters
-    ----------
+    ---------
```
maxalbert: Probably a nitpick, but it's possible that the docs formatter needs the underline to have the same length as the line above it for it to be displayed correctly.
```suggestion
    ----------
```

flowmachine/flowmachine/features/subscriber/subscriber_tacs.py
```diff
@@ -380,9 +491,10 @@ class SubscriberPhoneType(SubscriberFeature):
     Examples
     -------------
 
-    >>> subscriber_smart = SubscriberPhoneType('2016-01-01', '2016-01-07')
+    >>> subscriber_smart = SubscriberHandsetCharacteristic('2016-01-01', '2016-01-07',
+        'hnd_type')
     >>> subscriber_smart.get_dataframe()
-             subscriber handset_type
+             subscriber     value
```
maxalbert:
```suggestion
             subscriber        value
```

PR #1061, opened 2019-07-10
Flowmachine displacement

### LINKED ISSUE #1010: Enable displacement via api
- [ ] Marshmallow schema for `Displacement`
- [ ] Rename returned column to `value`
- [ ] Make modal_location argument mandatory
- [ ] Rename modal_location arg to reference_location
- [ ] Type hint reference_location
- [ ] Remove creation of modal location from args
- [ ] Remove any modal location only args
- [ ] Add to joined spatial aggregate allowable
- [ ] Add to client
- [ ] Add to `docs/source/developer/developer.md`

BhavinPanch:
Closes #1010 

### I have:

FlowAPI 'joined_spatial_aggregate' expose displacement value for referance location.


flowmachine/flowmachine/core/server/query_schemas/displacement.py
```diff
@@ -0,0 +1,63 @@
...
+from .custom_fields import SubscriberSubset, Statistic
+from .daily_location import DailyLocationSchema, DailyLocationExposed
+
+
+__all__ = ["DisplacementSchema", "DisplacementExposed"]
+
+
+class InputToDisplacementSchema(OneOfSchema):
+    type_field = "query_kind"
+    type_schemas = {"daily_location": DailyLocationSchema}
```
maxalbert: Please add modal location to this as well (because `displacement` doesn't care what _kind_ of reference location it receives as input).

maxalbert: (Note to self: at some point we likely need some kind of "GenericLocationSchema" which subsumes these, but that's certainly out of scope for this PR.)

flowmachine/flowmachine/features/subscriber/displacement.py
```diff
@@ -42,11 +42,11 @@ class Displacement(SubscriberFeature):
         e.g. 2016-01-01 or 2016-01-01 14:03:01
     stop : str
         As above
-    modal_locations : ModalLocation
+    reference_location : ModalLocation
```
maxalbert:
```suggestion
    reference_location : BaseLocation
```

flowmachine/flowmachine/features/subscriber/displacement.py
```diff
@@ -42,11 +42,11 @@ class Displacement(SubscriberFeature):
...
         The set of home locations from which to calculate displacement.
         If not given then ModalLocation Query wil be created over period
         start -> stop.
-    statistic : str
+    value : str
```
maxalbert: Actually, the _input_ parameter should still be called `statistic` (otherwise it's not really intuitive for the user what it means). So this needs to be changed back.

For clarity of anyone reading the code I would also suggest to keep the name `statistic` anywhere internally in this class.

So I'd only change the column name of the returned output to the name `value` (as you have done in L193 below).

flowmachine/flowmachine/features/subscriber/displacement.py
```diff
@@ -97,8 +97,8 @@ def __init__(
         self,
         start,
         stop,
-        modal_locations=None,
-        statistic="avg",
+        reference_location,
+        value="avg",
```
maxalbert: See comment above, I'd change this back to `statistic` (and similarly below, except the column name in the SQL query in L193).

flowmachine/flowmachine/features/subscriber/displacement.py
```diff
@@ -115,30 +115,31 @@ def __init__(
 
         self.start = start
 
-        if modal_locations:
-            if isinstance(modal_locations, ModalLocation):
-                hl = modal_locations
+        if reference_location:
```
maxalbert: Since `reference_location` is now mandatory, there is no need for this if statement any more. I'd also remove the commented else branch below please.

flowmachine/tests/test_displacement.py
```diff
@@ -26,54 +26,75 @@ def test_returns_expected_values(stat, sub_a_expected, sub_b_expected, get_dataf
     Test that we get expected return values for the various statistics
     """
     sub_a_id, sub_b_id = "j6QYNbMJgAwlVORP", "NG1km5NzBg5JD8nj"
+    rl = ModalLocation(
```
maxalbert: It's fine to use the original `ModalLocation` here. But now that this is lifted to the level of the tests (rather than implicitly created inside the `Displacement` class) it makes the tests _slightly_ more awkward to read than necessary.

Just for readability, let's switch this reference location to a plain `daily_location`, e.g.
```
rl = daily_location("2016-01-01")
```
This will require changing the expected values in the `pytest.mark.parametrize` statement above.

In order to determine the correct values, it may be helpful to run the test once from the command and grep for the various lines that report the failure and the mismatched values so that you can simply copy and paste them into the code. For example:
```
$ pipenv run pytest -sv tests/test_displacement.py -k test_returns_expected_values 2>&1 | egrep "(test_returns_expected_values\[.*\]|assert)"
__________ test_returns_expected_values[median-130.327306-151.547503] __________
>       assert df.loc[sub_a_id].value == pytest.approx(sub_a_expected)
E       assert 398.64699488375 == 130.327306 ± 1.3e-04
__________ test_returns_expected_values[stddev-132.73844-131.026341] ___________
>       assert df.loc[sub_a_id].value == pytest.approx(sub_a_expected)
E       assert 182.885304966881 == 132.73844 ± 1.3e-04
____________ test_returns_expected_values[avg-176.90362-178.910994] ____________
>       assert df.loc[sub_a_id].value == pytest.approx(sub_a_expected)
E       assert 370.301009034846 == 176.90362 ± 1.8e-04
___________ test_returns_expected_values[sum-5307.1086-5188.418836] ____________
>       assert df.loc[sub_a_id].value == pytest.approx(sub_a_expected)
E       assert 11109.0302710454 == 5307.1086 ± 5.3e-03
___ test_returns_expected_values[variance-17619494.3698313-17167902.1823453] ___
>       assert df.loc[sub_a_id].value == pytest.approx(sub_a_expected)
E       assert 33447034.7728291 == 17619494.3698313 ± 1.8e+01
```

I also suggest doing this for the other tests below that use the original `ModalLocation` definition.

flowmachine/tests/test_displacement.py
```diff
@@ -88,24 +109,28 @@ def test_error_when_modal_location_not_lon_lat():
     )
 
     with pytest.raises(ValueError):
-        Displacement("2016-01-01", "2016-01-02", modal_locations=ml, statistic="avg")
+        Displacement("2016-01-01", "2016-01-02", reference_location=ml, value="avg")
 
 
-def test_error_when_not_modal_location():
+def test_error_when_not_reference_location():
```
maxalbert:
```suggestion
def test_error_when_reference_location_is_not_a_base_location():
```

flowmachine/tests/test_displacement.py
```diff
@@ -88,24 +109,28 @@ def test_error_when_modal_location_not_lon_lat():
...
     """
-    Test that error is raised if modal_locations is not a ModalLocation.
+    Test that error is raised if reference_locations is not a object of type `BaseLocation`.
```
maxalbert:
```suggestion
    Test that error is raised if reference_locations is not an object of type `BaseLocation`.
```

PR #1063, opened 2019-07-10
Allow specifying only a subset of CDR sections in FlowETL config

### LINKED ISSUE #1062: FlowETL config should not require subsections for all CDR types (only relevant ones)
Currently when FlowETL verifies the `config.yml` file it checks whether a subsection is present for each supported CDR type. However, in some scenarios only some of these types are relevant (e.g. if only calls and sms are to be ingested). In this case it should be sufficient to only specify the relevant sections.

maxalbert:
Closes #1062 

### I have:

### DIFF
--- a/flowetl/etl/etl/config_parser.py
+++ b/flowetl/etl/etl/config_parser.py
@@ -40,7 +40,7 @@ def validate_config(*, global_config_dict: dict) -> Exception:
     if set(etl_keys).issubset(CDRType):
         exceptions.append(
             ValueError(
-                f"etl sections present in config.yml must be a subset of {[x.value for x in CDRType]}"
+                f"Etl sections present in config.yml must be a subset of {[x.value for x in CDRType]}. Got: {set(etl_keys)}"
             )
         )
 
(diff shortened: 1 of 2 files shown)


flowetl/etl/etl/config_parser.py
```diff
@@ -36,12 +36,6 @@ def validate_config(*, global_config_dict: dict) -> Exception:
             ValueError("default_args must be a toplevel key in the config file")
         )
 
-    etl_keys = global_config_dict.get("etl", {}).keys()
-    if etl_keys != CDRType._value2member_map_.keys():
```
danwilliams: Just to check, are the keys still a subset of the CDR_TYPE possible? Would suggest checking for something like:
```set(etl_keys).issubset(set(CDRType._value2member_map_.keys()))```

maxalbert: Yes, that's a great point. Indeed, they are still just a subset. This check is implicitly done elsewhere in the code (where we convert them via `CDRType(cdr_type)`, which will fail if that condition is not met) but good to have an explicit check here, too. 👍

PR #1099, opened 2019-07-28
Flowmachine histogram aggregation

### LINKED ISSUE #1075: Histogram aggregation of metrics in flowmachine
The `SpatialAggregate` class in flowmachine allows to aggregate metrics spatially - e.g. going from a location associated with each subscriber (such as `daily_locations`) to a subscriber count per admin region.

However, some metrics (e.g. subscriber-focused ones) don't allow intrinsic spatial aggregation. For example, `Displacement` or `RadiusOfGyration` associate a number (representing the distance of displacement / radius of gyration) with each subscriber but there is no way to intrinsically aggregate this to spatial units (without assigning a "reference location" for each subscriber - which is what `JoinedSpatialAggregate` allows to do).

For metrics of the latter kind it would be useful to support a histogram aggregation. This should accept a metric (such as `Displacement`, `RadiusOfGyration`, etc.) and the `bins` and `range` parameters as supported by [numpy.histogram](https://docs.scipy.org/doc/numpy/reference/generated/numpy.histogram.html). The return value should be the same pair `hist_values`, `bin_edges` as returned by `numpy.histogram`. (This can then separately be fed into a plotting function or otherwise processed.)

For now, as a first prototype it should be ok to do the calculation in Python, although this won't work for large data amounts and we will soon want to push the calculation into Postgres (but let's address this in a separate issue). Also, we will want to expose this functionality through the API, which will also be tracked in a separate issue.

BhavinPanch:
Closes #1075

### I have:

### Description

Adds a new, non-spatial, aggregation that can be applied to other queries: `HistogramAggregation`!

Behaves very similarly to numpy's `histogram` function, but:
- Returns the upper and lower edge for each bin in separate dataframe columns
- By default, returns an empty dataframe if one bin has values <15
- Doesn't support the automatic bin calculations at this time

In implementing this, I've decided _not_ to use `width_bucket` Postgres-side, because it only supports integer values for defined bins.


flowmachine/tests/test_histogram_aggregations.py
```diff
@@ -0,0 +1,29 @@
...
+def test_can_be_aggregated_admin3_distribution(get_dataframe):
+    """
+    
+    """
+    RoG = RadiusOfGyration("2016-01-01", "2016-01-02")
+
+    agg = HistogramAggregation(locations=RoG, bins=5)
+    df = get_dataframe(agg)
+    print("==================================")
+    print(np.histogram(df["value"], [5]))
```
greenape: Presumably what you want to do here is compare `np.histogram(get_dataframe(RoG).value, [5])` with what's in df?

flowmachine/flowmachine/features/utilities/histogram_aggregations.py
```diff
@@ -0,0 +1,49 @@
...
+        return ["value", "bins"]
+
+    def _make_query(self):
+        if isinstance(self.ranges, tuple):
+            max_range = max(self.ranges)
+            min_range = min(self.ranges)
+            sql = f"""
+            SELECT
+                value,
+                width_bucket(value,{max_range},{min_range},{self.bins}) as bins
```
greenape: Don't think this is quite right - what we want to get is a count of the number of values in each bucket, whereas this is gonna give you all the values and their bucket. So, you need a. an additional aggregation step before returning which groups by bucket number, and b. to calculate the buckets :)

flowmachine/tests/test_histogram_aggregations.py
```diff
@@ -0,0 +1,53 @@
...
+def test_create_histogram_using_list_of_bins_values(get_dataframe):
+    """
+    Create histogram using list of bins values.
+    """
+    RoG = RadiusOfGyration("2016-01-01", "2016-01-02")
+
+    agg = HistogramAggregation(locations=RoG, bins=[10, 20, 30, 40, 50, 60])
+    df = get_dataframe(agg)
+    de = np.histogram(get_dataframe(RoG).value, bins=[10, 20, 30, 40, 50, 60])
+    assert set(de[0].tolist()) == set(df.value.tolist())
```
greenape: Half way there :) Now just need to return the bin edges as in the numpy function, rather than the bucket number.

flowmachine/flowmachine/features/utilities/histogram_aggregations.py
```diff
@@ -0,0 +1,55 @@
...
+from ...core.query import Query
+from typing import List
+
+
+class HistogramAggregation(Query):
+    """ 
+
+    """
+
+    def __init__(self, *, locations, bins, ranges: tuple = None):
```
greenape: Can we change locations to metric or similar? This is for use with metrics, rather than locations.

flowmachine/flowmachine/features/utilities/histogram_aggregations.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def _make_query(self):
+        if isinstance(self.ranges, tuple) and not None:
+            max_range = max(self.ranges)
+            min_range = min(self.ranges)
+        elif self.ranges is None:
+            max_range = f"""select max(value) from ({self.locations.get_query()}) as to_agg """
+            min_range = f""" select min(value) from ({self.locations.get_query()}) as to_agg """
+        else:
+            raise ValueError("Range should be tuple of two values")
```
greenape: Raise ValueErrors in the constructor, because otherwise you've got no indication the thing won't run until you try to run it.

flowmachine/flowmachine/features/utilities/histogram_aggregations.py
```diff
@@ -0,0 +1,55 @@
...
+
+        filter_values = f"""select value from ({self.locations.get_query()}) AS to_agg
```
greenape: Can we uppercase the sql keywords pls.

flowmachine/flowmachine/features/utilities/histogram_aggregations.py
```diff
@@ -0,0 +1,55 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from ...core.query import Query
+from typing import List
+
+
+class HistogramAggregation(Query):
+    """ 
+
```
greenape: Needs a docstring obviously.

flowmachine/flowmachine/features/utilities/histogram_aggregations.py
greenape:
```suggestion
    def __init__(self, *, locations: "Query", bins: Union[List[float], int], ranges: Optional[Tuple[float, float]] = None) -> None:
```

flowmachine/flowmachine/features/utilities/histogram_aggregations.py
```diff
@@ -0,0 +1,55 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from ...core.query import Query
+from typing import List
```
greenape:
```suggestion
from typing import List, Tuple, Optional, Union
if TYPE_CHECKING:
    from .query import Query
```

flowmachine/flowmachine/features/utilities/histogram_aggregations.py
```diff
@@ -0,0 +1,55 @@
...
+                    """
+        elif isinstance(self.bins, list):
+
+            sql = f"""
+                select count(foo.value) as value, width_bucket(foo.value::numeric,Array{self.bins}::numeric[]) as bin_edges 
+                from ({filter_values}) AS foo
+                group by bin_edges
+                """
+        else:
+            raise ValueError("Bins should be integer or list of integer values.")
```
greenape:
```suggestion
            raise ValueError("Bins should be an integer or list of numeric values.")
```

flowmachine/tests/test_histogram_aggregations.py
```diff
@@ -0,0 +1,53 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from flowmachine.core import make_spatial_unit
+from flowmachine.features import (
```
greenape: Quite a few redundant imports here.

flowmachine/flowmachine/features/utilities/histogram_aggregations.py
```diff
@@ -0,0 +1,55 @@
...
+            min_range = min(self.ranges)
+        elif self.ranges is None:
+            max_range = f"""select max(value) from ({self.locations.get_query()}) as to_agg """
+            min_range = f""" select min(value) from ({self.locations.get_query()}) as to_agg """
+        else:
+            raise ValueError("Range should be tuple of two values")
+
+        filter_values = f"""select value from ({self.locations.get_query()}) AS to_agg
+                        where value between ({min_range}) and ({max_range})  """
+        if isinstance(self.bins, int):
```
greenape: Could use https://docs.python.org/3.8/library/functools.html#functools.singledispatchmethod to split this out into two smaller functions based on the types.

flowmachine/flowmachine/features/utilities/histogram_aggregations.py
greenape: (And put it in the constructor)

flowmachine/flowmachine/features/utilities/histogram_aggregation.py
```diff
@@ -0,0 +1,156 @@
...
+    3    108  212.513152  283.350869
+    4     15  283.350869  354.188587
+    """
+
+    def __init__(
+        self,
+        *,
+        metric: "Query",
+        bins: Union[List[float], int],
+        range: Optional[Tuple[float, float]] = None,
```
jch1g10: Probably not a good idea to use a keyword as an argument name

flowmachine/flowmachine/features/utilities/histogram_aggregation.py
```diff
@@ -0,0 +1,156 @@
...
+    Parameters
+    ----------
+    metric : Query
+        Query to build histogram over
+    bins : int, or list of float
+        Either an integer number of equally spaced bins, or a list of bin edges
+    range : tuple of float, default None
+        Optionally supply inclusive lower and upper bounds to build the histogram over. By default, the
+        histogram will cover the whole range of the data.
+    censor : bool, default True
```
jch1g10: Missing a description of the 'value_column' parameter

flowmachine/flowmachine/features/utilities/histogram_aggregation.py
```diff
@@ -0,0 +1,156 @@
...
+        self.range = range
+        self.value_column = value_column
+        self.censor = censor
+        if self.value_column not in self.metric.column_names:
+            raise ValueError(
+                f"'{self.value_column}' is not a column in this query. Must be one of '{self.metric.column_names}'"
+            )
+        if not isinstance(self.bins, (int, list)):
+            raise ValueError("Bins should be an integer or list of numeric values.")
+        if not isinstance(self.range, tuple) and None:
```
jch1g10: I don't think I understand this line

flowmachine/flowmachine/features/utilities/histogram_aggregation.py
jch1g10: Or rather, a Python built-in function name

greenape: Yeah, I agree - but following the numpy function signature: https://docs.scipy.org/doc/numpy/reference/generated/numpy.histogram.html

jch1g10: Ah, I see. Fair enough.

PR #1103, opened 2019-07-29
Fix flowauth stackfile

greenape:
Updates Flowauth's compose file to a stack file, fixes the secrets Quickstart docker stack invocation, and fixes Flowauth requiring that the private key be provided as an env var even if you've provided a secret.


jch1g10:
When I spin up using the `secrets_quickstart.sh` script and then try to log into FlowAuth, I get an "Error: Bad Gateway" dialog.

greenape:
Did you set the container tag env var?

PR #1141, opened 2019-08-09
Catch all errors in get_reply_for_message

### LINKED ISSUE #1140: Pass all FlowMachine errors back to FlowAPI
There are still some paths in the FlowMachine server through which errors can remain uncaught, leaving the API hanging indefinitely with no response. A generic `except Exception` in `get_reply_for_message` would catch any exceptions not handled elsewhere, and could ensure errors always get back to the API.

jch1g10:
Closes #1140 

### I have:

### Description

Adds another `except` at the end of `get_reply_for_message` which catches all exceptions and passes the error message back to FlowAPI before re-raising.

Also adds another try/except in FlowAPI to handle another edge case in `get_query_result`, which previously assumed all error responses would have a `"query_state"` in the response payload.


flowmachine/flowmachine/core/server/server.py
```diff
@@ -79,6 +79,10 @@ def get_reply_for_message(msg_str: str) -> ZMQReply:
         return ZMQReply(
             status="error", msg="Invalid JSON.", payload={"decode_error": exc.msg}
         )
+    except Exception as exc:
+        # Generic error boundary to catch any unexpected errors and pass to API
+        return ZMQReply(status="error", msg=f"{exc}")
+        raise exc
```
jch1g10: This line, of course, is pointless since it happens after `return`. Need to rethink this a bit.

greenape: Hmm. Well, obvs just remove the unreachable return?

I think my issue with this is you don't really want to be returning uncensored exceptions from a pokeable service, right? Plus, we'd want to be able disambiguate a 400 (you screwed up) from a 500 (we screwed up) so we can return a correct status code from the API.

jch1g10: Yes, perhaps we just want to send back a "something went wrong" message in this case, rather than the actual error message. My aim is to make sure we return _something_,  because at the moment if FlowMachine raises an error that the server isn't expecting it will just leave the API hanging, so from the user's perspective it's not obvious what's happened.

jch1g10: As for the `raise` after `return`, I was thinking if this is an unexpected error we might actually want the server to die (after sending an error response to the API). Also, if we don't re-raise the error then we don't get a traceback in the FlowMachine logs (although I would imagine there are other ways around this).

jch1g10: And I would imagine that if we're handling a generic exception that hasn't been caught by any of the more specific exception handlers then it's probably a 500 error (we should be able to predict what 400 errors a user could trigger, and handle those appropriately).

greenape: Mm. I'm down with returning a 500, just not without sanitising content. In this case, can we return a nondescript error and add a todo to disambiguate/properly code errors from flowapi?

As to the traceback, how about using the `__traceback__` attribute of the exception object?

greenape:
```python
import traceback
try:
    assert 1 == 2
except Exception as exc:
    trace = traceback.format_list(traceback.extract_tb(exc.__traceback__))
```

flowmachine/flowmachine/core/server/server.py
```diff
@@ -154,7 +155,17 @@ def get_reply_for_message(msg_str: str) -> ZMQReply:
     msg_contents : str
         JSON string with the message contents.
     """
-    reply_json = get_reply_for_message(msg_contents)
+    try:
+        reply_json = get_reply_for_message(msg_contents)
+    except Exception as exc:
+        # Catch and log any unhandled errors, and send a generic error response to the API
+        # TODO: Ensure that FlowAPI always returns the correct error code when receiving an error reply
+        logger.debug(
```
greenape: I think this should be error level rather than debug.

flowapi/tests/unit/conftest.py
```diff
@@ -16,13 +16,24 @@
 from asyncio import Future
 
 
-def async_return(result):
+@pytest.fixture
+def async_return():
```
greenape: Feels a bit unnecessary to have this as a fixture to me? Also I think the one time I see it used would also be achievable with https://asynctest.readthedocs.io/en/latest/asynctest.mock.html#asynctest.CoroutineMock?

integration_tests/tests/flowmachine_server_tests/test_action_run_query.py
```diff
@@ -171,7 +171,6 @@
     assert expected_error_messages == reply["payload"]["validation_error_messages"]
 
 
-@pytest.mark.skip(reason="Cannot currently test this because the sender hangs")
```
greenape: 🥇

PR #1168, opened 2019-08-16
Expose random sampling via API

### LINKED ISSUE #1007: Expose random sampling via api
I think with random sampling, while this is actually internally a distinct query class, it makes most sense to squash it into the query-to-be-sampled's params in the API, so it is always a modifier on another query.

e.g.
```json
{
        "query_kind": "subscriber_degree",
        ...
        "sampling": {...}
    }
```

So, a schema that captures `Random(..)`, and then post-load this is converted to the actual Random call.

jch1g10:
Closes #1007 

### I have:

### Description

Adds a `"sampling"` parameter to all query schemas except top-level aggregate queries.
The random sample parameter validation is handled by a `RandomSampleSchema`, which post-load returns a `RandomSampler` object with a `make_random_sample_object` method. The `make_random_sample_object` method takes a Query object and calls its `random_sample` method with the appropriate parameters.

Adds a corresponding `random_sample` function to FlowClient, to generate a random sample spec.

Other changes:
- Splits the `RandomBase` class into multiple classes for the different sampling methods, to simplify argument checking and avoid having a large if/else in `_make_query`.
- Modifies the FlowDB `random_ints` function, so that it always returns exactly the requested number of random ints.
- Removes the redundant `Query._db_store_cache_metadata` method.

Note: While it wouldn't make sense to expose a random sampling parameter for top-level exposed queries anyway (since these are all aggregates), it's worth noting that doing so would not be possible at the moment because un-seeded random samples cannot be stored, and so their results could not be retrieved by the API.


flowclient/flowclient/client.py
```diff
@@ -1654,15 +1652,17 @@ def random_sample(
     Returns
     -------
     dict
-        Dict which specifies the parameters for the random sample.
+        Dict which functions as the query specification.
     """
-    spec = {
+    sampled_query = query.copy()
```
greenape: Slightly more pythonic to do `dict(query)`, although obviously in both cases any nested dicts still refer to their underlying object.

jch1g10: Ah, good idea. I'm hoping a shallow copy is sufficient since `random_sample` doesn't modify any existing values in `query`, just adds a new one.

flowmachine/flowmachine/core/random.py
```diff
@@ -3,293 +3,475 @@
...
+    def _make_query(self):
+        raise NotImplementedError(
+            f"Class {self.__class__.__name__} does not implement the _make_query method."
+        )
+
+    # Overwrite the table_name method so that it cannot
+    # be stored by accident.
+    @property
+    def table_name(self):
+        if hasattr(self, "seed") and self.seed is not None:
```
greenape: Wondering if, rather than needing to check the logic in a bunch of places, it would be nicer to have a `SeededRandom` class to mix in, or `SeededRandom(RandomBase, metaclass=ABCMeta)` to use as another step in inheritance chain?

flowmachine/flowmachine/core/random.py
```diff
@@ -3,293 +3,497 @@
...
+    @property
+    def seed(self):
+        return self._seed
+
+    @property
+    def _sample_params(self):
+        """
+        Parameters passed when initialising this query.
+        """
+        return {
```
greenape: How about:

```python
return dict(seed=self.seed, **super()._sample_params)
```

flowmachine/flowmachine/core/random.py
```diff
@@ -3,293 +3,497 @@
...
+            size = self.size
+
+        sampled_query = f"""
+        SELECT {columns} FROM {self.query.fully_qualified_table_name} TABLESAMPLE SYSTEM_ROWS({size})
+        """
+
+        return sampled_query
+
+
+class SeededRandom(RandomBase, metaclass=ABCMeta):
```
greenape: Let's change this to `SeedableRandom`, because there's no guarantee it is _actually_ seeded.

flowmachine/flowmachine/core/random.py
```diff
@@ -3,293 +3,488 @@
...
+    def __call__(self, query, sampling_method, params):
+        return query.random_sample(sampling_method, **params)
+
+
+class RandomBase(metaclass=ABCMeta):
+    """
+    Base class for queries used to obtain a random sample from a table.
+    """
+
+    def __init__(self, query, *, size=None, fraction=None, estimate_count=True):
```
greenape: Type annotations while we're here.

PR #1229, opened 2019-09-02
Dogpile cache

### LINKED ISSUE #1173: And i also found another issue with insert `two_factor_auth` to database
And i also found another issue with insert `two_factor_auth` to database

This code block in `user_setting.py`

```python
    auth = TwoFactorAuth(user=current_user)
    auth.secret_key = secret

    auth.validate(code)
    auth.enabled = True
    db.session.add(auth)
```

and `validate()` function declared:
```python
      def validate(code: str) -> bool:
        if is_valid:
            if (
                self.last_used_two_factor_code == code
            ):  # Reject if the code is being reused
                raise Unauthorized("Code not valid.")
            else:
                self.last_used_two_factor_code = code
            db.session.add(self)
            db.session.commit()
            return True
        else:
            raise Unauthorized("Code not valid.")
```

So when execute. At beginning when user try to confirm two_factor, there is a database insertion conflict.
`auth` object already inserted to database during validate. Then after that when we try to update auth enabled attribute and insert again:
```python
 auth.validate(code)
 auth.enabled = True
 db.session.add(auth)
```
That will throw error. because auth object already gone

_Originally posted by @hprobotic in https://github.com/Flowminder/FlowKit/issues/1170#issuecomment-522508877_

greenape:
Closes #1173

### I have:

### Description

Switches to using dogpilecache to hold last used 2factor codes for Flowauth rather than storing them in the db. This allows the last used codes to expire, and means we avoid a hidden commit of the db session when verifying them.
Can either use a file backend, or a Redis backend.

Making this WIP as much as anything because I can't quite see where the docs for it should live.


greenape:
Added some docs, let's find 20 mins to walk @jch1g10 and @maxalbert through this one.

flowauth/backend/flowauth/models.py
```diff
@@ -271,13 +270,14 @@ def validate(self, code: str) -> bool:
         is_valid = pyotp.totp.TOTP(self.decrypted_secret_key).verify(code)
         if is_valid:
             if (
-                self.last_used_two_factor_code == code
+                current_app.config["CACHE_BACKEND"].get(f"{self.user_id}".encode())
```
jch1g10: Can there ever be more than one valid code at a time (excluding backup codes)? If so, then codes could be re-used by alternating between two valid codes, since this only checks the last-used code. But as long as there's only one valid code at a time then I think this is fine.

flowauth/frontend/cypress/integration/two_factor_auth.js
```diff
@@ -45,15 +45,24 @@ describe("Two-factor setup", function() {
...
                 cy.contains("My Servers").should("be.visible");
+                cy.get("#logout").click();
+                cy.get("#username")
+                  .type(username)
+                  .get("#password")
+                  .type(password)
+                  .get("button")
+                  .click()
+                  .get("#two_factor_code")
+                  .type(mfaCode.stdout) // Log in using backup code because we don't want to wait for a new mfa code
```
jch1g10: The comment here confused me. I assume this was just copy-pasted from below?

flowauth/backend/flowauth/models.py
jch1g10: We don't change the default `valid_window=0` when `verify`ing the code, so there's only one valid code at a time.

PR #1232, opened 2019-09-03
WIP: Tasks/improve synthetic data generation

danwilliams:
This introduces a new approach to synthetic data generation, which is more realistic in approach.


greenape:
Can we:

1. Get the CI suite to pass (looks like the issue is that this is behind master and there's some weird graphviz issue)
1. Run the `synth_data_*_generator_tests` job for this data generator as well, so it can be seen to pass the very minimal number of tests there?

greenape:
Ok, as of 52420e1 this needs to be rebased on the revised schema branch. Or, needs a branch off 
159cd55 to be put into a separate PR and merged, then the schema branch to be merged, then we can work on this.

Issue #1231, opened 2019-09-02
FlowDB build scripts (.sql and .sh) should live in the same folder


maxalbert:
FlowDB has a number of build scripts which currently mainly live in two folders:
- `flowdb/bin/build/*.sh` (bash scripts)
- `flowdb/sql/*.sql` (sql scripts)

These are numbered sequentially to ensure order of execution:
```
$ ls flowdb/bin/build/
001_check_environment.sh
002_tweak_config.sh
003_create_extensions.sh
004_postgis_ingest_data.sh
005_flowdb_version.sh
006_location_table.sh
08_configure_cache.sh
configurate.py
last_create_roles.sh
pg_config_template.conf

$ ls flowdb/sql/
00_schema_public.sql
01_schema_infrastructure.sql
02_schema_events.sql
03_schema_geography.sql
04_schema_other.sql
05_add_data.sql
06_available_tables.sql
07_schema_cache.sql
08_schema_dfs.sql
functions_000_postgis_vit_utilities.sql
functions_001_utilities.sql
functions_002_hartigan.sql
functions_003_grid.sql
functions_004_viewshed.sql
```

However, the bash and sql scripts are all copied into the same location in side the flowdb image (namely, `/docker-entrypoint-initdb.d/`). Moreover, the startup script `docker-entrypoint.sh` script (which is part of the official `postgres` docker image and runs when the container first starts up) simply [executes](https://github.com/docker-library/postgres/blob/master/docker-entrypoint.sh#L145) all files in order, no matter whether it's a bash or sql script (the filetype only determines _how_ the script gets executed, not at which point in time).

This is certainly not obvious from the flowdb folder structure and can lead to confusion. For example, the file `flowdb/build/08_configure_cache.sh` deliberately deviates from the numbering scheme of all the other `.sh` scripts (which all start with three-digit numbers) because it needs to be executed _after_ the `.sql` scripts.

I suggest to combine all `.sh` and `.sql` scripts in the same folder and simply number them sequentially in the order in which they are actually executed (with some gaps so that other scripts can be added in between if necessary).

PR #1255, opened 2019-09-09
Revised flowdb build scripts

maxalbert:
Towards #1231.

### I have:

### Description

This PR provides a tidy-up of the existing build scripts and adds support for running Alembic migrations as part of the flowdb startup.

At the moment it doesn't _replace_ any of the existing build scripts - all of these are still in place, so during startup it will still create the full set of tables as before. Instead, it creates a separate database called `flowdb_revised_schema` (in addition to the existing one called `flowdb`) against which any new build scripts and migrations are run. This is intended to be a temporary arrangement while the FlowDB schema revision is ongoing - eventually the new build scripts will supersede the existing ones.

The changes in more detail:

- Added a new folder `flowdb/build_scripts/`, which is intended to contain all build scripts for FlowDB (see [#1231](https://github.com/Flowminder/FlowKit/issues/1231)).

- The existing bash build scripts (currently in `flowdb/bin/build/*.sh`) have been copied over to `flowdb/build_scripts/`, and a few things which are not used have been tidied up/removed. In particular:
    - Removed everything related to the `template_postgis` database.
    - I didn't copy the script `bin/build/004_postgis_ingest_data.sh` because it attempts to run files found in the folder `/docker-entrypoint-initdb.d/data/sql/` inside the Docker container, but this folder never gets created so the script is effectively a no-op.
    - Moved the setting of `PGUSER=$POSTGRES_USER` to the Dockerfile, instead of repeating it in most build scripts.
    - A couple of other minor tweaks.

- A skeleton for Alembic migrations has been put in place, which will help with incremental changes to the FlowDB tables in the future and with upgrades between FlowDB versions. The Alembic configuration lives in `flowdb/alembic.ini` and the actual migration scripts will live in `flowdb/migrations/versions/` (currently this folder is empty). The migrations are run via the script `flowdb/build_scripts/0900_run_database_migrations.sh` during startup of the docker container (this basically just runs the commend `alembic upgrade head`).

  The connection to flowdb is controlled via the following env vars: `ALEMBIC_FLOWDB_USER`, `ALEMBIC_FLOWDB_PASSWORD`, `ALEMBIC_FLOWDB_DATABASE`. Note that currently these need to connect as the `flowdb` user, or some kind of alembic user needs to be created separately.


.circleci/config.yml
```diff
@@ -18,6 +18,9 @@ defaults:
     FLOWAPI_FLOWDB_USER: flowapi
     FLOWMACHINE_FLOWDB_PASSWORD: foo
     FLOWAPI_FLOWDB_PASSWORD: foo
+    ALEMBIC_FLOWDB_USER: flowdb
```
greenape: These'll need to be passed into the FlowDB spinups that happen in the flowdb build job I think.

Issue #1256, opened 2019-09-10
Server takes too long to respond


jc-harrison:
The first time an `EventsTablesUnion` object is constructed, all events tables for all available dates are "added to cache" (i.e. the query IDs for the corresponding `Table` objects are added to `cache.cached`, and the query state is set to `completed` in redis). This process can take several seconds.

As query object construction in the FlowMachine server happens on the main thread, before creating a separate thread to run the query, this is a blocking operation and causes the server to wait several seconds before responding. In some cases (e.g when building the docs without running the pre-cache script) this wait is too long and results in FlowClient raising:
```
FlowclientConnectionError: Unable to connect to FlowKit API at http://localhost:9090: ('Connection aborted.', RemoteDisconnected('Remote end closed connection without response'))
```

greenape:
This is actually pretty serious I think - Hypercorn's default timeout is, I think, 5s.

At present, while the flowmachine server is async it largely operates serially - so that means that if your request is in a queue which is going to take more than 5s to get through, you'll error out.

In principle, adding a `sleep(4)` inside `Query`'s `__init__` for example, and then firing off two distinct run query requests _should_ make the second fail, and will _definitely_ make any third request fail.

greenape:
So that gives us two issues: 

1. Query object creation can be pretty slow
2. Even if query object creation is pretty fast, it blocks the server from responding, so even if it took 1s on average you'd hit this issue with quite a small number of unique run query requests close together.

PR #1258, opened 2019-09-11
Speed

greenape:
Towards #1256 but doesn't _solve_ it.

### I have:

### Description

Several small tweaks which speed up creating query objects.

- Runs `available_dates` when first connecting to FlowDB to pre-create `Table` objects
- Runs `available_dates` periodically as a background thread in the server to keep that fresh
- Skips creating the state machine transitions if the state machine already exists
- Skips fast forwarding `Table` state machines if they're already complete
- Skips dumping a pickle if the object is already in cache
- Removes an inline import of `__version__` in `write_cache_metadata` - originally necessary to avoid a circular import, but not now
- Removes a call to `pg_notify` that isn't used

### DIFF
--- a/flowapi/tests/unit/test_access_control.py
+++ b/flowapi/tests/unit/test_access_control.py
@@ -24,7 +24,7 @@ async def test_protected_get_routes(route, app):
     response = await app.client.get(route)
     assert 401 == response.status_code
 
-    log_lines = log_lines = app.log_capture().access
+    log_lines = app.log_capture().access
     assert 1 == len(log_lines)  # One entry written to stdout
 
     assert "UNAUTHORISED" == log_lines[0]["event"]
--- a/flowapi/tests/unit/test_access_control.py
+++ b/flowapi/tests/unit/test_access_control.py
@@ -16,7 +16,7 @@ async def test_protected_get_routes(route, app):
     Parameters
     ----------
     app: tuple
-        Pytest fixture providing the flowapi, with a mock for the db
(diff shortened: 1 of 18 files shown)


flowmachine/flowmachine/core/connection.py
```diff
@@ -275,9 +275,6 @@ def _known_dates(self, table, schema):
             if x.isnumeric()
         )
 
-    @cached(
-        TTLCache(1024, 120)
-    )  # Many dates to cache, two minutes seems reasonable to balance db access against speed boost
```
jch1g10: Just for my understanding, what's the reason for removing this?

flowapi/tests/unit/test_access_control.py
```diff
@@ -9,25 +9,23 @@
...
         Route to test
     """
-    client, db, log_dir, app = app
 
-    response = await client.get(route)
+    response = await app.client.get(route)
     assert 401 == response.status_code
 
-    log_lines = json_log().out
+    log_lines = log_lines = app.log_capture().access
```
jch1g10:
```suggestion
    log_lines = app.log_capture().access
```

flowmachine/flowmachine/core/connection.py
greenape: Just to record the reasoning - this function is only called by `available_dates`, which is itself cached. The periodic update function added to server.py skips the cache for `available_dates` so it will always hit the db and create any table objects needed. That isn't guaranteed if this method is _also_ cached.

PR #1266, opened 2019-09-12
Calldays column names

### LINKED ISSUE #1265: IntereventPeriod can't be subset
`IntereventPeriod` returns time intervals. Those can't be subset using either of the subsetters available. That renders it useless for identifying subsets of subscribers based on call day gap behaviour.

greenape:
Closes #1265

### I have:

### Description

- Renames `IntereventPeriod` to `IntereventInterval`
- Adds a new `IntereventPeriod` class which returns intervals as fractional time units (1.5 hours, for example), meaning it can be subset.

(I've wavered a little over whether `IntereventPeriod` should take the args to `IntereventInterval`, or the object and gone with the former.)


greenape:
> I'm wondering whether the nested query in the `IntereventPeriod` class will slow things down, because it first calculates the internal `IntereventInterval` table and then needs to go over it again to divide by the `time_divisor`? It may not make a difference at all compared to doing it in a single pass over the data, but if you happen to have any benchmarks I'd be interested to see if there is any performance aspect.

I suspect it _will_ make a difference and will be marginally slower - I'd almost toss the Interval variant, but for the fact that if you cache the interval variant you get all of the time unit ones effectively 'free'.

maxalbert:
> > I'm wondering whether the nested query in the `IntereventPeriod` class will slow things down, because it first calculates the internal `IntereventInterval` table and then needs to go over it again to divide by the `time_divisor`? It may not make a difference at all compared to doing it in a single pass over the data, but if you happen to have any benchmarks I'd be interested to see if there is any performance aspect.
> 
> I suspect it _will_ make a difference and will be marginally slower - I'd almost toss the Interval variant, but for the fact that if you cache the interval variant you get all of the time unit ones effectively 'free'.

Fair enough. Let's keep it as it is, happy to merge this now (just needs resolving the merge conflict in the changelog).

maxalbert:
(I took the liberty to resolve the changelog conflict here on Github when merging in current master.)

PR #1272, opened 2019-09-16
More useful test output for flowmachine tests, more durations

greenape:
Changes the flowmachine test output to make it easier to find _which_ test actually failed, and increases the number of longest test durations shown.

### DIFF
--- a/flowmachine/tests/conftest.py
+++ b/flowmachine/tests/conftest.py
@@ -68,34 +68,12 @@ def exemplar_spatial_unit_param(request):
     yield make_spatial_unit(**request.param)
 
 
-def get_string_with_test_parameter_values(item):
-    """
-    If `item` corresponds to a parametrized pytest test, return a string
-    containing the parameter values. Otherwise return an empty string.
-    """
-    if "parametrize" in item.keywords:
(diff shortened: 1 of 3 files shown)


flowmachine/tests/conftest.py
```diff
@@ -98,8 +98,7 @@ def pytest_itemcollected(item):
     and also adds info about any parameters (if the test is parametrized).
     """
     if item.obj.__doc__:
-        item._nodeid = "* " + " ".join(item.obj.__doc__.split())
-        item._nodeid += get_string_with_test_parameter_values(item)
```
maxalbert: Did you intentionally drop the info about parameter values? I think I found this useful in the past to debug parametrised tests where it only fails for certain values.

But if you do want to drop it then maybe we should remove the function definition for`get_string_with_test_parameter_values()` because it's now not used anywhere any more?

Also, the exact same code snippet also lives in `flowetl/tests/conftest.py`, so would be worth porting your improvements to that conftest as well.

greenape: Pytest actually includes the params in the report for failed tests anyway :)

PR #1273, opened 2019-09-16
Random safety

### LINKED ISSUE #1263: Disallow unseedable random sample methods through API
Only the random sampling methods which can be seeded should be available via the API, because things returned by the API are cached.

And with this, we should change the default random sampling method to a seedable one.

greenape:
Closes #1263

### I have:

### Description

Removes system_rows from FlowAPI, makes `random_ids` the default sampling method.

### DIFF
--- a/flowmachine/flowmachine/core/query.py
+++ b/flowmachine/flowmachine/core/query.py
@@ -969,13 +969,13 @@ class Query(metaclass=ABCMeta):
         objs = Query.connection.fetch(qry)
         return (pickle.loads(obj[0]) for obj in objs)
 
-    def random_sample(self, sampling_method="system_rows", **params):
+    def random_sample(self, sampling_method="random_ids", **params):
         """
         Draws a random sample from this query.
 
         Parameters
         ----------
-        sampling_method : {'system', 'system_rows', 'bernoulli', 'random_ids'}, default 'system_rows'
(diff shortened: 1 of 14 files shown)


flowmachine/flowmachine/core/query.py
```diff
@@ -997,7 +997,7 @@ def random_sample(self, sampling_method="system_rows", **params):
         fraction : float, optional
             Fraction of rows to draw.
             Exactly one of the 'size' or 'fraction' arguments must be provided.
-        estimate_count : bool, default True
+        estimate_count : bool, default False
```
jch1g10: What's the reason for changing this default? (Just so we have a record of the decision)

flowmachine/tests/test_query_object_construction.py
```diff
@@ -36,22 +36,6 @@ def test_construct_query(diff_reporter):
...
-                "method": "last",
-                "subscriber_subset": None,
-                "sampling": {
-                    "sampling_method": "system_rows",
-                    "size": 10,
-                    "fraction": None,
-                    "estimate_count": False,
-                },
-            },
-        },
```
jch1g10: I wonder whether it's still worth having this test, but using a different sampling method?

flowmachine/flowmachine/core/query.py
greenape: Actually we could probably revert this, but with estimates on, you potentially under sample the higher end of the table, or return fewer than expected rows. Safer default.

PR #1283, opened 2019-09-18
Bump tha deps

greenape:
_Consumes_ #1281 and #1278


greenape:
The apispec update is blocked by https://github.com/timakro/apispec-oneofschema/issues/5

PR #1284, opened 2019-09-18
Cache dependencies when running queries through FlowAPI

### LINKED ISSUE #1152: Orchestration/scheduler/pre-cache
Because we don't allow any control over cache from FlowAPI, it would be desirable if a level of automatic pre-caching/scheduling/orchestration was applied to queries from that source.

There's quite a long way one can dive into this, but I think an initial approach which unrolls the query graph and does a traversal up it would be acceptable. I'd be inclined to say we want to be a _little_ smart, and at least not trigger caching if there's no direct reason to do so for the outer query (i.e. query-to-return is an aggregate of a modal, modal is cached but daily aren't - right thing to do there is _not_ to cache the daily).

I'd also be inclined to make this a behaviour that can be switched on or off via env var.

jch1g10:
Closes #1152

### I have:

### Description

This pull request adds a new argument `store_dependencies` (default `False`) to the `Query.store` method. Setting  `store_dependencies=True` will cause all the unstored dependencies of a query to be stored before the query itself is stored.

When running queries through the API the Flowmachine server will, by default, use `store(store_dependencies=True)` to execute and store a query. This behaviour can be switched off by setting the environment variable `FLOWMACHINE_SERVER_DISABLE_DEPENDENCY_CACHING=true`, in which case only the top-level query will be cached.

A side-effect of this change is that the pre-cache script is no longer required in the docs build, so I have removed it.

The dependency storing is handled by a new method `Query._store_dependencies_and_make_sql`, which constructs a dependency graph of unstored dependencies and calls `store()` on each of them in an appropriate order, then waits for all of those to finish before returning the output of `self._make_sql`. When calling `store` with `store_dependencies=True`, `_store_dependencies_and_make_sql` is passed instead of `_make_sql` as the `ddl_ops_func` in `write_query_to_cache`.

I initially tried handling the dependency storing in an outer function which then calls `query.store()` after storing all the dependencies, but eventually decided pushing this down to run within `write_query_to_cache` is better because this way the top-level query is in "executing" state from the start, so a different thread can't start storing the same query before all the dependencies have started running.

I've added a function `store_queries_in_order` in `utils.py`, which takes a dependency graph and stores the queries in an order that ensures each query is stored after its dependencies. I've also added a method `Query._unstored_dependencies_graph`, which creates a dependency graph of only the unstored dependencies of a query (and excludes unstored dependencies of stored queries). This is a `Query` method rather than a function in `utils.py` primarily due to import issues with `QueryStateMachine`.


flowmachine/flowmachine/core/model_result.py
```diff
@@ -136,17 +141,29 @@ def to_sql(self, name: str, schema: Union[str, None] = None) -> Future:
...
+            If store_dependencies = True
 
         Notes
         -----
 
         This method will return a Future immediately.
         """
+        if store_dependencies:
+            raise NotImplementedError(
+                "ModelResult does not support storing dependencies."
```
greenape: Why not?

flowmachine/flowmachine/core/query.py
```diff
@@ -112,7 +116,10 @@ def md5(self):
             self._md5 = md5(str(hashes).encode()).hexdigest()
             return self._md5
 
-    query_id = md5  # alias which is more meaningful to users than 'md5'
+    @property
+    def query_id(self):
```
greenape: Why not just rename the md5 attribute?

flowmachine/flowmachine/core/model_result.py
jch1g10: `ModelResult.store()`/`ModelResult.to_sql()` stores a pandas DataFrame which is the result of running a model. Since the result is already calculated, there's no efficiency benefit from storing the dependencies when storing the result.

flowmachine/flowmachine/core/query.py
jch1g10: Hmm, I don't think this change is actually relevant to this PR. But yes, addressing the root issue seems more sensible than fixing a workaround.

flowmachine/flowmachine/core/query.py
```diff
@@ -537,7 +544,87 @@ def _make_sql(self, name: str, schema: Union[str, None] = None) -> List[str]:
             )
         return queries
 
-    def to_sql(self, name: str, schema: Union[str, None] = None) -> Future:
+    def _unstored_dependencies_graph(self) -> "DiGraph":
```
greenape: Wondering if these need to live on query, or might be better off in cache.py or their own submodule?

flowmachine/flowmachine/core/query.py
```diff
@@ -699,11 +798,13 @@ def store(self):
         try:
             table_name = self.fully_qualified_table_name
         except NotImplementedError:
-            raise ValueError("Cannot store an object of this type with these params")
+            raise UnstorableQueryError(self)
```
greenape: 👍

flowmachine/flowmachine/core/server/action_handlers.py
```diff
@@ -124,9 +125,15 @@ def action_handler__run_query(**action_params: dict) -> ZMQReply:
     try:
         query_id = q_info_lookup.get_query_id(action_params)
     except QueryInfoLookupError:
-        # Set the query running (it's safe to call this even if the query was set running before)
+        store_dependencies = not (
```
greenape: Hmm. Not sure about reading the env var in the handler. I think I'd prefer it was happening in the main server and passed to the handler as an arg, or the handler was selected based on the env var?

flowmachine/tests/test_utils.py
```diff
@@ -247,3 +248,23 @@ def test_plot_dependency_graph():
...
+
+def test_store_queries_in_order():
+    """
+    Test that store_queries_in_order() stores each query's dependencies before storing that query itself.
+    """
+
+    class QueryWithStoreAssertions(DummyQuery):
+        def store(self):
+            for query in self.dependencies:
+                assert query.is_stored
```
greenape: Neat :)

flowmachine/flowmachine/core/query.py
jch1g10: Moved into dependency_graph.py, along with all the dependency-graph-related stuff from utils.py

flowmachine/flowmachine/core/model_result.py
greenape: Right, but the model will generally speaking use the results of some queries in calculation, and indeed, you should be able to see that link if you view the graph of any model result.

jch1g10: True. I guess the reason I'm slightly conflicted is that when `store_dependencies=True` we store only the necessary dependencies (i.e. we don't store queries that are only dependencies of already-stored queries). In this case the ModelResult isn't _stored_, but it has been _run_, so I can't decide whether it's more consistent to store the dependencies or not to.

jch1g10: On the other hand, the dependencies will already have been stored when the model was run, so unless they've been removed from the cache it doesn't make any difference whether or not we try to store them again. So I suppose it doesn't make much sense to disallow setting `store_dependencies=True`.

flowmachine/flowmachine/core/server/action_handlers.py
jch1g10: I've moved all the environment-variable-reading in the server into `server_config.get_server_config`, which is called by `server.main` and returns a `NamedTuple` that can be passed around to the other server functions and action handlers.

flowmachine/flowmachine/core/server/server_config.py
```diff
@@ -0,0 +1,46 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import os
+from typing import NamedTuple
+
+
+class FlowmachineServerConfig(NamedTuple):
```
greenape: Huh. I did not know you could do that!

PR #1296, opened 2019-09-20
Add config option to specify SQL snippet for determining available dates in FlowETL

### LINKED ISSUE #1295: SQL snippet to detect available dates should be configurable in FlowETL
When ingesting from an external Postgres database, by default the only information about available dates comes from the event_time column, so we need to run something along the lines of:
```
SELECT DISTINCT event_time::date as date FROM <source_table>
```
However, this is exceedingly slow for large data volumes and can lead to a timeout when detecting the available dates in FlowETL.

We can do a lot better if more is known about the data layout in the source database. The minimum we should do is allow configuring the SQL snippet to detect the available dates (and leave it to the user how to set things up so that this detection is sufficiently fast).

I'm not sure if we should go further and be a bit more specific about how we require the user to set things up. For example, we could require them to create an explicitly view or regularly updated table in the source database which is called `available_dates` and for which we create an additional foreign data wrapper on the flowdb side? Thoughts are welcome.

maxalbert:
Closes #1295 

### I have:

### Description

The main addition of this PR is a config option called `sql_find_available_dates` which can be provided in the FlowETL `config/config.yml`. This can be used to specify a custom SQL snippet which is run to determine the dates available for ingestion, so that FlowETL can start ingestion processes for them (if not specified, the default for this is `"SELECT DISTINCT event_time::date as date FROM <source_table>"`, but this can be _very_ slow - see #1295).

The helper function `find_distinct_dates_in_table` in `etl_utils.py` has been removed because it is now superseded by this config setting.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Random sampling is now exposed via the API, for all non-aggregated query kinds. [#1007](https://github.com/Flowminder/FlowKit/issues/1007)
 - New aggregate added to FlowMachine - `HistogramAggregation`, which constructs histograms over the results of other queries. [#1075](https://github.com/Flowminder/FlowKit/issues/1075)
 - New `IntereventInterval` query class - returns stats over the gap between events as a time interval.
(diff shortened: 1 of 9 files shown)


maxalbert:
@greenape Do you have any specific ideas how this could be improved?

PR #1309, opened 2019-09-24
Distance matrix speed

### LINKED ISSUE #1271: Add distance matrix as a core table?
For discussion - should we add a (cell) distance matrix as a core table?

Pros: potentially significant speed boost to any cell/tower-distance metrics (e.g. displacement, radius of gyration, contact reference location stats)
Cons: pretty big table

greenape:
Closes #1271

Distance matrix is, for even moderate numbers of cells, _crazy_ slow. Storing a full distance matrix is basically untenable because they're massive. This compromises, by allowing us to store the most compact possible distance matrix - the distinct geoms.

There are also some minor changes to classes using it, to ensure they use it effectively (i.e. a whole n x n distance matrix never gets calculated), and some corrections to previously erroneous tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,8 +25,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Attempting to store a query that does not have a standard table name (e.g. `EventTableSubset` or unseeded random sample) will now raise an `UnstorableQueryError` instead of `ValueError`.
 - In the FlowETL deployment example, the external ingestion database is now set up separately from the FlowKit components and connected to FlowDB via a docker overlay network. [#1276](https://github.com/Flowminder/FlowKit/issues/1276)
 - The `md5` attribute of the `Query` class has been renamed to `query_id` [#1288](https://github.com/Flowminder/FlowKit/issues/1288).
-- `DistanceMatrix` no longer returns rows for zero distances, e.g. cells on the same site.
 - Previously, `Displacement` defaulted to returning `NaN` for subscribers who have a location in the reference location but were not seen in the time period for the displacment query. These subscribers are no longer returned unless the `return_subscribers_not_seen` argument is set to `True`.
+- `DistanceMatrix` no longer returns duplicate rows for the lon-lat spatial unit.
+- Previously, `Displacement` defaulted to returning `NaN` for subscribers who have a location in the reference location but were not seen in the time period for the displacement query. These subscribers are no longer returned unless the `return_subscribers_not_seen` argument is set to `True`.
 
 
 ### Fixed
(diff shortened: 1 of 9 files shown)


flowmachine/flowmachine/features/spatial/distance_matrix.py
```diff
@@ -49,21 +77,41 @@ def __init__(
...
         col_names = [f"{c}_from" for c in self.spatial_unit.location_id_columns]
         col_names += [f"{c}_to" for c in self.spatial_unit.location_id_columns]
-        col_names += ["distance"]
+        col_names += ["value"]
         if self.return_geometry:
             col_names += ["geom_origin", "geom_destination"]
         return col_names
 
     def _make_query(self):
+        if self.spatial_unit == LonLatSpatialUnit():
```
jch1g10: I don't like to re-introduce an if/else for handling different spatial units, but I don't think there's an easy way around it in this case.

flowmachine/tests/test_displacement.py
```diff
@@ -19,7 +19,7 @@
         ("stddev", 182.885304966881, 218.698764480622),
         ("avg", 370.301009034846, 389.155343931983),
         ("sum", 11109.0302710454, 11285.5049740275),
-        ("variance", 33447034.7728291, 47829149.5853505),
+        ("variance", 33447.0347728291, 47829.1495853505),
```
jch1g10: Looks like `Displacement` was previously giving wrong answers for `statistic="variance"` when `unit="km"`, but would have been correct when `unit="m"`. Can we add a `statistic="variance"` case to `test_returns_expected_result_for_unit_m` to ensure it's correct for both choices of unit?

flowmachine/tests/test_displacement.py
```diff
@@ -127,7 +127,9 @@ def test_get_all_users_in_reference_location(get_dataframe):
     p2 = ("2016-01-01 12:01:00", "2016-01-01 15:20:00")
 
     rl = daily_location("2016-01-01", spatial_unit=make_spatial_unit("lon-lat"))
-    d = Displacement(p2[0], p2[1], reference_location=rl)
+    d = Displacement(
+        p2[0], p2[1], reference_location=rl, return_subscribers_not_seen=True
```
jch1g10: Can we also have a test for the new behaviour with `return_subscribers_not_seen=False`?

CHANGELOG.md
```diff
@@ -25,6 +25,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
...
 - In the FlowETL deployment example, the external ingestion database is now set up separately from the FlowKit components and connected to FlowDB via a docker overlay network. [#1276](https://github.com/Flowminder/FlowKit/issues/1276)
 - The `md5` attribute of the `Query` class has been renamed to `query_id` [#1288](https://github.com/Flowminder/FlowKit/issues/1288).
+- Previously, `Displacement` defaulted to returning `NaN` for subscribers who have a location in the reference location but were not seen in the time period for the displacment query. These subscribers are no longer returned unless the `return_subscribers_not_seen` argument is set to `True`.
+- `DistanceMatrix` no longer returns duplicate rows for the lon-lat spatial unit.
+- Previously, `Displacement` defaulted to returning `NaN` for subscribers who have a location in the reference location but were not seen in the time period for the displacement query. These subscribers are no longer returned unless the `return_subscribers_not_seen` argument is set to `True`.
```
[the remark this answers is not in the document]
greenape: Rebase issue I think :)

PR #1312, opened 2019-09-25
Faster population weighted opportunities

greenape:
The PWO model has previous partially implemented in Python. This converts it fully to SQL, moves it out of the models hierarchy and substantially speeds it up (goes from 7-8s on CI to 1-2s). Nice side effect is that we'll be able to expose it via the api.

There were a number of tests which actually covered aspects of the ModelResult class in here that I haven't yet reinstated. I'll add some of those back in to this PR once the coverage report is complete.


flowmachine/tests/test_model.py
```diff
@@ -0,0 +1,208 @@
...
+
+import pandas as pd
+import pytest
+
+from flowmachine.core import Query
+from flowmachine.core.model import Model, model_result
+from flowmachine.core.model_result import ModelResult
+
+
+class DummyQuery(Query):
```
jch1g10: Not a serious issue, but is it worth renaming this to avoid confusion with `flowmachine.core.dummy_query.DummyQuery`?

flowmachine/flowmachine/features/location/pwo.py
```diff
@@ -0,0 +1,297 @@
...
+    ----------
+    .. [1] Yan X-Y, Zhao C, Fan Y, Di Z, Wang W-X. 2014 "Universal predictability of mobility patterns in cities". J. R. Soc. Interface 11: 20140834. http://dx.doi.org/10.1098/rsif.2014.0834
+    """
+
+    def __init__(
+        self,
+        start: str,
+        stop: str,
+        *,
+        spatial_unit: Optional[LonLatSpatialUnit] = None,
```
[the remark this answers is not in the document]
jch1g10: No, ignore that...

flowmachine/flowmachine/features/location/pwo.py
```diff
@@ -0,0 +1,297 @@
...
+    the number of people that depart a certain area. This
+    model is useful for studying mobility pattern in cities.
+
+    Parameters
+    ----------
+    start : str
+        ISO format date string denoting the first day of data to include
+    stop : str
+        ISO format date string denoting the last day of data to include
+    spatial_unit : flowmachine.core.spatial_unit.*SpatialUnit, default admin3
```
jch1g10:
```suggestion
    spatial_unit : flowmachine.core.spatial_unit.*SpatialUnit, default versioned-site
```

flowmachine/tests/test_model.py
greenape: We have purpose specific `DummyQuery` classes in a few places, so I'm not super worried about that.

PR #1314, opened 2019-09-25
Distance series query

### LINKED ISSUE #1313: Distance series
For some purposes, e.g. https://arxiv.org/pdf/1908.02377.pdf it is necessary to have a time series per-subscriber of their distance from their home location or an arbitrary reference point.

This is sort of possible using `Displacement`, but it isn't really convenient or performant to use  that to construct a time series where what you actually want is some stat over the distance per day.

greenape:
Closes #1313

### I have:

### Description

Adds a `DistanceSeries` query to flowmachine. 

This outputs, per subscriber, a time series of some stat over the distances there were observed to be from a reference location, or arbitary reference point. E.g. minimum distance from 0,0 per day, or  median distance from home per week.

### DIFF
--- a/flowmachine/flowmachine/features/subscriber/distance_series.py
+++ b/flowmachine/flowmachine/features/subscriber/distance_series.py
@@ -14,7 +14,17 @@ from ..utilities.subscriber_locations import SubscriberLocations, BaseLocation
 
 
 valid_stats = {"sum", "avg", "max", "min", "median", "stddev", "variance"}
-valid_time_buckets = ["second", "minute", "hour", "day", "week", "month", "quarter", "year", "century"]
+valid_time_buckets = [
+    "second",
+    "minute",
+    "hour",
+    "day",
+    "week",
+    "month",
+    "quarter",
+    "year",
+    "century",
+]
 
 
 class DistanceSeries(SubscriberFeature):
@@ -70,7 +80,6 @@ class DistanceSeries(SubscriberFeature):
             raise ValueError(
                 f"'{time_bucket}' is not a valid value for time_bucket. Use one of {valid_time_buckets}"
             )
-
 
         if statistic not in valid_stats:
             raise ValueError(
--- a/flowmachine/flowmachine/features/subscriber/distance_series.py
+++ b/flowmachine/flowmachine/features/subscriber/distance_series.py
@@ -6,7 +6,7 @@
 """
 Per subscriber time series of distances from some reference location.
 """
-from typing import List, Optional
+from typing import List, Optional, Union, Tuple
 
 from flowmachine.features.spatial import DistanceMatrix
 from .metaclasses import SubscriberFeature
@@ -20,14 +20,21 @@ valid_time_buckets = ["second", "minute", "hour", "day", "month", "year", "centu
 class DistanceSeries(SubscriberFeature):
     """
     Per subscriber time series of distance in meters from some reference location.
+    For the time series, returns the first date/datetime within the time bucket for each
+    row, e.g. 1/1/1999 for a year bucket, 1/1/2026, 1/2/2026 and so on for a month bucket.
+
+    Notes
+    -----
+    The datetime column will contain dates for time buckets longer than an hour, and datetimes for
+    time buckets less than a day.
 
     Parameters
     ----------
     subscriber_locations : SubscriberLocations
         A subscriber locations query with a lon-lat spatial unit to build the distance series against.
-    reference_location : BaseLocation or None, default None
-        The set of home locations from which to calculate distance at each sighting.
-        If not given then distances will be from (0, 0).
+    reference_location : BaseLocation or tuple of int, default (0, 0)
+        The set of home locations from which to calculate distance at each sighting, or a tuple
+        of lon-lat in WS84 projection.
     statistic : str
         the statistic to calculate one of 'sum', 'avg', 'max', 'min', 
         'median', 'stddev' or 'variance'
@@ -50,7 +57,7 @@ class DistanceSeries(SubscriberFeature):
         self,
         *,
         subscriber_locations: SubscriberLocations,
-        reference_location: Optional[BaseLocation] = None,
+        reference_location: Union[BaseLocation, Tuple[float, float]] = (0, 0),
         statistic: str = "avg",
         time_bucket: str = "day",
     ):
@@ -71,9 +78,8 @@ class DistanceSeries(SubscriberFeature):
             )
         self.start = subscriber_locations.start
         self.stop = subscriber_locations.stop
-        if reference_location is None:
-            # Using 0, 0
-            self.reference_location = None
+        if isinstance(reference_location, tuple):
+            self.reference_location = reference_location
             self.joined = subscriber_locations
         elif isinstance(reference_location, BaseLocation):
             if reference_location.spatial_unit != subscriber_locations.spatial_unit:
@@ -98,7 +104,7 @@ class DistanceSeries(SubscriberFeature):
             )
         else:
             raise ValueError(
-                "Argument 'reference_location' should be an instance of BaseLocation class or None. "
+                "Argument 'reference_location' should be an instance of BaseLocation class or a tuple of two floats. "
                 f"Got: {type(reference_location)}"
             )
 
@@ -109,9 +115,9 @@ class DistanceSeries(SubscriberFeature):
         return ["subscriber", "datetime", "value"]
 
     def _make_query(self):
-        if self.reference_location is None:
+        if isinstance(self.reference_location, tuple):
             joined = f"""
-            SELECT subscriber, time as time_to, ST_Distance(ST_Point(0, 0)::geography, ST_Point(lon, lat)::geography) as value_dist
+            SELECT subscriber, time as time_to, ST_Distance(ST_Point{self.reference_location}::geography, ST_Point(lon, lat)::geography) as value_dist
             FROM ({self.joined.get_query()}) _
             """
         else:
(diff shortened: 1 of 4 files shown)


flowmachine/flowmachine/features/subscriber/distance_series.py
```diff
@@ -0,0 +1,140 @@
...
+        time_bucket: str = "day",
+    ):
+        subscriber_locations.spatial_unit.verify_criterion("has_geography")
+        subscriber_locations.spatial_unit.verify_criterion("has_lon_lat_columns")
+        self.spatial_unit = subscriber_locations.spatial_unit
+        if time_bucket.lower() in valid_time_buckets:
+            self.aggregate_by = time_bucket.lower()
+        else:
+            raise ValueError(
+                f"'{time_bucket}' is not a valid value for time_bucket. Use one of {valid_time_buckets}"
```
maxalbert: I realise that we often perform ad-hoc validations whenever a parameter is passed (not just in this here, also elsewhere in the code base). In addition, we often use strings as a replacement for more meaningful types but then also add extra behaviour (e.g. casting the time bucket to a date depending on the granularity). It may be worth creating explicit classes which can absorb this behaviour (e.g `TimeBucket` and `AggregationStatistic` for the strings used in this file). Definitely not needed for this PR, just as a thought for the future.

flowmachine/flowmachine/features/subscriber/distance_series.py
```diff
@@ -0,0 +1,140 @@
...
+            )
+
+        self.statistic = statistic.lower()
+        if self.statistic not in valid_stats:
+            raise ValueError(
+                "{} is not a valid statistic. Use one of {}".format(
```
maxalbert: Use f-strings for formatting?

flowmachine/flowmachine/features/subscriber/distance_series.py
```diff
@@ -0,0 +1,140 @@
...
+                    self.statistic, valid_stats
+                )
+            )
+        self.start = subscriber_locations.start
+        self.stop = subscriber_locations.stop
+        if reference_location is None:
+            # Using 0, 0
+            self.reference_location = None
```
maxalbert: Is there a reason why we can't directly say something like `self.reference_location = (0, 0)` here? (Not sure what the right syntax is to create a point location, but you get the idea.)

Or does this prevent us from doing some sort of optimisation later because the value is constant?

maxalbert: (I would hope that setting the `reference_location` here would allow us to get rid of the if/else branch in `_make_query()` below.)

flowmachine/flowmachine/features/subscriber/distance_series.py
```diff
@@ -0,0 +1,140 @@
...
+"""
+from typing import List, Optional
+
+from flowmachine.features.spatial import DistanceMatrix
+from .metaclasses import SubscriberFeature
+from ..utilities.subscriber_locations import SubscriberLocations, BaseLocation
+
+
+valid_stats = {"sum", "avg", "max", "min", "median", "stddev", "variance"}
+valid_time_buckets = ["second", "minute", "hour", "day", "month", "year", "century"]
```
maxalbert: Is `"week"` deliberately missing? Just checking - I can see that it is a slightly ambiguous concept and we may not want to include it because of this.

flowmachine/flowmachine/features/subscriber/distance_series.py
```diff
@@ -0,0 +1,140 @@
...
+            joined = f"""
+            SELECT subscriber, time as time_to, ST_Distance(ST_Point(0, 0)::geography, ST_Point(lon, lat)::geography) as value_dist
+            FROM ({self.joined.get_query()}) _
+            """
+        else:
+            joined = self.joined.get_query()
+
+        if valid_time_buckets.index(self.aggregate_by) > valid_time_buckets.index(
+            "hour"
+        ):  # Slightly nicer to cast things which aren't timestamps to a date
```
maxalbert: I can definitely see the convenience in this. I'm slightly in two minds about it because it makes the return type inconsistent (which may make further processing awkward if it's not done manually), but that's probably ok. We should mention the automated casting in the docstring, though.

flowmachine/flowmachine/features/subscriber/distance_series.py
maxalbert: An advantage of having a dedicated class for this would be that we don't need to repeat the valid values in the docstring here (or anywhere else where it is used) but can just refer to the docstring of `TimeBucket`/`AggregationStatistic`.

flowmachine/flowmachine/features/subscriber/distance_series.py
```diff
@@ -0,0 +1,140 @@
...
+    subscriber_locations : SubscriberLocations
+        A subscriber locations query with a lon-lat spatial unit to build the distance series against.
+    reference_location : BaseLocation or None, default None
+        The set of home locations from which to calculate distance at each sighting.
+        If not given then distances will be from (0, 0).
+    statistic : str
+        the statistic to calculate one of 'sum', 'avg', 'max', 'min', 
+        'median', 'stddev' or 'variance'
+    time_bucket : {"second", "minute", "hour", "day", "month", "year", "century"}, default "day"
+        Time bucket to calculate the statistic over.
```
maxalbert: When I tried this out I was slightly confused that for `time_bucket="month"/"year"` it would still return single dates in the `datetime` column. What do these dates refer to? Are they the first date in the time bucket? Would be good to clarify this in the docstring.

flowmachine/tests/test_subscriber_distance_series.py
```diff
@@ -0,0 +1,152 @@
...
+            )
+        )
+
+
+def test_error_on_spatial_unit_mismatch():
+    """
+    Test that error is raised if the spatial unit of the subscriber locations isn't point.
+    """
+
+    rl = daily_location("2016-01-01")
```
maxalbert: Can we make it explicit in the constructor here that this uses a non-point spatial unit, rather than relying on the default (which may change int the future)?

flowmachine/flowmachine/features/subscriber/distance_series.py
maxalbert: (And it would allow us to get rid of various tests for invalid time buckets / statistics, which are currently probably repeated in multiple places, because invalid values would only be tested once for those classes.)

flowmachine/tests/test_subscriber_distance_series.py
```diff
@@ -0,0 +1,152 @@
...
+    with pytest.raises(ValueError):
+        DistanceSeries(
+            subscriber_locations=SubscriberLocations(
+                "2016-01-01", "2016-01-07", spatial_unit=make_spatial_unit("lon-lat")
+            ),
+            time_bucket="NOT_A_BUCKET",
+        )
+
+
+def test_reference_raises_error():
```
maxalbert:
```suggestion
def test_invalid_reference_raises_error():
```

flowmachine/tests/test_subscriber_distance_series.py
```diff
@@ -0,0 +1,152 @@
...
+
+
+def test_error_on_spatial_unit_mismatch():
+    """
+    Test that error is raised if the spatial unit of the subscriber locations isn't point.
+    """
+
+    rl = daily_location("2016-01-01")
+
+    with pytest.raises(ValueError):
```
maxalbert: Would be good to include the `matches=...` argument in any calls to `pytest.raises` to ensure that the errors are raised for the expected reasons. (Every now and then I hit some code that raises the expected exception type but for a different reason than I thought.)

E.g.:
```suggestion
    with pytest.raises(ValueError, match="reference_location must have the same spatial unit as subscriber_locations"):
```

(Also in the other tests below.)

flowmachine/flowmachine/features/subscriber/distance_series.py
```diff
@@ -0,0 +1,140 @@
...
+        subscriber_locations.spatial_unit.verify_criterion("has_lon_lat_columns")
+        self.spatial_unit = subscriber_locations.spatial_unit
+        if time_bucket.lower() in valid_time_buckets:
+            self.aggregate_by = time_bucket.lower()
+        else:
+            raise ValueError(
+                f"'{time_bucket}' is not a valid value for time_bucket. Use one of {valid_time_buckets}"
+            )
+
+        self.statistic = statistic.lower()
```
maxalbert: Swap this line with the check below? Otherwise it will report the statistic in lowercase in the error message:
```
ValueError: not_a_statistic is not a valid statistic. Use one of {'sum', 'min', 'avg', 'stddev', 'variance', 'max', 'median'}
```

flowmachine/flowmachine/features/subscriber/distance_series.py
maxalbert: Again, definitely not for this PR. I'll probably add separate Github issues for this so that we can discuss/track it there.

greenape: Agree - one nice thing about a class for time buckets would be that one could do ordered comparisons between them as well.

flowmachine/flowmachine/features/subscriber/distance_series.py
```diff
@@ -0,0 +1,154 @@
...
+        subscriber_locations.spatial_unit.verify_criterion("has_lon_lat_columns")
+        self.spatial_unit = subscriber_locations.spatial_unit
+        if time_bucket.lower() in valid_time_buckets:
+            self.aggregate_by = time_bucket.lower()
+        else:
+            raise ValueError(
+                f"'{time_bucket}' is not a valid value for time_bucket. Use one of {valid_time_buckets}"
+            )
+
+        if statistic not in valid_stats:
```
maxalbert:
```suggestion
        if statistic.lower() not in valid_stats:
```

PR #1338, opened 2019-10-02
Imputed distance series

### LINKED ISSUE #1337: Imputed distance series
Following on from #1313, we need an imputed variation of the distance series, presumably using the rolling median method described in https://arxiv.org/pdf/1908.02377.pdf

greenape:
Closes #1337

### I have:

### Description

Builds on #1314 to add an imputed version of distance series.


flowmachine/tests/test_subscriber_imputed_distance_series.py
```diff
@@ -0,0 +1,78 @@
...
+import pandas as pd
+
+from flowmachine.features.subscriber.imputed_distance_series import (
+    ImputedDistanceSeries,
+)
+
+
+def fill_in_dates(df, window, start, stop):
+    """
+    :param df: input df, assumed to have columns ['subscriber', 'date', 'dist_to_hl']
```
maxalbert: Shouldn't we use the usual docstring style here?

flowmachine/tests/test_subscriber_imputed_distance_series.py
```diff
@@ -0,0 +1,78 @@
...
+    :param window: window size for calculating rolling median
```
maxalbert: Can we call this parameter `window_size` instead of `window`?

flowmachine/tests/test_subscriber_imputed_distance_series.py
```diff
@@ -0,0 +1,78 @@
...
+        if ds_df[ds_df.subscriber == sub].datetime.nunique() > 3:
+            to_be_imputed = ds_df[ds_df.subscriber == sub].sort_values("datetime")
+            imputed = fill_in_dates(to_be_imputed, 3, sl.start, sl.stop)
+            assert imputed.value.values.tolist() == pytest.approx(
+                sql[sql.subscriber == sub].value.tolist()
+            )
+
+
+@pytest.mark.parametrize(
+    "size, match", [(1, "positive"), (0, "positive"), (-1, "positive"), (4, "odd")]
```
maxalbert: For clarity, would be good to expand the error messages a little - e.g. replace `"positive"` with `"should be positive"`; and for the first one maybe also include `"... and greater than 1"` because 1 is actually positive. ;)

PR #1446, opened 2019-10-21
Automatic cache management

### LINKED ISSUE #1307: Automatic cache management
#1284 changes the default behaviour of the FlowMachine server to cache all dependencies of queries requested through the API. This will significantly increase the rate at which the cache fills up.

This makes it more important to add some form of automatic cache shrinking to FlowMachine, to reduce the burden on the system administrator to keep the cache size in check.

_Originally posted by @greenape in https://github.com/Flowminder/FlowKit/pull/1284#pullrequestreview-291629014_

greenape:
Closes #1307

### I have:

### Description

Adds a watch-and-shrink service to flowmachine, which runs in a background thread and periodically calls `shrink_below_size` to reduce the size of the cache. The frequency of shrinkages is controlled by the `FLOWMACHINE_CACHE_PRUNING_FREQUENCY` environment variable to flowmachine.

Queries are now immune from being shrunk for a short time period after initially being cached, to prevent the shrinker removing things before they get a chance to be used. The length of immunity is controlled by the `cache_protected_period` config key in `cache.cache_config`, and can be set using the `CACHE_PROTECTED_PERIOD` environment variable to FlowDB, or using the `set_cache_protected_period` function in flowmachine's cache submodule.

It can also be overridden when calling any cache management function.


flowmachine/tests/server/test_server_config.py
```diff
@@ -14,11 +14,15 @@ def test_get_server_config(monkeypatch):
...
     monkeypatch.setenv("FLOWMACHINE_SERVER_DISABLE_DEPENDENCY_CACHING", "true")
+    monkeypatch.setenv("FLOWMACHINE_CACHE_PRUNING_FREQUENCY", 1)
+    monkeypatch.setenv("FLOWMACHINE_CACHE_PRUNING_TIMEOUT", 2)
     config = get_server_config()
-    assert len(config) == 3
+    assert len(config) == 5
     assert config.port == 5678
-    assert config.debug_mode == True
-    assert config.store_dependencies == False
+    assert config.debug_mode
```
jch1g10: Hmm. This caught me out in the past - `config.debug_mode` was getting set to `"true"` instead of `True`. Being a truthy string, `config.debug_mode="true"` would still pass this test.

flowmachine/flowmachine/core/cache.py
```diff
@@ -363,31 +365,45 @@ def get_query_object_by_id(connection: "Connection", query_id: str) -> "Query":
...
 ) -> List[Tuple["Query", int]]:
     """
     Get all cached query objects in ascending cache score order.
 
     Parameters
     ----------
     connection : Connection
+    protected_period : int, default None
+        Optionally specify a number of seconds within which cache entries are excluded. If None,
+        the value stored in cache.cache_config will be used. Set to -1 to ignore cache protection
```
jch1g10: I'd prefer to say 'set to a value less than 0', otherwise this gives the incorrect impression that -1 is a special value.

flowmachine/flowmachine/core/cache.py
```diff
@@ -430,22 +455,35 @@ def shrink_below_size(
...
     removed = []
     logger.info(
-        f"Shrinking cache from {initial_cache_size} to below {size_threshold}{' (dry run)' if dry_run else ''}."
+        f"Shrinking cache from {initial_cache_size} to below {size_threshold}{' (dry run)' if dry_run else ''}.",
+        initial_cache_size=initial_cache_size,
+        size_threshold=size_threshold,
+        dry_run=dry_run,
     )
 
     if dry_run:
```
jch1g10: I'm not sure dry_run will work correctly if a 'wet' run would remove more than one cached query - looks like a dry run would repeatedly 'remove' the same query at each iteration. There's a test to ensure that dry and wet runs give the same results, but in that test only one query is removed from cache.

I can't see a straightforward fix, though, and this issue isn't directly related to this PR, so I'm happy to make an issue and leave it for now if you'd prefer.

jch1g10: No, I read this a bit too quickly. Looks like it should work, but I don't think the test is checking this because it's only removing one query.

PR #1500, opened 2019-11-06
New schema

### LINKED ISSUE #1231: FlowDB build scripts (.sql and .sh) should live in the same folder

greenape:
Spinning this off from #1491. Closes #1231

This adds a new schema alongside the existing one, which will eventually supplant it altogether.

I've made some slight modifications from the version in #1491:

- The date and time dimensions are pre-populated and have many more fields
- The event type table are actual children of the events supertable.

Oh, and I've put all build scripts in bin/build and given them hopefully less confusing numbering with a spacing of 10 between any that depend on a previous one. Many have the same number, because it doesn't matter what order they run in relative to one another.

Original diagram:

image

This PR:

image-2


flowdb/bin/build/0030_schema_interactions.sql
```diff
@@ -0,0 +1,159 @@
...
+        timestamp               TIMESTAMPTZ NOT NULL,
+        called_subscriber_id    BIGINT REFERENCES interactions.subscriber(id),
+        called_party_cell_id    BIGINT REFERENCES interactions.locations(cell_id),
+        calling_party_msisdn    TEXT,
+        called_party_msisdn     TEXT,
+        call_duration           NUMERIC
+        CHECK ( event_type = 'calls' ), PRIMARY KEY (event_id, date_sk)
+    ) PARTITION BY RANGE (date_sk);
+
+    CREATE TABLE IF NOT EXISTS interactions.sms(
```
greenape: As I look at this again, I kind of wonder if my attempt to capture the fact than an SMS can go to multiple parties is actually helpful, given that our experience is that the data necessary to allow you to record that accurately has never been available.

Could rejig this a little so that there's an extra level of inheritance, which puts both the calls and sms under an 'interactions' table with their common fields, and treat SMS as only between two parties? Alternatively, would seem reasonable to create a view that captures that information.

flowdb/bin/build/0030_schema_interactions.sql
```diff
@@ -0,0 +1,159 @@
...
+    ) PARTITION BY RANGE (date_sk);
+
+    CREATE TABLE IF NOT EXISTS interactions.subscriber_sightings(
+
+        sighting_id             BIGSERIAL,
+        subscriber_id           BIGINT REFERENCES interactions.subscriber(id),
+        cell_id                 BIGINT REFERENCES interactions.locations(cell_id),
+        time_sk                 BIGINT REFERENCES public.d_time(time_dim_id),
+        date_sk                 BIGINT REFERENCES public.d_date(date_dim_id),
+        event_id                BIGINT, /* REFERENCES interactions.event_supertable(event_id), - this can be added in PG12 */
```
greenape: (Still can't actually make this a reference with the inheritance structure.)

flowdb/bin/build/0030_schema_interactions.sql
```diff
@@ -0,0 +1,158 @@
...
+    CREATE INDEX ON interactions.subscriber (id);
+    CREATE INDEX ON interactions.subscriber (id, msisdn);
+
+
+
+    CREATE TABLE IF NOT EXISTS interactions.locations(
+
+        cell_id                 BIGSERIAL PRIMARY KEY,
+        site_id                 TEXT,
+        mno_cell_code           TEXT
```
greenape: I think this should really be a FK referencing the cells table. That's a little awkward, because the cells table is PKd on the mno id and version, so we might want to adjust that to be keyed on an id we generate and uniqued on those fields?

flowdb/bin/build/0030_schema_interactions.sql
greenape: Would also be possible to:

- Make the supertable a distinct partitioned table
- Make the event type tables separate partitioned tables which ref to the PK of that table and contain only non common info
- Create a view to capture the interactions (calls & sms) commonality

flowdb/bin/build/0010_schema_public.sql
```diff
@@ -0,0 +1,153 @@
...
+           END AS is_std_weekend
+    FROM (SELECT '1979-01-01'::DATE+ SEQUENCE.DAY AS datum
+          FROM GENERATE_SERIES (0,29219) AS SEQUENCE (DAY)
+          GROUP BY SEQUENCE.DAY) DQ
+    ORDER BY 1;
+
+    CREATE INDEX d_date_date_actual_idx
+        ON d_date(date_actual);
+
+    CREATE TABLE IF NOT EXISTS d_time(
```
greenape: My hunch is that we don't really want this at minute resolution and hour is better for our purposes. We could of course add in a higher resolution time dimension should that become necessary.

flowdb/bin/build/0020_schema_geography.sql
```diff
@@ -0,0 +1,42 @@
...
+-----------------------------------------------------------
+*/
+CREATE SCHEMA IF NOT EXISTS geography;
+
+CREATE TABLE IF NOT EXISTS geography.geoms (
+    gid BIGSERIAL PRIMARY KEY,
+    added_date TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT NOW(),
+    short_name VARCHAR, /* e.g. admin3pcod */
+    long_name TEXT, /* e.g. admin3name */
+    kind VARCHAR DEFAULT 'admin_unit', /* The type of the geom - admin unit, coverage polygon etc. */
```
greenape: Should probably kick this out to a separate table?

flowdb/bin/build/0020_schema_geography.sql
```diff
@@ -0,0 +1,42 @@
...
+SELECT AddGeometryColumn('geography','geoms','geom','4326','MULTIPOLYGON',2);
+CREATE INDEX "geography_geom_gist" ON "geography"."geoms" USING GIST ("geom");
+
+CREATE TABLE IF NOT EXISTS geography.geo_bridge (
+    cell_id BIGINT,
+    geo_id BIGINT REFERENCES  geography.geoms(gid),
+    valid_from DATE,
+    valid_to DATE,
+    weight DOUBLE PRECISION DEFAULT 1, /* Where a cell links to multiple geoms, this may be less than 1 */
+    linkage_method TEXT, /* Indicator for how the geometry is linked */
```
greenape: Should probably also kick to separate table

flowdb/bin/build/0030_schema_interactions.sql
greenape: I've reworked this so that the events table is a partitioned fact table with an event type field, and additional metadata is split out into the corresponding event type tables.

Also made sms effectively the same as calls, but without a duration which reflects the fact that we've never encountered reliable data to identify multi-recipient sms messages.

flowdb/bin/build/0010_schema_public.sql
```diff
@@ -0,0 +1,147 @@
...
+    INSERT INTO d_time
+    SELECT
+        -- Primary key
+        to_char(hour,'hh24')::INT AS time_dim_id,
+        -- Time of day
+        hour AS time_of_day,
+        -- Hour of the day (0 - 23)
+        EXTRACT(HOUR FROM hour) AS hour_of_day,
+        -- AM/PM
+        to_char(hour, 'am') AS meridian_indicator
```
traceyli: This one seems a bit arbitary?

flowdb/bin/build/0010_schema_public.sql
```diff
@@ -0,0 +1,147 @@
...
+      week_of_month            INT NOT NULL,
+      week_of_year             INT NOT NULL,
+      week_of_year_iso         CHAR(10) NOT NULL,
+      month_actual             INT NOT NULL,
+      month_name               VARCHAR(9) NOT NULL,
+      month_name_abbreviated   CHAR(3) NOT NULL,
+      quarter_actual           INT NOT NULL,
+      quarter_name             VARCHAR(9) NOT NULL,
+      year_actual              INT NOT NULL,
+      first_day_of_week        DATE NOT NULL,
```
traceyli: Weeks are Monday to Sunday? Would we adjust for places where 'weekends' are defined differently because e.g. Friday is a holy day? (I don't actually have a view about this at the moment, just ruminating for now...)

flowdb/bin/build/0010_schema_public.sql
```diff
@@ -0,0 +1,147 @@
...
+      last_day_of_week         DATE NOT NULL,
+      first_day_of_month       DATE NOT NULL,
+      last_day_of_month        DATE NOT NULL,
+      first_day_of_quarter     DATE NOT NULL,
+      last_day_of_quarter      DATE NOT NULL,
+      first_day_of_year        DATE NOT NULL,
+      last_day_of_year         DATE NOT NULL,
+      mmyyyy                   CHAR(6) NOT NULL,
+      mmddyyyy                 CHAR(10) NOT NULL,
+      is_std_weekend           BOOLEAN NOT NULL
```
traceyli: Related to my comment above about how weekends are defined.

flowdb/bin/build/0010_schema_public.sql
```diff
@@ -0,0 +1,147 @@
...
+           datum +(1 -EXTRACT(DAY FROM datum))::INT AS first_day_of_month,
+           (DATE_TRUNC('MONTH',datum) +INTERVAL '1 MONTH - 1 day')::DATE AS last_day_of_month,
+           DATE_TRUNC('quarter',datum)::DATE AS first_day_of_quarter,
+           (DATE_TRUNC('quarter',datum) +INTERVAL '3 MONTH - 1 day')::DATE AS last_day_of_quarter,
+           TO_DATE(EXTRACT(isoyear FROM datum) || '-01-01','YYYY-MM-DD') AS first_day_of_year,
+           TO_DATE(EXTRACT(isoyear FROM datum) || '-12-31','YYYY-MM-DD') AS last_day_of_year,
+           TO_CHAR(datum,'mmyyyy') AS mmyyyy,
+           TO_CHAR(datum,'mmddyyyy') AS mmddyyyy,
+           CASE
+             WHEN EXTRACT(isodow FROM datum) IN (6,7) THEN TRUE
```
traceyli: ISO definition of weekend is only useful for us if it matches with cultural weekends. And it's easy for the user to filter on day_name or day_of_week to choose their own weekends. So not sure if we should keep this here.

flowdb/bin/build/0010_schema_public.sql
```diff
@@ -0,0 +1,147 @@
...
+           TO_CHAR(datum,'YYYY"-W"IW-') || EXTRACT(isodow FROM datum) AS week_of_year_iso,
+           EXTRACT(MONTH FROM datum) AS month_actual,
+           TO_CHAR(datum,'Month') AS month_name,
+           TO_CHAR(datum,'Mon') AS month_name_abbreviated,
+           EXTRACT(quarter FROM datum) AS quarter_actual,
+           CASE
+             WHEN EXTRACT(quarter FROM datum) = 1 THEN 'First'
+             WHEN EXTRACT(quarter FROM datum) = 2 THEN 'Second'
+             WHEN EXTRACT(quarter FROM datum) = 3 THEN 'Third'
+             WHEN EXTRACT(quarter FROM datum) = 4 THEN 'Fourth'
```
traceyli: Seems like overkill to have this as well as quarter_actual.

flowdb/bin/build/0010_schema_public.sql
```diff
@@ -0,0 +1,147 @@
...
+      mmyyyy                   CHAR(6) NOT NULL,
+      mmddyyyy                 CHAR(10) NOT NULL,
+      is_std_weekend           BOOLEAN NOT NULL
+    );
+
+
+    INSERT INTO d_date
+    SELECT TO_CHAR(datum,'yyyymmdd')::INT AS date_dim_id,
+           datum AS date_actual,
+           EXTRACT(epoch FROM datum) AS epoch,
```
traceyli: Any examples of a use case for epoch? Just curious.

flowdb/bin/build/0010_schema_public.sql
```diff
@@ -0,0 +1,147 @@
...
+           EXTRACT(isodow FROM datum) AS day_of_week,
+           EXTRACT(DAY FROM datum) AS day_of_month,
+           datum - DATE_TRUNC('quarter',datum)::DATE +1 AS day_of_quarter,
+           EXTRACT(doy FROM datum) AS day_of_year,
+           TO_CHAR(datum,'W')::INT AS week_of_month,
+           EXTRACT(week FROM datum) AS week_of_year,
+           TO_CHAR(datum,'YYYY"-W"IW-') || EXTRACT(isodow FROM datum) AS week_of_year_iso,
+           EXTRACT(MONTH FROM datum) AS month_actual,
+           TO_CHAR(datum,'Month') AS month_name,
+           TO_CHAR(datum,'Mon') AS month_name_abbreviated,
```
traceyli: Only need one out of month_name or month_name_abbreviated

traceyli:
I started looking at this and realise that I'm not sure how to look at it... Should I look at it from the perspective of an analyst-type user that would want to use all the tables (which is how I started looking at it)? Or should I think about it more from the perspective where a lot of the tables are there to help calculations along, but probably won't be looked at or queried directly by an analyst?

flowdb/bin/build/0010_schema_public.sql
```diff
@@ -0,0 +1,147 @@
...
+    processed BOOLEAN,
+    datetime TIMESTAMPTZ,
+    size NUMERIC,
+    status TEXT,
+    record_number NUMERIC,
+    log TEXT
+
+    );
+
+    CREATE TABLE d_date
```
greenape:
```suggestion
    CREATE TABLE IF NOT EXISTS d_date
```

flowdb/bin/build/0010_schema_public.sql
greenape: Well I guess past a certain point it is on the deployer to modify the dimension to suit their needs - even at the several decades size populated here, it isn't a big deal to update values or even add a new column. Weeks are populated as ISO standard, so Monday is first day.

flowdb/bin/build/0010_schema_public.sql
greenape: I guess you don't _need_ either because everything can be extracted from the date. Given the creation of the table is one time, and the storage is negligible though, is the abbreviated name a sufficient convenience?

greenape:
> I started looking at this and realise that I'm not sure how to look at it... Should I look at it from the perspective of an analyst-type user that would want to use all the tables (which is how I started looking at it)? Or should I think about it more from the perspective where a lot of the tables are there to help calculations along, but probably won't be looked at or queried directly by an analyst?

I think both are useful perspectives - definitely no claim that this is exclusively all tables that should exist, so interested in identifying gaps (for example, I think we may well want a bunch of dimensions off the subscribers table). 
I think what we're aiming for here is a viable minimal storage structure that gives us a better basis to build from than the current schema that we can arrive at without additional processing. So, that needs to be useful to analysts as well, even if it doesn't cover all the things they might want, right?

chrisjbrooks:
general comments before I review in detail:
- avoid using SQL key words a column names e.g. timestamp, epoch
- `timestamp` vs `timestamptz`....."with tz" is the generally recommended option. Noting that I don't think we ever have ingested with a timezone defined other than UTC, since the data has never had timezone info included. Interested to understand if there are implications when indexing e.g. indexing on DATE(my_timestamptz) wouldn't be possible since my_timestamptz as a timestamptz data type is not immutable.

chrisjbrooks:
See if the new [generated columns feature](https://www.2ndquadrant.com/en/blog/partitioning-enhancements-in-postgresql-12/) is of use here. [Noting that](https://www.2ndquadrant.com/en/blog/generated-columns-in-postgresql-12/) generated columns cannot be used to partition, however indexes are fine on them.

PR #1552, opened 2019-11-18
Redact rows with small subscriber counts

### LINKED ISSUE #1026: Enforce redaction for small subscriber counts
Particularly for joined spatial aggregate - this should return NA where there are 15 or fewer subscribers in the spatial unit.

greenape:
Closes #1026

### I have:

### Description

Adds `RedactedX` wrapper classes for flows, unique subscriber counts, total location events, spatial aggregate and joined spatial aggregate which remove results where there are less than 16 sims in the aggregate. Replaces calls from flowapi to the original versions with calls to the redacted versions.


flowmachine/flowmachine/features/location/redacted_unique_subscriber_counts.py
```diff
@@ -0,0 +1,47 @@
...
+
+from typing import List
+
+from flowmachine.core import Query
+from flowmachine.core.mixins import GeoDataMixin
+from flowmachine.features import UniqueSubscriberCounts
+from flowmachine.features.location.spatial_aggregate import SpatialAggregate
+
+
+class RedactedUniqueSubscriberCounts(GeoDataMixin, Query):
```
jch1g10: The exposed "unique subscriber counts" is still just using `UniqueSubscriberCounts`, not `RedactedUniqueSubscriberCounts`.

greenape:
> is there anything an analyst could deduce if `LocationIntroversion` gives a result for a region but the corresponding region is redacted from a `UniqueSubscriberCounts` query?
> 

Hmm. Yeah. I suppose you leak that there's a nonzero number of subscribers there, which is not otherwise apparent.

flowmachine/flowmachine/features/location/meaningful_locations_aggregate.py
```diff
@@ -0,0 +1,55 @@
...
+
+    def __init__(
+        self,
+        *,
+        meaningful_locations: MeaningfulLocations,
+        spatial_unit: Optional[AnySpatialUnit] = None,
+    ) -> None:
+        self.meaningful_locations = meaningful_locations
+        if spatial_unit is None:
+            self.spatial_unit == make_spatial_unit("admin", level=3)
```
jch1g10:
```suggestion
            self.spatial_unit = make_spatial_unit("admin", level=3)
```

flowmachine/flowmachine/features/location/meaningful_locations_od.py
```diff
@@ -0,0 +1,81 @@
...
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from typing import Optional, List
+
+from flowmachine.core import Query, make_spatial_unit
+from flowmachine.core.spatial_unit import AnySpatialUnit
+from flowmachine.features import MeaningfulLocations
+
+
+class MeaningfulLocationsOD(Query):
```
jch1g10: Worth using the `FlowLike` mixin here?

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - FlowETL is now built on Airflow [10.1.6](https://airflow.apache.org/changelog.html#airflow-1-10-6-2019-10-28).
 - FlowETL now defaults to disabling Airflow's REST API, and enables RBAC for the webui. [#1516](https://github.com/Flowminder/FlowKit/issues/1516)
 - FlowETL now requires that the `FLOWETL_AIRFLOW_ADMIN_USERNAME` and `FLOWETL_AIRFLOW_ADMIN_PASSWORD` environment variables be set, which specify the default web ui account. [#1516](https://github.com/Flowminder/FlowKit/issues/1516)
+- FlowAPI will no longer return a result for rows in spatial aggregate, joined spatial aggregate, flows, total events or unique subscriber count where the aggregate would contain less than 16 sims. [#1026](https://github.com/Flowminder/FlowKit/issues/1026)
```
```suggestion
- FlowAPI will no longer return a result for rows in spatial aggregate, joined spatial aggregate, flows, total events, meaningful locations aggregate, meaningful locations od, or unique subscriber count where the aggregate would contain less than 16 sims. [#1026](https://github.com/Flowminder/FlowKit/issues/1026)
```

flowmachine/flowmachine/features/location/redacted_total_events.py
```diff
@@ -0,0 +1,85 @@
...
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from typing import List
+
+from flowmachine.core import Query
+from flowmachine.features import TotalLocationEvents
+
+
+class RedactedTotalEvents(Query):
```
jch1g10: I'm wondering whether it would make sense to implement this class as joining the `TotalLocationEvents` query to a `RedactedUniqueSubscriberLocations` query using the location ID columns? That should avoid having to re-implement the `TotalLocationEvents` SQL here, and would mean that `RedactedLocationMetric` is used by all redacted queries.

greenape: Ahh. Nice :)

greenape: Although the tradeoff is that one then needs to copy over all the args to the unique subscriber counts query, which is a bummer.

jch1g10: Yeah, that's not ideal. All the required arguments are attribute of the `TotalLocationEvents` query, so wouldn't be a problem to implement, but it's a little messy.

Alternatively, `RedactedTotalLocationEvents` could take the same args as `TotalLocationEvents` and construct both queries internally, but that's inconsistent with the other `Redacted*` queries and seems like a step in the opposite direction to several recent flowmachine changes.

greenape: I think on balance, I prefer the modified SQL really - in reality, this is likely to be necessary in a bunch of cases _anyway_.

greenape:
Something to consider for the future as well - another way to implement this would be to _drop_ the redacted versions in favour of adding an extra column to all aggregated queries containing the number of subscribers, and have a row level security policy that checked that. We wouldn't necessarily want to return that column of course.

PR #1909, opened 2020-02-05
Use etl records to determine available dates

### LINKED ISSUE #1873: Use FlowETL records to determine available dates
FlowETL keeps a record of all available dates in FlowDB. We should use that to determine available dates instead of the current fairly convoluted method.

greenape:
Closes #1873

### I have:

### Description

- Changes `Connection.available_dates` to a property, which returns data from the `etl.etl_records` table.
- Removes some redundant tests
- Removes the background refresh task for this, because large numbers of table objects should no longer be created on each request
- Adds etl records to test and synthetic data
- Removes the unused `Connection.inspector` attribute
- Removes unused `Connection.tables()` and `Connection.columns()` methods


flowmachine/tests/test_class_database.py
```diff
@@ -136,66 +78,16 @@ def test_strict_max_date(flowmachine_connect):
...
-
-@pytest.mark.parametrize(
-    "tables", [(1,), ("calls", 1), ["calls", "sms", "mds", "topups"], 0.1, 1]
-)
-def test_bad_table_inputs_to_available_dates(tables, flowmachine_connect):
-    """TypeError should be raised for bad inputs to available_dates."""
-    with pytest.raises(TypeError):
-        flowmachine_connect.available_dates(table=tables)
+    for table in tables:
+        assert isinstance(flowmachine_connect.available_dates[table], list)
```
jc-harrison: What is this testing now? Previously it tested that the tables requested were all keys of `available_dates()`, but `available_dates` no longer has a 'table' argument.

At the moment this tests that `available_dates[table]` is a list for all specified tables, which would be true for _any_ `table` key by virtue of `available_dates` being a `defaultdict`. This is a reasonable thing to test, but we may also want to test that `available_dates` contains a _non-empty_ list for all available CDR types.

In either case, since `available_dates` no longer has  arguments this test no longer needs to be parametrised.

PR #1999, opened 2020-02-24
Singleton

### LINKED ISSUE #391: Connection should be passed into flowmachine query classes
Closely related to #391, the query classes should be refactored so that the flowdb connection is passed into them, rather than the connection being instantiated globally and accessed through `Query.connection`. This would have the advantage that we can more easily pass a dummy connection during tests, and hopefully allow creating query objects without a active connection.

greenape:
Closes #391

### I have:

### Description

This makes the database connection, redis connection, and thread pool [context variables](https://docs.python.org/3.7/library/contextvars.html) which are accessed by getter methods rather than stuck on `Query`.

This means they can be more easily manipulated, there's no longer a hard requirement that they're set during query init, and that they can be temporarily overridden using a context manager (so, for example, one could connect to multiple flowdb instances from flowmachine).

### DIFF
--- a/flowmachine/flowmachine/core/cache.py
+++ b/flowmachine/flowmachine/core/cache.py
@@ -98,7 +98,7 @@ def write_query_to_cache(
 
     """
     logger.debug(f"Trying to switch '{query.query_id}' to executing state.")
-    q_state_machine = QueryStateMachine(redis, query.query_id)
+    q_state_machine = QueryStateMachine(redis, query.query_id, connection.conn_id)
     current_state, this_thread_is_owner = q_state_machine.execute()
     if this_thread_is_owner:
         logger.debug(f"In charge of executing '{query.query_id}'.")
@@ -303,7 +303,9 @@ def resync_redis_with_cache(connection: "Connection", redis: StrictRedis) -> Non
     logger.debug("Flushing redis.")
     for event in (QueryEvent.QUEUE, QueryEvent.EXECUTE, QueryEvent.FINISH):
         for qid in queries_in_cache:
-            new_state, changed = QueryStateMachine(redis, qid[0]).trigger_event(event)
+            new_state, changed = QueryStateMachine(
+                redis, qid[0], connection.conn_id
+            ).trigger_event(event)
             logger.debug(
                 "Redis resync",
                 fast_forwarded=qid[0],
--- a/flowmachine/flowmachine/core/cache.py
+++ b/flowmachine/flowmachine/core/cache.py
@@ -802,7 +802,7 @@ async def watch_and_shrink_cache(
     while True:
         logger.debug("Checking if cache should be shrunk.")
 
-        try:
+        try:  # Set the shrink function running with a copy of the current execution context (db conn etc) in background thread
             await asyncio.wait_for(
                 asyncio.get_running_loop().run_in_executor(
                     pool, copy_context().run, shrink_func
--- a/flowmachine/flowmachine/core/cache.py
+++ b/flowmachine/flowmachine/core/cache.py
@@ -9,6 +9,7 @@ Functions which deal with inspecting and managing the query cache.
 """
 import asyncio
 import pickle
+from contextvars import copy_context
 from concurrent.futures import Executor, TimeoutError
 from functools import partial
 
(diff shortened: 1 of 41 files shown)


flowmachine/flowmachine/core/cache.py
```diff
@@ -801,7 +804,9 @@ def cache_table_exists(connection: "Connection", query_id: str) -> bool:
 
         try:
             await asyncio.wait_for(
-                asyncio.get_running_loop().run_in_executor(pool, shrink_func),
+                asyncio.get_running_loop().run_in_executor(
+                    pool, copy_context().run, shrink_func
```
greenape: Might be worth adding a comment here to explain what the `copy_context` incantation is about?

flowmachine/flowmachine/core/connection.py
```diff
@@ -93,6 +94,10 @@ def __init__(
             pool_timeout=None,
             connect_args=connect_args,
         )
+        conn_id = md5(str(self.engine.url.host).encode())
+        conn_id.update(str(self.engine.url.port).encode())
+        conn_id.update(str(self.engine.url.database).encode())
+        self.conn_id = conn_id.hexdigest()
```
greenape: Should probably add a comment for this - purpose it to give the conn a per-db id so we can disambiguate if you happen to use the same redis with multiple flowdbs.

flowmachine/flowmachine/core/init.py
```diff
@@ -10,35 +10,201 @@
...
+    flowdb_password: Optional[str] = None,
+    flowdb_host: Optional[str] = None,
+    flowdb_connection_pool_size: Optional[int] = None,
+    flowdb_connection_pool_overflow: Optional[int] = None,
+    redis_host: Optional[str] = None,
+    redis_port: Optional[int] = None,
+    redis_password: Optional[str] = None,
+    conn: Optional[Connection] = None,
+) -> None:
+    """
```
greenape: Lot of doctoring duplication between these, which is annoying, but I think better than just reffing between them or assigning `__doc__` on balance.

flowmachine/flowmachine/core/query.py
```diff
@@ -70,7 +77,7 @@ def __init__(self, cache=True):
         obj = Query._QueryPool.get(self.query_id)
         if obj is None:
             try:
-                self.connection
+                get_db()
```
greenape: We could probably remove this to be honest.

flowmachine/flowmachine/core/query.py
```diff
@@ -70,7 +77,7 @@ def __init__(self, cache=True):
...
             except AttributeError:
```
greenape: This should definitely be removed cos you'll never hit the exception - `get_db` will raise it by itself.

flowmachine/flowmachine/core/query.py
```diff
@@ -163,7 +170,7 @@ def __format__(self, fmt=""):
         return f"<{', '.join(all_descriptions)}>"
 
     def __iter__(self):
-        con = self.connection.engine
+        con = get_db().engine
```
greenape: One thing I'm wrestling a little with across this whole PR is where one should use the context var and where one should pass in the object you get from it.

flowmachine/flowmachine/core/server/action_handlers.py
```diff
@@ -183,7 +188,7 @@ def _get_query_kind_for_query_id(query_id: str) -> Union[None, str]:
             status="error", msg=f"Unknown query id: '{query_id}'", payload=payload
         )
     else:
-        q_state_machine = QueryStateMachine(Query.redis, query_id)
+        q_state_machine = QueryStateMachine(get_redis(), query_id, get_db().conn_id)
```
greenape: There's an argument to be made here that the config param should be replaced by a contextvar in the action handers.

flowmachine/flowmachine/core/server/server.py
```diff
@@ -268,7 +265,7 @@ def main():
     # Read config options from environment variables
     config = get_server_config()
     # Connect to flowdb
-    flowdb_connection = flowmachine.connect()
+    flowmachine.connect()
```
greenape: Would also be reasonable to use the context manager variant here.

flowmachine/flowmachine/core/table.py
```diff
@@ -64,7 +65,7 @@ def __init__(self, name=None, schema=None, columns=None):
 
         """
         try:
-            self.connection
+            get_db()
```
greenape: Don't need this - `get_db` will raise it.

PR #2063, opened 2020-03-04
Direction enum

### LINKED ISSUE #357: Argument "direction" should be an enum, or a proper class
A bunch of metrics take a `direction` argument which can be one of the values `{"in", "out", "both"}`. There is a lot of duplicated logic that either checks the validity of these parameters or takes decisions based on one of the values. At the very least `direction` should be an enum to avoid these checks. It could possibly help to even promote it to a proper class so that other code doesn't need to take decisions on `direction`'s behalf when the logic really belongs there.

greenape:
Closes #357

### I have:

### Description

Adds a `Direction` enum, and uses this where we have previously faffed around with validating strings.

### DIFF
--- a/flowmachine/flowmachine/features/subscriber/event_type_proportion.py
+++ b/flowmachine/flowmachine/features/subscriber/event_type_proportion.py
@@ -4,11 +4,11 @@
 
 # -*- coding: utf-8 -*-
 
-import warnings
+from typing import Union
 
-from ...core import Table
-from .event_count import EventCount
-from .metaclasses import SubscriberFeature
+from flowmachine.features.subscriber.event_count import EventCount
+from flowmachine.features.subscriber.metaclasses import SubscriberFeature
+from flowmachine.features.utilities.direction_enum import Direction
 
 
 class ProportionEventType(SubscriberFeature):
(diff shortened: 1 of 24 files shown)


flowmachine/flowmachine/features/subscriber/event_type_proportion.py
```diff
@@ -35,7 +35,7 @@ class ProportionEventType(SubscriberFeature):
         If provided, string or list of string which are msisdn or imeis to limit
         results to; or, a query or table which has a column with a name matching
         subscriber_identifier (typically, msisdn), to limit results to.
-    direction : {'in', 'out', 'both'}, default 'both'
+    direction : {'in', 'out', 'both'} or Direction, default Direction.BOTH
```
jc-harrison: 'numerator_direction' docstring line should also be updated.

greenape:
> `ProportionEventReciprocal`, `IntereventInterval`, `IntereventPeriod`, `SubscriberCallDurations` and `PerLocationSubscriberCallDurations` all default to `Direction.OUT`; all others default to `Direction.BOTH`. Is there a reason for this?

Not beyond a desire to not break the tests in this case. The call durations ones, the original reasoning is that they're fairly slow and less data as a default is preferable.

Issue #1980, opened 2020-02-19
Object-oriented FlowClient?


jc-harrison:
I wonder whether it would be worth implementing a `Query` class in FlowClient, which contains a query specification along with a `run` method, and `status` and `result` properties. The flowclient functions that currently return a query spec could instead return a `fllowclient.Query` object.

My feeling is that this might be more intuitive for users, would remove the need for them to deal directly with query IDs, and may help to avoid confusion between query _specifications_ and query _results_.

PR #2095, opened 2020-03-10
Progress

### LINKED ISSUE #1202: Support for reporting progress of long-running queries
It would be useful to have better support for reporting progress / estimated duration until completion for long-running queries. Use cases are UIs that interact with FlowKit via the API (e.g. for visualisation) as well as debugging, code development and optimisation (to get a better feel for how queries scale).

As far as I'm aware PostgreSQL doesn't provide a lot of built-in support for this, but it is possible that we can provide some custom scaffolding for certain query types - e.g. if a large query is composed of smaller ones and these are cached then we can determine how many of them have already been calculated and estimate the progress this way.

(On a tangential note, a similar approach may also be useful when starting up flowdb containers with large sample data - we could add support for querying how many days of sample data have already been generated and display a progress bar based on this.)

I'm not sure how much we'll be able to do in this regard, but especially for user feedback through the API it would be good to have _some_ support.

greenape:
Closes #1202

### I have:

### Description

Adds supplementary information to poll and run returns, with the number of queries in cache vs number of cacheable subqueries for the query. This is currently written out to info level logging in flowclient, but I imagine we'd want something more apparent as part of #1980

While looking at this, it occurred to me that we'll shortly want to revisit the caching strategy, because it doesn't make sense to set all uncached queries running if they're in a layer below a totally cached one (which is entirely possible if the cache cleanup is working properly).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## [Unreleased]
 
 ### Added
+- FlowAPI now reports the proportion of subqueries cached for a query when polling. [#1202](https://github.com/Flowminder/FlowKit/issues/1202)
+- FlowClient now logs info messages with the proportion of subqueries cached for a query when polling. [#1202](https://github.com/Flowminder/FlowKit/issues/1202)
 
 ### Changed
 
--- a/flowapi/flowapi/query_endpoints.py
+++ b/flowapi/flowapi/query_endpoints.py
@@ -31,6 +31,20 @@ async def run_query():
               schema:
                 format: url
                 type: string
+          content:
+            application/json:
+              schema:
+                properties:
+                  query_id:
+                    type: string
+                  completed:
+                    items:
+                      format: int32
+                      type: integer
+                    type: array
+                    maxItems: 2
+                    minItems: 2
+                type: object
         '401':
           description: Unauthorized.
         '403':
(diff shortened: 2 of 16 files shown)


jc-harrison:
> While looking at this, it occurred to me that we'll shortly want to revisit the caching strategy, because it doesn't make sense to set all uncached queries running if they're in a layer below a totally cached one (which is entirely possible if the cache cleanup is working properly).

We already only set uncached dependencies running if they're dependencies of other uncached queries - `unstored_dependencies_graph` doesn't include dependencies of stored dependencies. This means that the dependencies ratio won't necessarily reach 100% when the query is ready. I'm not sure about the best way to handle that - perhaps count dependencies that don't need to be stored (because everything that depends on them is already stored) as completed?

greenape:
> > While looking at this, it occurred to me that we'll shortly want to revisit the caching strategy, because it doesn't make sense to set all uncached queries running if they're in a layer below a totally cached one (which is entirely possible if the cache cleanup is working properly).
> 
> We already only set uncached dependencies running if they're dependencies of other uncached queries - `unstored_dependencies_graph` doesn't include dependencies of stored dependencies. This means that the dependencies ratio won't necessarily reach 100% when the query is ready. I'm not sure about the best way to handle that - perhaps count dependencies that don't need to be stored (because everything that depends on them is already stored) as completed?

So we do - misread the func there :\

Hmm. I think better I find a way to not count them really.

greenape:
Right, I've tweaked this so you now get an object back of the form:

```json
{'eligible':10, 'queued':5, 'executing':5}
```

For a progress bar, this would put some additional onus on the client to keep the state of these values, but I think that's not untenable.

PR #2130, opened 2020-03-13
Add CSV and GeoJSON results streaming to flowapi

### LINKED ISSUE #2003: Helper method and/or API route to join query results to geography
A common pattern when using FlowClient is to get the result of a spatially-aggregated query, get the corresponding geography data, and join them so they can be used together to create a map visualisation. This join is an extra step that the user has to implement themselves. The simplest way to do this is to load the geography geojson into a geopandas GeoDataFrame and join this to the pandas DataFrame query result, but geopandas is not a dependency of flowclient, and has its own dependencies that can sometimes cause conflicts, so some users may prefer not to use geopandas.

It would be useful to add an API route from which a user could get a geojson object with geography data and the results of one or more spatially-aggregated queries as properties. This could either be implemented as a new query kind, an additional parameter in the `geography` endpoint (which could take one or more query IDs to join), or a new API endpoint. I'd favour one of the last two options, since the data type returned (geojson) will be different from that of an ordinary query.

I expect this would also be useful in scenarios where a different application (such as a dashboard) is interacting with the API, and it would be convenient to get a single geojson object for visualisation purposes with minimal client-side manipulation of the results.

greenape:
Closes #2003 

### I have:

### Description

FlowAPI will now try to return csv data from `/get/<query_id>.csv` and geojson from `/get/<query_id>.geojson`. It still returns regular json by default, and from `/get/<query_id>.json`.

FlowClient now has added `get_geojson_result_by_query_id` and `get_geojson_result` functions which do what one would expect.


greenape:
> When getting results as geojson, I think we should check the user has the appropriate `geography` permissions, in addition to permissions for the query itself.

I can't think of a non-annoying way to implement this one (i.e. without retrieving the params and searching through them for age units) :\ I'm a little inclined to say that permission to get the result is permission to get the result - irrespective of format. 

> We might want to revisit the `geojson_query` method for flows - currently this returns just the total inflows and outflows for each region, not all the flows. I'm not sure what the optimal representation would be in this case - ideally we'd want to avoid returning n^2 geometries for n regions, but still return the full flows data in a form that could be easily loaded into a GeoDataFrame and/or used in visualisation libraries such as Kepler.
> 

Should include the per-region flows - we (should) test this here: https://github.com/Flowminder/FlowKit/blob/8ebc6872e0b1eb6eda2348c4eb3504f94715a47f/flowmachine/tests/test_flows.py#L81

> I wonder whether there would be an advantage to enabling getting multiple query results as properties in a single geojson - the workflow I'm imagining is that an analyst runs all queries they need, then gets a single geojson with multiple properties (e.g. location introversion, subscriber counts, ...) which could then be used for styling different aspects of a single visualisation. I suppose it's a question of where we draw the line between what's implemented in FlowKit and what's left to the analyst to implement.
> 
> We could also consider the option of returning CSVs with geo data, as [supported by Kepler](https://github.com/keplergl/kepler.gl/blob/master/docs/user-guides/b-kepler-gl-workflow/a-add-data-to-the-map.md#supported-file-formats). Probably most beneficial for point locations (e.g. for creating arc plots of flows), although other geometries could be supported as WKT or geojson. Unclear whether this adds much benefit that isn't already covered by the geojson return type, though.

I'm meh on the first because I think that at that point you really might as well just do it with csvs or client side, and I'm up for the 2nd as a future extension of functionality.

jc-harrison:
> Should include the per-region flows

Ah, you're right. I mis-read the function. We might want some client-side helper functions to transform that into something more convenient, but that's a separate issue.

> I'm meh on the first because I think that at that point you really might as well just do it with csvs or client side, and I'm up for the 2nd as a future extension of functionality.

Agreed, let's consider it for the future, but it's out of scope for now.

jc-harrison:
> I can't think of a non-annoying way to implement this one (i.e. without retrieving the params and searching through them for age units) :\ I'm a little inclined to say that permission to get the result is permission to get the result - irrespective of format.

Yes, I see the problem. Disadvantage then is that to restrict a user from accessing sensitive geo data (e.g. network geographies), the permissions would need to be set for each query kind independently. Maybe we could add an aggegation-unit-for-all-queries selector to the FlowAuth UI to help with this? (separate issue)

greenape:
Mm. That's true.

Thinking alternatives... Bump the filetypes up to the front of the permissions dropdown (get a bit of an explosion of options), promote the various spatial units to classes and give them an aggregation_unit prop from which this can be retrieved, dig around in the query params for agg units.

The least untidy would seem make it a property of the spatial units to me.

Issue #2031, opened 2020-02-28
Rearrange FlowClient into submodules


jc-harrison:
Currently it can be tricky to explore the query kinds available in FlowClient, because they're mixed in with the other functions (e.g. `post_json`, `get_result`). I suggest rearranging FlowClient into submodules, to make the documentation a bit more navigable. Something like:

- core (functions for communicating with FlowAPI)
- locations (`daily_location` etc.)
- metrics (`radius_of_gyration` etc.)
- aggregates (`spatial_aggregate`, `location_introversion` etc.)
- network (`total_network_objects`)
- utils (`get_available_dates`, `get_geography`)

PR #2181, opened 2020-03-20
Flowclient query objects

### LINKED ISSUE #1980: Object-oriented FlowClient?

jc-harrison:
Closes #1980

### I have:

### Description

- Adds a `Query` class to FlowClient. This takes init parameters `connection` (API connection) and `parameters` (query spec), and has `run`, `get_result` and `wait_until_finished` methods and a `status` property. This doesn't replace any existing flowclient functions, just sits on top of them.
- Adds helper functions to create Query objects for query kinds that can be run directly (i.e. aggregates). These are defined in the `flowclient.aggregates` submodule (in a first step towards #2031).
- Appends '_spec' to the names of all functions that return a query specification, and renames the 'query' argument of `run_query` to 'query_spec'.
- Previously if a token expired, the solution would be to create a new connection object by calling `flowclient.connect()` again, and use the new connection everywhere, but now there's the possibility of a token expiring between creating a query object and getting the result/status.  We don't want to require the user to create a new query object if their token expires half way through, but I didn't want to allow the `connection` attribute to be replaced either (because the new connection could be for a different server, in which case the query object would then represent an entirely different query), so I've added a `Connection.update_token()` method to provide a new token to an existing connection object.
- `get_status` now returns `"not_running"` instead of raising a `FileNotFoundError` if query is not running (i.e. one of 'known', 'cancelled', 'resetting' or unknown). `get_status` still raises an error if the query state is 'errored'.

I haven't yet updated the examples to use Query objects. Once this is done we may want to remove the `flowclient.get_result` function (but keep `get_result_by_query_id`).

### Example

Previously, the workflow to run a `unique_subscriber_counts` query was:
```python
# Connect to API
conn = connect(...)
# Create query spec
query_spec = unique_subscriber_counts(start_date="2016-01-01", end_date="2016-01-08", aggregation_unit="admin3")
# Run query
query_id = run_query(connection=conn, query=query_spec)
# Check query status
print(get_status(connection=conn, query_id=query_id))
# Get result
result = get_result_by_query_id(connection=conn, query_id=query_id)
```
Using a Query object, this becomes:
```python
# Connect to API
conn = connect(...)
# Create query object
query = unique_subscriber_counts(connection=conn, start_date="2016-01-01", end_date="2016-01-08", aggregation_unit="admin3")
# Run query
query.run()
# Check query status
print(query.status)
# Get result
result = query.get_result()
```
(calling `query.run()` is optional - `query.get_result()` will run the query if necessary, if you're prepared to wait).

The 'old way' is still available, with a couple of small changes:
```python
# Connect to API
conn = connect(...)
# Create query spec
query_spec = aggregates.unique_subscriber_counts_spec(start_date="2016-01-01", end_date="2016-01-08", aggregation_unit="admin3")
# Run query
query_id = run_query(connection=conn, query_spec=query_spec)
# Check query status
print(get_status(connection=conn, query_id=query_id))
# Get result
result = get_result_by_query_id(connection=conn, query_id=query_id)
```


docs/source/analyst/flowclient/example_usage.ipynb
```diff
@@ -195,19 +195,19 @@
...
    "source": [
-    "dl1 = flowclient.daily_location(\n",
+    "dl1 = flowclient.daily_location_spec(\n",
     "    date=\"2016-01-01\",\n",
     "    aggregation_unit=\"admin3\",\n",
     "    method=\"last\",\n",
     "    subscriber_subset=None,\n",
     ")\n",
-    "dl2 = flowclient.daily_location(\n",
+    "dl2 = flowclient.daily_location_spec(\n",
```
greenape: I know we're planning an overhaul of these, but let's open an issue to track that?

flowclient/flowclient/aggregates.py
```diff
@@ -0,0 +1,1007 @@
...
+    References
+    ----------
+    .. [1] S. Isaacman et al., "Identifying Important Places in People's Lives from Cellular Network Data", International Conference on Pervasive Computing (2011), pp 133-151.
+    .. [2] Zagatti, Guilherme Augusto, et al. "A trip to work: Estimation of origin and destination of commuting patterns in the main metropolitan regions of Haiti using CDR." Development Engineering 3 (2018): 133-165.
+
+    Notes
+    -----
+    Does not return any value below 15.
+    """
+    return Query(
```
greenape: Can we change the name of the class? I think I'd prefer APIRequest or something like that. My rationale is that it disambiguates it from the `Query` class in FM.

flowclient/flowclient/query.py
```diff
@@ -0,0 +1,149 @@
...
+            - "not_running"
+            - "queued"
+            - "executing"
+            - "completed"
+        """
+        if not hasattr(self, "_query_id"):
+            return "not_running"
+        return get_status(connection=self.connection, query_id=self._query_id)
+
+    def get_result(
```
greenape: Definitely want a non-blocking version of this (spin out new issue)

flowclient/flowclient/query.py
```diff
@@ -0,0 +1,149 @@
...
+    Connection,
+    run_query,
+    get_status,
+    get_result_by_query_id,
+    get_geojson_result_by_query_id,
+    wait_for_query_to_be_ready,
+)
+
+
+class Query:
```
greenape: Can we also give this some sensible `__str__` rep?

flowclient/flowclient/aggregates.py
jc-harrison: Is `APIQuery` a fair compromise? To my mind, `APIRequest` sounds like it would encompass all API requests, including things like `get_available_dates`.

greenape: Fine with that :+1:

PR #2193, opened 2020-03-23
Progress bar

greenape:
### I have:

### Description

Adds a three-component progress bar when waiting for queries in flowclient.

Main bar displays x/n parts remaining to be run. Two ancillary bars display x/y running z/y queued. The main bar remains after the run completes, the two monitor bars disappear.

I think we want to merge #2181 and then I'll rebase onto that.


jc-harrison:
I tried running the following query (in JupyterLab):
```
query = fc.spatial_aggregate(
    connection=conn,
    locations=fc.modal_location_from_dates_spec(
        start_date="2016-01-01",
        end_date="2016-02-29",
        aggregation_unit="admin3",
        method="last",
    ),
)

query.run()
```
I then did
```
res = query.get_result()
```
and got 3 progress bars that all sat at 0% until the query finished, and then jumped to this:
image
I'll continue trying it out, but that doesn't look quite right.

greenape:
Mmm. Aside from anything else, the second two should vanish! 
I’ll get back to this one in a few days. Might need to chuck some of the polling or updating into another thread.

greenape:
Okay.. I've removed the nested bars, and this seems to work now.

jc-harrison:
Still not working for me. The progress bar sits at 0% (with the elapsed time updating ~ every 10 seconds) until the query finishes, and then jumps to 100%.

Screen Shot 2020-04-03 at 17 24 58

greenape:
That is super weird.. I wonder if it I’m not seeing it because I’m testing with a smaller dataset?

jc-harrison:
> That is super weird.. I wonder if it I’m not seeing it because I’m testing with a smaller dataset?

Hmm, possibly. I'll try with the test dataset and a smaller query.

greenape:
Let me know - I’ll do the inverse. Obviously if it is failing because of the larger data, then there’s a an issue that needs fixing.

jc-harrison:
No, I get the same issue with the smaller test dataset. The elapsed time updates more frequently (every second, not every 10 seconds), but progress stays at 0% until the result is ready.

greenape:
Requirements all up to date?

Did you say you were using the lab rather than the notebook?

jc-harrison:
I've tried in Jupyter lab (in the worked-examples container), notebook and iPython terminal - same result in each. I'm working from the FlowClient directory, with `pipenv install --dev` (and `pipenv install ipywidgets`, since that's not in the dev dependencies. I don't think it's likely to be an `ipywidgets` issue, though, since I'm seeing the same issue in the iPython terminal).

jc-harrison:
> Ok. How're you running flowmachine + flowapi?

Dockerised. I tried removing and rebuilding the images, in case the Makefile wasn't updating them for some reason, but that didn't help.

PR #3604, opened 2021-02-05
Allow flux check to be disabled

### LINKED ISSUE #3603: Make flux check optional in FlowETL dags
There are some situations where the flux check task in a FlowETL dag is unnecessary (e.g. a two-stage ingestion process where data from a multiple-day file are ingested into a staging table, and then daily dag runs extract from that staging table - a flux check is needed in the first stage, but the second-stage daily runs do not need to check for flux again). Since the table flux check can be quite slow, `create_dag` should provide an option to disable this check.

jc-harrison:
Closes #3603

### I have:

### Description

Adds a `use_flux_sensor` argument to `create_dag`, which can take the following values:
- `True` (default): use `FileFluxSensor` when extracting from a file, or `TableFluxSensor` otherwise
- `False`: Skip the flux check entirely
- 'table': use `TableFluxSensor` (replaces `use_file_flux_sensor=False`)
- 'file': use `FileFluxSensor` (will raise an error if not used with a file-based dag)

Setting `use_file_flux_sensor=False` will still force using TableFluxSensor, but will produce a DeprecationWarning.


flowetl/flowetl/flowetl/util.py
```diff
@@ -205,6 +221,27 @@ def create_dag(
     from flowetl.sensors.file_flux_sensor import FileFluxSensor
     from flowetl.sensors.table_flux_sensor import TableFluxSensor
 
+    # Choose flux sensor
```
greenape, line 224: Maybe stick this in a separate func?

greenape (approved):
Sounds good.

PR #3721, opened 2021-02-24
Expose most frequent location, harmonise start and end date argument names, expose hours argument

### LINKED ISSUE #3165: Change default daily location method
Given the results shown in https://arxiv.org/abs/2010.08814, we should change our default daily location method to 'most-frequent' from 'last', and make the hours parameter available via the API.

greenape:
Closes #3165

### I have:

### Description

- Exposes 2 new query types:
  - most frequent location, which is the 'raw' version of one daily location method and can be used as a reference location
  - total active periods, which allows one to calculate call days for subscribers (this was, weirdly, in there already but not completed)
- Renames any start and end date arguments in the api to `start_date` and `end_date
- Allows use of the hour slicing argument via the api where applicable
- Does some tidying up of how fields are attached to make it less of a pain to keep common ones in sync


jc-harrison (commented):
Ah, I got here too late. Never mind, here are a few comments.

Should `MostFrequentLocationSchema` also be added to `InputToModalLocationSchema`? Or does DailyLocation cover that use-case?

PR #4470, opened 2021-10-29
M4h industrialisation

Thingus:
Closes #ISSUE_NUMBER

### I have:

- [x ] Formatted any Python files with [black](https://github.com/ambv/black)
- [x ] Added any relevant Github labels
- [x ] Added tests for any new additions

### Description

PR that introduces M4H queries; ActiveSubsciber, UniqueActiveSubscriber, HomeLocationMonthly and MobilityEstimation. Also added a new test fixture, test_events_table, for using a set of data defined at test run time

ActiveSubscriber: Selects sub and date that have been seen X days out of Y previous
UniqueActiveSubscriber: List of unique subscribers from same
HomeLocationMonthly: Implements M4H home location technique
MobilityEstimation: Returns stats based on HomeLocation
test_events_table fixture: Creates and populates a new events table, events.test,that can be used as the events table in subscriber query objects. The table is dropped on the end of the test.


flowmachine/flowmachine/features/subscriber/active_subscribers.py
```diff
@@ -0,0 +1,120 @@
...
+            start_date,
+            end_date,
+            tables=events_tables,
+            subscriber_identifier=subscriber_id,
+            columns=[subscriber_id, "datetime"],
+        )
+        super().__init__()
+
+    @property
+    def start_date(self):
```
greenape, line 43: Use `standardise_date` func on the input  instead of this?

flowmachine/flowmachine/features/subscriber/active_subscribers.py
```diff
@@ -0,0 +1,120 @@
...
+        # Review questions:
+        # How should we pass dates around internally in fm?
+        # Should this return subscribers-day pairs, or just a list of subscribers?
+        # Should we offer the choice?
+
+        sql = f"""
+WITH ordered_events AS(
+	SELECT  subscriber, datetime
+	FROM ({self.events_table.get_query()}) AS tbl
+	WHERE datetime BETWEEN date('{self._window_start :%Y-%m-%d}') AND date('{self._end_date:%Y-%m-%d}')
```
greenape, line 85: This seems to be duplicating the functionality of `EventsTablesUnion`, why's it here?

flowmachine/flowmachine/features/subscriber/active_subscribers.py
```diff
@@ -0,0 +1,120 @@
...
+from typing import List, Union, Optional
+from flowmachine.core.query import Query
+from flowmachine.features.subscriber.call_days import CallDays
+from flowmachine.features.subscriber.interevent_interval import IntereventInterval
+from flowmachine.features.utilities.events_tables_union import EventsTablesUnion
+
+
+class ActiveSubscribers(Query):
+    """Returns a list of subscribers active `active_days`
+    out of `interval`, optionally with at least Z call-hours active"""
```
greenape, line 12: Checking my understanding here:

A row of this table would be: the `date`, the `subscriber`, the number of days in `date-interval` - `date` they were active.

Active means 'observed `active_hours`/24 hours` for a given day?

flowmachine/flowmachine/features/subscriber/home_location_monthly.py
```diff
@@ -0,0 +1,197 @@
...
+from flowmachine.features.subscriber.modal_location import ModalLocation
+from flowmachine.features.subscriber.unique_active_subscribers import (
+    UniqueActiveSubscribers,
+    ActiveSubscribers,
+)
+from flowmachine.features.subscriber.last_location import LastLocation
+from flowmachine.features.subscriber.location_visits import LocationVisits
+
+
+class HomeLocationMonthly(Query):
```
greenape, line 19: Can we make this more generic by saying period instead of month? Understand that our immediate usage context is monthly, but the functionality would equally be applied to fortnights, or biannually, right?

flowmachine/flowmachine/features/subscriber/home_location_monthly.py
```diff
@@ -0,0 +1,197 @@
...
+    def column_names(self) -> List[str]:
+        cols = ["subscriber", "location", "month"]
+        return cols
+
+    @property
+    def window_start(self):
+        return self._window_start.strftime("%Y-%m-%d")
+
+    @window_start.setter
+    def window_start(self, value):
```
greenape, line 101: Thinking about it, I'm not overwhelmingly opposed to using getters and setters for this stuff, but I'd rather they were defined once in the base class.

flowmachine/flowmachine/features/subscriber/home_location_monthly.py
```diff
@@ -0,0 +1,197 @@
...
+        last_locations_clause = ""
+
+        for last_location in self.last_locations:
+            last_locations_clause += f"""
+SELECT subscriber, pcod, '{last_location.start}' AS day
+FROM ({last_location.get_query()}) AS tbl
+UNION ALL
+"""
+
+        last_locations_clause = last_locations_clause.rstrip("UNION ALL\n")
```
greenape, line 133: Use a reduce and `.union(all=True)`, or `join` on the list? Or use `MultiLocation` which includes the date bit + handles non-admin levels.

flowmachine/flowmachine/features/subscriber/mobility_estimation.py
```diff
@@ -0,0 +1,93 @@
...
+            "high_mobility_resident_percent",
+            "displaced_resident_percent",
+        ]
+
+    # Questions for Galena;
+    # Is the kernel overlapping? skimage.median_filter points to no
+    # Are the csv files you're iterating over daily event records?
+    def _make_query(self):
+
+        home_location_monthly = HomeLocationMonthly(
```
greenape, line 57: Query objects should be created and assigned as attributes in `__init__` or the dependency relationships don't get picked up by caching.

flowmachine/tests/conftest.py
```diff
@@ -221,6 +230,28 @@ def get_length(flowmachine_connect):
     yield lambda query: len(pd.read_sql_query(query.get_query(), con=get_db().engine))
 
 
+@pytest.fixture()
+def test_events_table(flowmachine_connect):
+    """Creates a test event table and grants"""
+    # Creds hard-coded here for two reasons; 1) they're public anyway 2) This fixture will crash when run live
+    with connections(flowdb_user="flowdb", flowdb_password="flowflow"):
```
greenape, line 237: Get them from the environment vars instead.

flowmachine/flowmachine/features/subscriber/active_subscribers.py
```diff
@@ -0,0 +1,120 @@
+import datetime
+from datetime import timedelta, date, datetime
+from typing import List, Union, Optional
+from flowmachine.core.query import Query
+from flowmachine.features.subscriber.call_days import CallDays
+from flowmachine.features.subscriber.interevent_interval import IntereventInterval
+from flowmachine.features.utilities.events_tables_union import EventsTablesUnion
+
+
+class ActiveSubscribers(Query):
```
greenape, line 10: I see the value of this in parallel with `UniqueSubscribers`, which gives you active subscribers for a period/day without the 'n hours each day' bit. Although it is also kind of duplicating the functionality of `TotalActivePeriods` + a `numeric_subset` (e.g. I think you could produce with a convenience function that makes a union of subset `TotalActivePeriods` with a tiny bit of SQL for window func).

flowmachine/flowmachine/features/subscriber/active_subscribers.py
greenape: Ahh no, I see why. Could just use two `EventsTablesUnion` for the same effect (and it will probably resolve to the same plan tbh).

flowmachine/tests/test_active_subscribers.py
```diff
@@ -0,0 +1,56 @@
...
+from flowmachine.features.subscriber.active_subscribers import ActiveSubscribers
+from flowmachine.features.subscriber.unique_active_subscribers import (
+    UniqueActiveSubscribers,
+)
+from datetime import date, datetime
+from flowmachine.core.context import get_db
+
+
+@pytest.fixture()
+def active_sub_test_data(test_events_table):
```
greenape, line 11: I don't like this.

I get the convenience of doing it, but, if every test did it because it was convenient then the already huge runtime of the test suite would get untenable, and I _don't_ see the necessity - having verified (and if necessary cos it is doing something complicated, demonstrated), that the query does what it should and fails as one would expect etc. then it doesn't matter that the test data doesn't show a realistic case, only that it continues to behave the same way and give the same result.

flowmachine/tests/test_home_location_monthly.py
```diff
@@ -0,0 +1,80 @@
...
+        window_start="2016-01-01",
+        window_stop="2016-01-07",
+        agg_unit=AdminSpatialUnit(level=3),
+        home_this_month=3,
+        home_last_month=2,
+        events_tables=["events.test"],
+        modal_lookback=10,
+    )
+
+    out = list(iter(home_location_monthly))
```
greenape, line 48: Use the `get_dataframe` fixture in preference to this (because iterating over query results using the query object is still tabled for removal at some stage!)

greenape (changes requested):
Got a few comments - I think this needs to be broken down into a few smaller PRs also pls.

Pipfile
```diff
@@ -0,0 +1,14 @@
+[[source]]
```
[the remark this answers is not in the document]
Thingus: Done in next commit

flowmachine/flowmachine/features/subscriber/active_subscribers.py
```diff
@@ -0,0 +1,120 @@
...
+	SELECT  subscriber, datetime
+	FROM ({self.events_table.get_query()}) AS tbl
+	WHERE datetime BETWEEN date('{self._window_start :%Y-%m-%d}') AND date('{self._end_date:%Y-%m-%d}')
+	ORDER BY subscriber, datetime
+), seen_on_days AS(
+	SELECT DISTINCT ON (event_date, subscriber)
+	subscriber, datetime::date as event_date
+	FROM ordered_events
+	ORDER BY subscriber
+), dates_of_interest AS (
```
jc-harrison, line 92: Checking my understanding: the 'dates_of_interest' business is so we get an output row for a subscriber who was not active on day `x`, but was active on at least `active_days` in the interval up to day `x` (which would otherwise be missed because 'seen_on_days' doesn't have a row for day `x`)?

flowmachine/flowmachine/features/subscriber/active_subscribers.py
```diff
@@ -0,0 +1,120 @@
...
+	SELECT i::date as dates_of_interest
+	FROM generate_series('{self._window_start :%Y-%m-%d}', '{self._end_date:%Y-%m-%d}', '1 day'::interval) AS i
+), active_as_of AS (
+	SELECT dates_of_interest, subscriber
+	FROM dates_of_interest
+	LEFT OUTER JOIN seen_on_days ON dates_of_interest.dates_of_interest = seen_on_days.event_date
+	ORDER BY dates_of_interest
```
jc-harrison, line 99: Again, I don't think this needs to be ordered (since you've got an `ORDER BY dates_of_interest` in the window below)

jc-harrison (commented):
A few comments on ActiveSubscribers

flowmachine/flowmachine/features/subscriber/active_subscribers.py
Thingus: That's a precursor step; the row would be just `date` and `subscriber`, filtered where the number of days a subscriber is active in `date - interval` and `date` is >= `active_days`.
`active_hours` I've not implemented yet, but yes - `active_hours'/24 means a sub is active on that day.

flowmachine/flowmachine/features/subscriber/home_location_monthly.py
Thingus: Yes, that was my thought too - realised this during implementation.

flowmachine/tests/test_home_location_monthly.py
Thingus: Will do

flowmachine/tests/test_active_subscribers.py
Thingus: I'll drop the fixture and replace it with the regular test_data before merge, but I'm planning on keeping it till then as a dev tool - it's already helped me catch two errors in `home_location`.
I get your concerns about test suite runtime - I tried to build the parent fixture with `scope=session` to avoid repeated CREATE, INSERT and DELETE,  but that involved creating a factory class for this, and then you need to design a bespoke dataset that is useful for every test it's used for and _then_ you've fallen down the artificial test data hole again.

flowmachine/tests/conftest.py
Thingus: Will change, but as this is going away it's moot - see comment for https://github.com/Flowminder/FlowKit/pull/4470/files/9910196c325f5accac4e91563ed8771947694007#r740146136

flowmachine/flowmachine/features/subscriber/home_location_monthly.py
Thingus: I could have sworn I saw a mixin for doing this that already existed, but when I went looking for it I couldn't find it. Think we need a new one?

flowmachine/flowmachine/features/subscriber/active_subscribers.py
```diff
@@ -0,0 +1,120 @@
...
+        interval: int,
+        active_hours: int = None,
+        subscriber_id: str = "msisdn",
+        events_tables: Optional[List[str]] = None,
+    ):
+        self.start_date = start_date
+        self.end_date = end_date
+        self.active_days = active_days
+        self.interval = interval
+        self.active_hours = active_hours
```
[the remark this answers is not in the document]
Thingus: Correct

flowmachine/flowmachine/features/subscriber/active_subscribers.py
Thingus: Correct

flowmachine/Pipfile
```diff
@@ -24,6 +24,9 @@ marshmallow-oneofschema = "*"
 apispec-oneofschema = "*"
 get_secret_or_env_var = "*"
 redis = "*"
+watchdog = "*"
```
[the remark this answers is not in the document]
Thingus: Will do

flowmachine/flowmachine/features/subscriber/active_subscribers.py
```diff
@@ -0,0 +1,120 @@
...
+    def __init__(
+        self,
+        start_date: Union[date, str],
+        end_date: Union[date, str],
+        active_days: int,
+        interval: int,
+        active_hours: int = None,
+        subscriber_id: str = "msisdn",
+        events_tables: Optional[List[str]] = None,
+    ):
```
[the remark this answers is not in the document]
Thingus: Will do

flowmachine/flowmachine/features/subscriber/active_subscribers.py
Thingus: Will do

flowmachine/flowmachine/features/subscriber/mobility_estimation.py
Thingus: Gotcha - will move.

flowmachine/flowmachine/features/subscriber/home_location_monthly.py
```diff
@@ -0,0 +1,198 @@
...
+        events_tables=None,
+        modal_lookback=40,
+        active_days=4,
+        interval=7,
+    ):
+
+        self.window_start = window_start
+        self.window_stop = window_stop
+        self.ref_location = ref_location
+        self.agg_unit = agg_unit
```
jc-harrison, line 44: I'd prefer to call this parameter 'spatial_unit', I think, for consistency with the rest of flowmachine. It's not critical here, because we're not using GeoDataMixin (which explicitly looks for a `spatial_unit` attribute), but it's good to be consistent. (I know we use 'aggregation_unit' for the equivalent parameters exposed through the API, and we may want to harmonise these at some point, but I think that harmonisation would be easier if flowmachine and flowapi are separately self-consistent).

flowmachine/flowmachine/features/subscriber/home_location_monthly.py
```diff
@@ -0,0 +1,198 @@
...
+WITH last_locations AS (
+    {last_locations_clause}
+), modal_locations AS (
+    {self.modal_locations.get_query()}
+), location_histogram AS (
+	SELECT subscriber, pcod, count(pcod) as y
+	FROM last_locations
+	GROUP BY subscriber, pcod
+), this_month_known_homes AS (
+	SELECT modal_locations.subscriber, modal_locations.pcod
```
jc-harrison, line 146: `pcod` is not necessarily the right column to use here (depending on the spatial unit). You'll want to use `self.spatial_unit.location_id_columns`.

jc-harrison (commented):
A few comments on HomeLocationMonthly (probably more to follow).

flowmachine/flowmachine/features/subscriber/active_subscribers.py
```diff
@@ -0,0 +1,120 @@
...
+        self.start_date = start_date
+        self.end_date = end_date
+        self.active_days = active_days
+        self.interval = interval
+        self.active_hours = active_hours
+        self.sub_id_column = subscriber_id
+        self._window_start = self._start_date - timedelta(days=self.interval - 1)
+        self.events_table = EventsTablesUnion(
+            start_date,
```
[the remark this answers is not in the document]
Thingus: Yep; will do

flowmachine/flowmachine/features/subscriber/home_location_monthly.py
```diff
@@ -0,0 +1,198 @@
...
+     location is modal location and last location on more than X days last month), then choose this as their home
+     location. Otherwise set their home location to “unlocatable” (a special value, not NULL)."""
+
+    def __init__(
+        self,
+        window_start: Union[str, datetime],
+        window_stop: Union[str, datetime],
+        agg_unit: AnySpatialUnit,
+        home_this_month: int,
+        home_last_month: int,
```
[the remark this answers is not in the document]
Thingus: I've been calling them `known_location_threshold` and `unknown_location_threshold`, or just `X` and `Y` after the docstring - but none of these quite capture what it is that those are doing. Will think on it.

flowmachine/flowmachine/features/subscriber/home_location_monthly.py
Thingus: How about `XDaysInYLocation`?

Thingus: How about `ReferencedModalLocation`?

flowmachine/flowmachine/features/subscriber/home_location_monthly.py
Thingus: I'm going to leave them in for now, as it's useful to be able to have the dates as Datetime internally but exposed as a string for compatability with the rest of Flowmachine - agree that they need to be defined once, though.

flowmachine/flowmachine/features/subscriber/home_location_monthly.py
```diff
@@ -0,0 +1,198 @@
...
+
+    """For each subscriber, if their modal last location was their last location on more than X days this month
+    (choose X=15 here), then choose this as their home location. If not, but it was their last location on more than Y
+     days this month (Y<X; choose Y=10 here) AND it is the same as their “reference location” (in our case reference
+     location is modal location and last location on more than X days last month), then choose this as their home
+     location. Otherwise set their home location to “unlocatable” (a special value, not NULL)."""
+
+    def __init__(
+        self,
+        window_start: Union[str, datetime],
```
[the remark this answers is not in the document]
jc-harrison: Actually, in this situation the ActiveSubscribers query wouldn't be needed in this query at all - the subscriber subset would be applied to each of the other queries before passing them into this query.

flowmachine/flowmachine/features/subscriber/home_location_monthly.py
```diff
@@ -0,0 +1,198 @@
...
+	AND modal_locations.subscriber = location_histogram.subscriber
+	AND location_histogram.y >= {self.home_this_month}
+), """
+
+        if self.ref_location:
+            sql += f"""
+reference_locations AS (
+    SELECT subscriber, pcod
+    FROM ({self.ref_location.get_query()}) AS tbl
+), last_month_known_homes AS (
```
jc-harrison, line 159: Personally, I'd be tempted to combine 'this_month_known_homes' and 'last_month_known_homes', on the basis that the query result should have one row for each row in 'modal_locations' so it feels neater to avoid the `subscriber NOT IN (SELECT ...)`, and only join 'modal_locations' to 'location_histogram' once.

Something like this (square brackets mark reference_location-dependent lines):
```
  SELECT modal_locations.subscriber, modal_locations.pcod
  FROM modal_locations
  INNER JOIN location_histogram USING ( subscriber, pcod )
[ LEFT JOIN reference_locations USING (subscriber)  ]
  WHERE location_histogram.y >= {self.home_this_month}
[ OR ( modal_locations.pcod = reference_locations.pcod AND location_histogram.y >= {self.home_last_month} )  ]
```

It's a matter of preference though, I suppose - what seems neater to me might seem less so to others.

jc-harrison: Could even combine 'unknown_homes' into this as well, by moving the where condition into a `CASE WHEN` in the select:
```
  SELECT modal_locations.subscriber,
         CASE WHEN ( location_histogram.y >= {self.home_this_month} )
[             OR ( modal_locations.pcod = reference_locations.pcod AND location_histogram.y >= {self.home_last_month} )  ]
              THEN modal_locations.pcod
              ELSE 'unknown'
         END AS pcod
  FROM ...
```

But again, personal preference...

flowmachine/flowmachine/features/subscriber/home_location_monthly.py
```diff
@@ -0,0 +1,198 @@
...
+    SELECT subscriber, pcod
+    FROM this_month_known_homes
+    UNION ALL
+    SELECT subscriber, pcod
+    FROM last_month_known_homes
+), """
+
+        sql += f"""
+unknown_homes AS(
+	SELECT DISTINCT(subscriber), 'unknown' AS pcod
```
jc-harrison, line 187: Following on from my earlier comment about using `self.spatial_unit.location_id_columns` instead of `pcod` to handle spatial units other than admin, I've just realised setting the 'unknown' value could be problematic if the spatial unit doesn't have a single text field as the location ID (e.g. VersionedCell, which has two columns, or LonLat, which has a pair of coordinates). Not sure how to cleanly handle that situation.

jc-harrison (commented):
A few more comments (hopefully we can shuffle these across to the split PRs)

Issue #1009, opened 2019-07-01
Enable subscriber subsetting via api


greenape:

greenape:
Some things this should accept:

- Any metric admissible to joined spatial aggregate
- Any locator admissible to joined spatial aggregate
- Subsets of the above
- Random samples from the above

greenape:
Plus unique subscribers - one would like to enable a scenario where you can run queries with a random sample of all subscribers, via the api.

PR #4481, opened 2021-11-02
Active subscribers

Thingus:
Implementation of active_subscribers query

### I have:

### Description

Implements a query for retrieving subscribers seen (either at all or Z hours out of 24) X days out of Y. Also includes test fixtures for same.


flowmachine/flowmachine/features/subscriber/unique_active_subscribers.py
```diff
@@ -0,0 +1,20 @@
...
+from flowmachine.features.subscriber.active_subscribers import ActiveSubscribers
+
+
+class UniqueActiveSubscribers(ActiveSubscribers):
+    @property
+    def column_names(self) -> List[str]:
+        return ["subscriber"]
+
+    def _make_query(self):
+        list = super()._make_query()
```
greenape, line 12: Composition over inheritance, by which I mean to say: this should be a query which takes the `ActiveSubscribers` query as an argument and does something with it.

flowmachine/flowmachine/features/subscriber/active_subscribers.py
```diff
@@ -0,0 +1,123 @@
...
+        # Review questions:
+        # How should we pass dates around internally in fm?
+        # Should this return subscribers-day pairs, or just a list of subscribers?
+        # Should we offer the choice?
+
+        sql = f"""
+WITH ordered_events AS(
+	SELECT  subscriber, datetime
+	FROM ({self.events_table.get_query()}) AS tbl
+	WHERE datetime BETWEEN date('{self._window_start :%Y-%m-%d}') AND date('{self._end_date:%Y-%m-%d}')
```
jc-harrison, line 88: Shouldn't need to filter `datetime` here - that's already done in the `EventsTablesUnion` query.

flowmachine/flowmachine/features/subscriber/active_subscribers.py
```diff
@@ -0,0 +1,123 @@
...
+	ORDER BY subscriber, datetime
+), seen_on_days AS(
+	SELECT DISTINCT ON (event_date, subscriber)
```
jc-harrison, line 91: I suspect it would be more efficient to implement this as
```
SELECT subscriber, datetime::date AS event_date
FROM ({self.events_table.get_query()}) AS tbl
GROUP BY subscriber, datetime::date
```
(and skip the 'ordered_events' CTE entirely)

flowmachine/flowmachine/features/subscriber/active_subscribers.py
```diff
@@ -0,0 +1,123 @@
...
+	subscriber, datetime::date as event_date
+	FROM ordered_events
+	ORDER BY subscriber
```
jc-harrison, line 94: Perhaps I'm missing an implementation detail, but I can't see a need for this to be ordered.

flowmachine/flowmachine/features/subscriber/active_subscribers.py
```diff
@@ -0,0 +1,123 @@
...
+), dates_of_interest AS (
+	SELECT i::date as dates_of_interest
+	FROM generate_series('{self._window_start :%Y-%m-%d}', '{self._end_date:%Y-%m-%d}', '1 day'::interval) AS i
+), active_as_of AS (
+	SELECT dates_of_interest, subscriber
+	FROM dates_of_interest
+	LEFT OUTER JOIN seen_on_days ON dates_of_interest.dates_of_interest = seen_on_days.event_date
+	ORDER BY dates_of_interest
```
jc-harrison, line 102: Again, I don't think this needs to be ordered (since you've got an `ORDER BY dates_of_interest` in the window below)

jc-harrison (commented):
Copying my comments across from the old PR

jc-harrison:
I'm wondering whether it would be better to rearrange the workload between ActiveSubscribers and UniqueActiveSubscribers. My thinking is that ActiveSubscribers could return just one day of the result it currently returns (i.e. "find all subscribers who were active on at least `active_days` days between `start_date` and `end_date`, with 'active' meaning 'active in at least `active_hours` hours during the day'"), and then UniqueActiveSubscribers could take a list of ActiveSubscribers queries (which in our case would be one for each day of the month, with `start_date = today-7` and `end_date=today`) and return the unique subscribers from all queries.

This would simplify ActiveSubscribers considerably (and generally speaking, a simpler query is more likely to be re-usable as a component of a different workflow), and UniqueActiveSubscribers could be a generic query for selecting the unique subscribers from any collection of queries with a `subscriber` column (which would be generally useful for composing subscriber subsets. We might already have something that can do this - I'll have a look).

Thingus:
If I've got you right here, you're proposing:

```
SELECT subscriber  
FROM events_table
WHERE start_date <= event_date <= end_date
AND active_hours_count >= active_hours
```
and then loop over that for each `day` in month, with `start_date = day-7, end_date = day`  for `UniqueActiveSubscribers`?

I'm concerned that this leads to a slower query, as we'll be visiting each row in `subscriber` up to 7 times (row 7, say, getting visited on days 7 through 14, row 8 getting visited on 8 through 15, and so on).

I agree that it is simpler, but I'm not sure if it's better in this context. What do you think?

greenape:
If you're thinking of it as a one shot query it'll be less efficient, but if you're going to cache the intermediate parts then it'll be more efficient, esp when you consider, say, rolling it forward by one day - you then don't need to recompute everything, and would most likely end up only needing to compute a single day and then do a fast query over the parts.

Thingus:
I think I follow; so you compute `active_subs_on_day` for each as day in range an intermediate, build `active_subs_in_period` by bolting those together (thus avoid recomputing them, even if we visit the same day again), then getting `count(subscriber) >= active_days`?

flowmachine/flowmachine/features/subscriber/active_subscribers.py
```diff
@@ -0,0 +1,73 @@
...
+        active_hours: int,
+        active_days: int,
+        subscriber_id: str = "msisdn",
+        events_tables: Optional[List[str]] = None,
+        subscriber_subset=None,
+    ):
+        self.start_date = start_date
+        self.end_date = end_date
+        self.active_hours = active_hours
+        self.sub_id_column = subscriber_id
```
jc-harrison, line 32: Parameters that are just getting passed down to the sub-queries (sub_id_column and events_tables here) don't necessarily need to be stored as attributes, if they're not directly used within this query.

I'd also say it would be preferable to name these parameters 'subscriber_identifier' and 'tables', for consistency with most (admittedly not all) other queries that accept these parameters. (Side note: I think the fact that we have so many queries that accept the same set of parameters, which all get passed down to an underlying `EventsTablesUnion` query, is a sign that we could do with some refactoring. But that's something for another time, and will be easier to refactor if the names of these parameters are reasonably consistent to start with).

flowmachine/flowmachine/features/subscriber/active_subscribers.py
```diff
@@ -0,0 +1,73 @@
...
+
+        self.hour_queries = [
+            TotalActivePeriodsSubscriber(
+                start=day,
+                total_periods=24,
+                period_length=1,
+                period_unit="hours",
+                table=self.events_tables,
+                subscriber_identifier=self.sub_id_column,
+                subscriber_subset=self.events_table_query,
```
jc-harrison, line 53: I think you want `subscriber_subset=subscriber_subset` here, and `self.events_tables_query` is not required any more.

flowmachine/flowmachine/features/subscriber/rolling_count_threshold_subscribers.py
```diff
@@ -0,0 +1,61 @@
...
+from typing import List
+from dateutil.rrule import DAILY, rrule
+from dateutil.relativedelta import relativedelta
+
+from flowmachine.core.mixins.exposed_datetime_mixin import ExposedDatetimeMixin
+from flowmachine.core.query import Query
+from flowmachine.features.subscriber.active_subscribers import ActiveSubscribers
+
+
+class RollingCountThresholdSubscribers(ExposedDatetimeMixin, Query):
```
jc-harrison, line 11: I'm questioning whether this query in this form is actually necessary. I guess this comes back to the question of function vs purpose. We could have a query that just takes a list of `ActiveSubscribers` queries, and returns the set of unique subscribers from the union of those queries - that would be functionally equivalent to this, but more general-purpose in that a user could pass in any list of queries (such as active subscribers during each of three disconnected periods, rather than a rolling window).

If the process of defining the rolling-window set of `ActiveSubscriber` queries is too much of a burden to leave for the user to implement outside of FlowKit (by which I essentially mean "if the rolling-window set of `ActiveSubscriber` queries is something we'll be using regularly"), I'd argue this functionality would be better-placed in a FlowClient helper function (similar to the `modal_location_from_dates` flowclient function, as a convenience alternative to `modal_location`).

flowmachine/flowmachine/features/subscriber/rolling_count_threshold_subscribers.py
```diff
@@ -0,0 +1,61 @@
...
+
+        super().__init__()
+
+    @property
+    def column_names(self) -> List[str]:
+        return ["subscriber"]
+
+    def _make_query(self):
+        sql = f"""
+        SELECT DISTINCT subscriber
```
jc-harrison, line 57: As a general rule, in PostgreSQL it's more efficient to use
```
SELECT subscriber
FROM ...
GROUP BY subscriber
```
than
```
SELECT DISTINCT subscriber
FROM ...
```
(for no particularly good reason; just the way they're implemented)

jc-harrison (commented):
A few comments here; I'll have another look tomorrow.

flowmachine/flowmachine/core/mixins/exposed_datetime_mixin.py
```diff
@@ -0,0 +1,34 @@
+from datetime import date, datetime
+
+
+class ExposedDatetimeMixin:
```
greenape, line 4: Can this go to a separate PR and be accompanied by tests please? See also: https://github.com/Flowminder/FlowKit/blob/008d7f4e9c720629d3b3d45db40c200899078b05/flowmachine/flowmachine/utils.py#L57 which harmonises date inputs to `"YYYY-MM-DD HH:MM:SS"` form.

flowmachine/flowmachine/features/subscriber/active_subscribers.py
```diff
@@ -0,0 +1,73 @@
...
+        self.events_table_query = EventsTablesUnion(
+            self.start_date,
+            self.end_date,
+            tables=events_tables,
+            subscriber_identifier=subscriber_id,
+            columns=[subscriber_id, "datetime"],
+            subscriber_subset=subscriber_subset,
+        )
+
+        self.hour_queries = [
```
greenape, line 45: Don't really need to hold onto this as an attribute.

flowmachine/flowmachine/features/subscriber/active_subscribers.py
```diff
@@ -0,0 +1,73 @@
...
+            TotalActivePeriodsSubscriber(
+                start=day,
+                total_periods=24,
```
greenape, line 48: Can we expose these in the top level query and not assume that the outer period is a day, and the inner one an hour? Also can we make the hours param available - much more flexible that way.

flowmachine/flowmachine/features/subscriber/rolling_count_threshold_subscribers.py
greenape: Assuming the `ActiveSubscriber` queries have only one column, `union` with `all=False` will actually do the whole thing really.

jc-harrison: That's true. We'd still need to make that functionality accessible through the API in some way.

I wonder whether we could push this functionality down into the subscriber subsetter (so that the `subscriber_subset` parameter could take a list of queries, and the unioning and uniqueification is handled by the subsetter query)?

greenape: Links to #1009 on the api exposure.

We definitely _could_ pass it down, but whether one should is maybe another question. What if you actually _want_ to sample-with-replacement, or something?

jc-harrison: True. It seems to me the simplest option would be to add a new `UniqueSubscribersFromQueries` query which basically just wraps the `union(all=False)` functionality (we could make handle queries with additional columns as well, which would make it feel a bit less redundant). Then we can expose that as an input to the `subscriber_subset` parameter through the API. How does that sound?

greenape: That sounds reasonable to me. I don't think one is compelled to use the union class here btw - would also seem totally reasonable to do it in sql (probably better, because you don't then have a heavily nested union, and can do a select subscriber from ... on the input queries).

flowmachine/flowmachine/features/subscriber/active_subscribers.py
```diff
@@ -0,0 +1,79 @@
...
+            TotalActivePeriodsSubscriber(
+                start=day,
+                total_periods=total_periods,
+                period_length=period_length,
+                period_unit=period_unit,
+                table=tables,
+                subscriber_identifier=subscriber_identifier,
+                subscriber_subset=subscriber_subset,
+            ).numeric_subset("value", low=active_hours, high=total_periods)
+            for day in rrule(DAILY, dtstart=self._start_dt, until=self._end_dt)
```
greenape, line 61: Be nice if this was 'a period duration greater than the total active periods duration' rather than daily?

jc-harrison: Specifically, this period needs to be `total_periods * period_length`, right?

flowmachine/flowmachine/features/subscriber/active_subscribers.py
```diff
@@ -0,0 +1,181 @@
...
+
+SubscriberSubsetType = NewType("SubscriberSubsetType", Union[str, list, Query, "Table"])
+
+
+class ActiveSubscribers(ExposedDatetimeMixin, Query):
+    """
+    Class that represents subscribers seen to be active.
+
+    To determine this, we regard the timeframe between `start_date` and `end_date` as the major period.
+    We then break the major period into `minor_period_count` minor periods.
```
jc-harrison, line 25: I'm getting a bit confused by all the periods here, but I don't think this docstring is quite describing the correct correspondence between the parameters and the query. We've got three different period lengths here:
- The innermost period ( = `minor_period_length` * `period_unit`),
- The middle-length period, which I think you're calling "minor period" here ( = `minor_period_count` * "innermost period"),
- The total period from `start_date` to `end_date`, which I think you're calling "major period".

On that basis, `minor_period_count` is the number of "innermost periods" per "minor period", and _not_ the number of "minor periods" per "major period". On the other hand, we could equally well call the innermost period "minor period" and the middle-length period "major period", in which case the description of `minor_period_count` here is correct but "we regard the timeframe between `start_date` and `end_date` as the major period" is incorrect.

flowmachine/flowmachine/features/subscriber/active_subscribers.py
```diff
@@ -0,0 +1,181 @@
...
+
+    A subscriber is considered to be active in a minor period if they are seen at least `minor_period_
+    threshold` times within that period.
```
jc-harrison, line 28: I think we want to say "seen in at least `minor_period_threshold` distinct 'innermost periods' within that minor period", or something like that,.

flowmachine/flowmachine/features/subscriber/active_subscribers.py
```diff
@@ -0,0 +1,181 @@
...
+    for subscribers who are active at least `minor_period_threshold` hours throughout the day.
+
+    Examples
+    --------
+    Returns subscribers who were active on at least three hours between 2016-01-01 and 2016-01-02
+
+
+    >>>     active_subscribers = ActiveSubscribers(
+                start_date=date(year=2016, month=1, day=1),
+                end_date=date(year=2016, month=1, day=2),
```
jc-harrison, line 70: Personally I find this first example more confusing than helpful. This could be achieved without an `ActiveSubscribers` query, just using `TotalActivePeriodsSubscriber` with `numeric_subset`, and by counting subscribers active on 1 out of 1 days it's hard to see all the layers of the query. I'd suggest maybe getting rid of this one and just keeping the other two examples.

flowmachine/flowmachine/features/subscriber/active_subscribers.py
```diff
@@ -0,0 +1,181 @@
...
+        period_unit: Literal["days", "hours", "minutes"] = "hours",
+    ):
+        self.start_date = start_date
+        self.end_date = end_date
+        self.minor_period_threshold = minor_period_threshold
+        self.sub_id_column = subscriber_identifier
+        self.events_tables = tables
+        self.major_period_threshold = major_period_threshold
+
+        self.events_table_query = EventsTablesUnion(
```
jc-harrison, line 132: I don't think this is needed any more.

flowmachine/flowmachine/features/subscriber/active_subscribers.py
```diff
@@ -0,0 +1,181 @@
...
+            subscriber_identifier=subscriber_identifier,
+            columns=[subscriber_identifier, "datetime"],
+            subscriber_subset=subscriber_subset,
+        )
+
+        date_generator = rr.rrule(
+            self.period_to_rrule_mapping[period_unit],
+            interval=minor_period_count * minor_period_length,
+            dtstart=self._start_dt,
+            until=self._end_dt - relativedelta(seconds=1),
```
jc-harrison, line 145: If the total length of the major period from `start_date` to `end_date` is an exact integer number of minor periods then this is fine. If it's not, though, then the last `TotalActivePeriodsSubscriber` would cover a minor period that extends past `end_date`.  Alternatively, if we did `until=self._end_dt - minor_period_count * minor_period_length` then the last `TotalActivePeriodsSubscriber` would cover the last minor period that fits entirely within the major period (so there would be a gap of less than one minor period at the end, not covered by any `TotalActivePeriodsSubscriber` query).

I'm really not sure which of these two options we would want - the intention is that the major period should consist of an integer number of minor periods, but the parameters allow for that not to be the case. A third option would be to enforce the "integer minor periods per major period" behaviour by getting rid of the `end_date` parameter and replacing it with a `total_minor_periods` parameter (equivalent to the `TotalActivePeriodsSubscriber` parameters for the inner periods) - that would be cleaner, in a way, but from a user perspective it might be nicer to specify the end date than to have to work out the correct number of periods.

jc-harrison (commented):
A few comments here - a common theme among them is that I'm finding the "minor period" and "major period" naming a bit confusing, and we might need to rethink the names a bit.

flowmachine/flowmachine/features/subscriber/active_subscribers.py
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    def __init__(
+        self,
+        start_date: Union[date, datetime, str],
+        end_date: Union[date, datetime, str],
+        minor_period_threshold: int,
+        major_period_threshold: int,
+        subscriber_identifier: Optional[str] = "msisdn",
+        tables: Optional[List[str]] = None,
```
jc-harrison, line 119: I think it would be good to expose the `hours` parameter here as well (and pass it down to the `TotalActivePeriodsSubscriber` queries) - sometimes we might want to use this query to find subscribers who are regularly active specifically between certain hours.

flowmachine/flowmachine/features/subscriber/active_subscribers.py
jc-harrison: Having slept on this, I think my suggestion would be to have parameters:
- `start_date`
- `total_major_periods`
- `minor_periods_per_major_period`
- `minor_period_length`
- `period_unit`
(where, in the "active during X distinct hours on Y days over the total period" case, I'm referring to '1 hour' as the minor period and '1 day' as the major period - I'm fully open to other suggestions on names).

When we expose this through the API, we could add a convenience function in flowclient to handle the (probably most commonly used) case of "active X hours on Y days" (i.e. the defaults `period_unit='hour'`, `minor_period_length=1` and `minor_periods_per_major_period=24`). This convenience function could take `start_date` and `end_date` as parameters, and internally calculate the correct value for `total_major_periods`.

flowmachine/flowmachine/features/subscriber/active_subscribers.py
```diff
@@ -0,0 +1,174 @@
...
+
+class ActiveSubscribers(ExposedDatetimeMixin, Query):
+    """
+    Class that represents subscribers seen to be active across a search range
+
+    The search range is split into major periods. Each major period is split into minor periods.
+
+    A single minor period is `minor_period_length * period_units` long
+
+    A single major period is `minor_periods_per_major_period` long
```
jc-harrison, line 28:
```suggestion
    A single major period is `minor_periods_per_major_period` minor periods long
```

flowmachine/flowmachine/features/subscriber/active_subscribers.py
```diff
@@ -0,0 +1,174 @@
...
+SubscriberSubsetType = NewType("SubscriberSubsetType", Union[str, list, Query, "Table"])
+
+
+class ActiveSubscribers(ExposedDatetimeMixin, Query):
+    """
+    Class that represents subscribers seen to be active across a search range
+
+    The search range is split into major periods. Each major period is split into minor periods.
+
+    A single minor period is `minor_period_length * period_units` long
```
jc-harrison, line 26:
```suggestion
    A single minor period is `minor_period_length * period_unit` long
```

flowmachine/flowmachine/features/subscriber/active_subscribers.py
```diff
@@ -0,0 +1,170 @@
...
+
+    A single major period is `minor_periods_per_major_period` long
+
+    This makes the search range between `start_date` and `start_date +
+    (period_units * minor_period_length * minor_periods_in_major_period)`
```
jc-harrison, line 31:
```suggestion
    (period_unit * minor_period_length * minor_periods_per_major_period * total_major_periods)`
```

flowmachine/flowmachine/features/subscriber/active_subscribers.py
```diff
@@ -0,0 +1,170 @@
...
+
+    A subscriber is considered to be active in a major period if they are seen in at least `minor_period_
+    threshold` minor periods within that period
+
+    A subscriber is considered active over the entire serach range if they are active in at least
```
jc-harrison, line 36:
```suggestion
    A subscriber is considered active over the entire search range if they are active in at least
```

flowmachine/flowmachine/features/subscriber/active_subscribers.py
```diff
@@ -0,0 +1,170 @@
...
+        total_major_periods: int,
+        minor_period_threshold: int,
+        major_period_threshold: int,
+        period_unit: Literal["days", "hours", "minutes"] = "hours",
+        subscriber_identifier: Optional[str] = "msisdn",
+        tables: Optional[List[str]] = None,
+        subscriber_subset: Optional[SubscriberSubsetType] = None,
+        hours: Optional[Tuple[int, int]] = None,
+    ):
+        self.start_date = start_date
```
jc-harrison, line 123: Given that `ExposedDatetimeMixin` is used here (so this query does have an `end_date` attribute), we could set the end date attribute to the value `start_date + (period_unit * minor_period_length * minor_periods_per_major_period * total_major_periods)`.

flowmachine/flowmachine/features/subscriber/active_subscribers.py
```diff
@@ -0,0 +1,170 @@
...
+        self.minor_period_threshold = minor_period_threshold
+        self.sub_id_column = subscriber_identifier
```
jc-harrison, line 125: Up to you on this one, but I just want to note:

Regarding which parameters to set as attributes, the only essential thing is that no two queries with different parameters end up with exactly the same attributes (because then they would both have the same query ID). In this case, `major_period_threshold` is the only parameter that _needs_ to be set as an attribute, because all the others are captured within the `major_period_queries` sub-queries.

There's no problem with setting other parameters as attributes too, and in some cases it may be useful for a user (or possibly another query using this as a sub-query) to check these attributes rather than needing to search through the sub-queries. For example, I think `start_date` is a useful attribute here even though it's not essential (and same for `end_date`, as per my comment above). On the other hand, I personally wouldn't keep the `sub_id_column` attribute. I'm fairly neutral on the various major/minor period parameters (other than `major_period_threshold`) - I'll let you make the call on those.

jc-harrison (approved):
A few minor tweaks to the docstring here, and a couple of other comments that you can address or not as you see fit. Other than that, I'm happy.

PR #4482, opened 2021-11-02
Home location

Thingus:
### I have:

### Description
Implements home_location_monthly (name tbd). Dependent on #4481

### DIFF
--- a/flowmachine/flowmachine/features/subscriber/home_location_monthly.py
+++ b/flowmachine/flowmachine/features/subscriber/home_location_monthly.py
@@ -35,10 +35,10 @@ class HomeLocationMonthly(Query):
         ref_location: Union["HomeLocationMonthly", None] = None,
         **kwargs,
     ):
-        super().__init__(*args, **kwargs)
+
         self.window_start = window_start
         self.window_stop = window_stop
-        self.ref_location = copy(ref_location)
+        self.ref_location = ref_location
         self.agg_unit = agg_unit
         self.known_threshold = known_threshold
         self.unknown_threshold = unknown_threshold
@@ -65,6 +65,8 @@ class HomeLocationMonthly(Query):
         # self.daily_location_frequency = LocationVisits(
         #     self.last_locations  # Ask Jono about this
         # )
+
+        super().__init__(*args, **kwargs)
 
     @property
     def column_names(self) -> List[str]:
--- a/flowmachine/flowmachine/features/subscriber/home_location_monthly.py
+++ b/flowmachine/flowmachine/features/subscriber/home_location_monthly.py
@@ -42,8 +42,8 @@ class HomeLocationMonthly(Query):
         self.window_stop = window_stop
         self.ref_location = ref_location
         self.agg_unit = agg_unit
-        self.home_this_month = home_last_month
-        self.home_last_month = home_this_month
+        self.home_this_month = home_this_month
+        self.home_last_month = home_last_month
         self.modal_lookback = modal_lookback
 
         self.active_subs = ActiveSubscribers(
@@ -123,6 +123,7 @@ class HomeLocationMonthly(Query):
 
         last_locations_clause = ""
 
+        # NOTE: last_locations seems to be dropping the final record? This is weird.
         for last_location in self.last_locations:
             last_locations_clause += f"""
 SELECT subscriber, pcod, '{last_location.start}' AS day
@@ -183,7 +184,7 @@ known_homes AS (
 
         sql += f"""
 unknown_homes AS(
-	SELECT subscriber, 'unknown' AS pcod
+	SELECT DISTINCT(subscriber), 'unknown' AS pcod
 	FROM location_histogram
 	WHERE subscriber NOT IN (SELECT subscriber FROM known_homes)
 )
--- a/flowmachine/flowmachine/features/subscriber/home_location_monthly.py
+++ b/flowmachine/flowmachine/features/subscriber/home_location_monthly.py
@@ -6,17 +6,12 @@ from dateutil.relativedelta import relativedelta
 from dateutil.rrule import rrule, DAILY
 from copy import copy
 
-from flowmachine.features.subscriber.daily_location import daily_location
 from flowmachine.features.subscriber.modal_location import ModalLocation
-from flowmachine.features.subscriber.unique_active_subscribers import (
-    UniqueActiveSubscribers,
-    ActiveSubscribers,
-)
-from flowmachine.features.subscriber.last_location import LastLocation
 from flowmachine.features.subscriber.location_visits import LocationVisits
+from flowmachine.core.errors import InvalidSpatialUnitError
 
 
-class HomeLocationMonthly(Query):
+class MajorityLocation(Query):
 
     """For each subscriber, if their modal last location was their last location on more than X days this month
     (choose X=15 here), then choose this as their home location. If not, but it was their last location on more than Y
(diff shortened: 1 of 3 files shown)


flowmachine/flowmachine/features/subscriber/home_location_monthly.py
```diff
@@ -0,0 +1,198 @@
...
+
+    """For each subscriber, if their modal last location was their last location on more than X days this month
+    (choose X=15 here), then choose this as their home location. If not, but it was their last location on more than Y
+     days this month (Y<X; choose Y=10 here) AND it is the same as their “reference location” (in our case reference
+     location is modal location and last location on more than X days last month), then choose this as their home
+     location. Otherwise set their home location to “unlocatable” (a special value, not NULL)."""
+
+    def __init__(
+        self,
+        window_start: Union[str, datetime],
```
jc-harrison, line 29: Rather than taking a date range and constructing the sub-queries internally, I think it would be better to take three queries (ModalLocation, LocationVisits and ActiveSubscribers) as parameters, because:
- the date ranges aren't necessarily the same for each of these queries (e.g. in the original query plan, modal location is over 40 days, but the daily location visit counts are over one month)
- this would make the query more flexible (e.g. if we want to use an additional rolling filter before calculating the modal location, that could be done without changing the implementation of this query)
- the ActiveSubscribers query here is just a subscriber subset - passing the query in directly as a 'subscriber_subset' parameter would mean users could choose to use an entirely different subscriber subset if preferred.

jc-harrison: Actually, in this situation the ActiveSubscribers query wouldn't be needed in this query at all - the subscriber subset would be applied to each of the other queries before passing them into this query.

flowmachine/flowmachine/features/subscriber/home_location_monthly.py
```diff
@@ -0,0 +1,198 @@
...
+"""
+
+        last_locations_clause = last_locations_clause.rstrip("UNION ALL\n")
+
+        sql = f"""
+WITH last_locations AS (
+    {last_locations_clause}
+), modal_locations AS (
+    {self.modal_locations.get_query()}
+), location_histogram AS (
```
jc-harrison, line 141: location_histogram can be replaced with a `LocationVisits` sub-query (which would avoid having to re-implement the union of daily locations within this query).

flowmachine/flowmachine/features/subscriber/home_location_monthly.py
```diff
@@ -0,0 +1,198 @@
...
+     location is modal location and last location on more than X days last month), then choose this as their home
+     location. Otherwise set their home location to “unlocatable” (a special value, not NULL)."""
+
+    def __init__(
+        self,
+        window_start: Union[str, datetime],
+        window_stop: Union[str, datetime],
+        spatial_unit: AnySpatialUnit,
+        home_this_month: int,
+        home_last_month: int,
```
jc-harrison, line 33: The names 'home_this_month' and 'home_last_month' don't really capture what these parameters are - perhaps something like 'threshold_for_non_reference_location' and 'threshold_for_reference_location' would be more descriptive?

Reply from @Thingus:
> I've been calling them `known_location_threshold` and `unknown_location_threshold`, or just `X` and `Y` after the docstring - but none of these quite capture what it is that those are doing. Will think on it.

flowmachine/flowmachine/features/subscriber/home_location_monthly.py
```diff
@@ -0,0 +1,198 @@
...
+WITH last_locations AS (
+    {last_locations_clause}
+), modal_locations AS (
+    {self.modal_locations.get_query()}
+), location_histogram AS (
+	SELECT subscriber, pcod, count(pcod) as y
+	FROM last_locations
+	GROUP BY subscriber, pcod
+), this_month_known_homes AS (
+	SELECT modal_locations.subscriber, modal_locations.pcod
```
jc-harrison, line 146: `pcod` is not necessarily the right column to use here (depending on the spatial unit). You'll want to use `self.spatial_unit.location_id_columns`.

flowmachine/flowmachine/features/subscriber/home_location_monthly.py
```diff
@@ -0,0 +1,198 @@
...
+), location_histogram AS (
+	SELECT subscriber, pcod, count(pcod) as y
+	FROM last_locations
+	GROUP BY subscriber, pcod
+), this_month_known_homes AS (
+	SELECT modal_locations.subscriber, modal_locations.pcod 
+	FROM modal_locations
+		INNER JOIN  location_histogram ON location_histogram.subscriber = modal_locations.subscriber
+	WHERE modal_locations.pcod = location_histogram.pcod
+	AND modal_locations.subscriber = location_histogram.subscriber
```
jc-harrison, line 150: I think this line is redundant - duplicates the join condition

flowmachine/flowmachine/features/subscriber/home_location_monthly.py
```diff
@@ -0,0 +1,198 @@
...
+	AND location_histogram.y >= {self.home_this_month}
+), """
+
+        if self.ref_location:
+            sql += f"""
+reference_locations AS (
+    SELECT subscriber, pcod
+    FROM ({self.ref_location.get_query()}) AS tbl
+), last_month_known_homes AS (
```
jc-harrison, line 159: Personally, I'd be tempted to combine 'this_month_known_homes' and 'last_month_known_homes', on the basis that the query result should have one row for each row in 'modal_locations' so it feels neater to avoid the `subscriber NOT IN (SELECT ...)`, and only join 'modal_locations' to 'location_histogram' once.

Something like this (square brackets mark reference_location-dependent lines):
```
  SELECT modal_locations.subscriber, modal_locations.pcod
  FROM modal_locations
  INNER JOIN location_histogram USING ( subscriber, pcod )
[ LEFT JOIN reference_locations USING (subscriber)  ]
  WHERE location_histogram.y >= {self.home_this_month}
[ OR ( modal_locations.pcod = reference_locations.pcod AND location_histogram.y >= {self.home_last_month} )  ]
```

It's a matter of preference though, I suppose - what seems neater to me might seem less so to others.

jc-harrison: Could even combine 'unknown_homes' into this as well, by moving the where condition into a `CASE WHEN` in the select:
```
  SELECT modal_locations.subscriber,
         CASE WHEN ( location_histogram.y >= {self.home_this_month} )
[             OR ( modal_locations.pcod = reference_locations.pcod AND location_histogram.y >= {self.home_last_month} )  ]
              THEN modal_locations.pcod
              ELSE 'unknown'
         END AS pcod
  FROM ...
```

But again, personal preference...

flowmachine/flowmachine/features/subscriber/home_location_monthly.py
```diff
@@ -0,0 +1,198 @@
...
+    SELECT subscriber, pcod
+    FROM this_month_known_homes
+    UNION ALL
+    SELECT subscriber, pcod
+    FROM last_month_known_homes
+), """
+
+        sql += f"""
+unknown_homes AS(
+	SELECT DISTINCT(subscriber), 'unknown' AS pcod
```
jc-harrison, line 187: Following on from my earlier comment about using `self.spatial_unit.location_id_columns` instead of `pcod` to handle spatial units other than admin, I've just realised setting the 'unknown' value could be problematic if the spatial unit doesn't have a single text field as the location ID (e.g. VersionedCell, which has two columns, or LonLat, which has a pair of coordinates). Not sure how to cleanly handle that situation.

flowmachine/flowmachine/features/subscriber/home_location_monthly.py
```diff
@@ -0,0 +1,198 @@
...
+from flowmachine.features.subscriber.modal_location import ModalLocation
+from flowmachine.features.subscriber.unique_active_subscribers import (
+    UniqueActiveSubscribers,
+    ActiveSubscribers,
+)
+from flowmachine.features.subscriber.last_location import LastLocation
+from flowmachine.features.subscriber.location_visits import LocationVisits
+
+
+class HomeLocationMonthly(Query):
```
jc-harrison, line 19: I'm not keen on the name 'HomeLocation' - there are many different ways home location could be estimated, so it would be good to have a more descriptive name. I haven't come up with a better name yet, though - I'll come back to you on that one.

(Sorry, I realise the name came from my diagram...)

Reply from @Thingus:
> How about `ReferencedModalThresholdLocation`?

jc-harrison (commented):
Copying my comments across from the old PR

Thingus:
AFter discussion with the analysis team + James, we're proposing a new design for this query.
 * The new design calls for a subscriber's home to be considered known in a month if they're there more than 50% of the time
 * A home location can be known if it appears some `threshold` times this month, and that location was known last month.
 * We can `COALESCE` across

PR #4486, opened 2021-11-04
Implementing UniqueSubscribersFromQueries

### LINKED ISSUE #4490: UniqueValuesFromQuery
Repeated small snippets of SQL to deduplicate data

A class that deduplicates data from multiple queries

Writing the SQL, but that skips caching steps

Thingus:
Closes #4490

### I have:

### Description

Creates UniqueSubscribersFromQueries (see https://github.com/Flowminder/FlowKit/pull/4481/files/157ed94a748f157aaa1b553b953f021877917072#r742694002)


flowmachine/flowmachine/features/subscriber/unique_subscribers_from_queries.py
```diff
@@ -0,0 +1,29 @@
...
+        return ["subscribers"]
+
+    def _make_query(self):
+
+        sql = ""
+        for query in self.query_list:
+            sql += f"""
+            SELECT subscriber FROM
+            ({query.get_query()}) as tbl
+            UNION
```
jc-harrison, line 26: Could use the string `join` method here to avoid needing to `rstrip` the last "UNION"

flowmachine/flowmachine/features/subscriber/unique_subscribers_from_queries.py
```diff
@@ -0,0 +1,28 @@
...
+
+    def __init__(self, query_list: List[Query]):
+        self.query_list = query_list
+        super().__init__()
+
+    @property
+    def column_names(self) -> List[str]:
+        return ["subscribers"]
+
+    def _make_query(self):
```
jc-harrison, line 19: This will work fine provided there are at least two queries, but won't de-duplicate if there's only one query, which is not the behaviour a user would intuitively expect.

Could instead do
```
SELECT subscriber
FROM (
    SELECT subscriber FROM ...
    UNION ALL
    ...
) unioned
GROUP BY subscriber
```
which would work for any number of input queries.

flowmachine/flowmachine/features/subscriber/unique_subscribers_from_queries.py
```diff
@@ -0,0 +1,28 @@
...
+from flowmachine.core.query import Query
+
+
+class UniqueSubscribersFromQueries(Query):
+    """
+    Given a list of queries with a 'subscriber' column, returns a table of unique subscribers.
+    """
+
+    def __init__(self, query_list: List[Query]):
+        self.query_list = query_list
```
jc-harrison, line 12: Perhaps best to convert this to a set when setting the attribute, so that the query ID doesn't depend on the order of queries in the input list.

jc-harrison: (At least, I'd hope using a set here would have the desired effect. I'm not 100% sure...)

greenape: Shouldn't matter - they'll get sorted by their query ids by the id builder.

flowmachine/flowmachine/features/subscriber/unique_subscribers_from_queries.py
greenape: Without the `ALL` tho, right? You'd sort of hope it resolves to the same plan _anyway_ but I bet it doesn't.

jc-harrison: Result should be the same with or without the `ALL` if we then do a `GROUP BY`. My thinking was that if there's any difference in the query plans, the `UNION ALL` version would probably be the faster one, but I could be wrong.

Thingus: Wouldn't this lead to the deep nesting we were hoping to avoid by going around query.union()?

jc-harrison: Shouldn't do - it's only one extra level of nesting, regardless of the number of queries. And it's an extra level of nesting in the SQL, but not in the flowmachine objects, so it shouldn't matter.

flowmachine/flowmachine/features/subscriber/unique_subscribers_from_queries.py
jc-harrison: Ah, so they will! OK, this is fine as it is, then.

flowmachine/flowmachine/features/subscriber/unique_subscribers_from_queries.py
Thingus: Ah, I misunderstood your ellipsis. Will sort.

flowmachine/flowmachine/features/subscriber/unique_subscribers_from_queries.py
```diff
@@ -0,0 +1,52 @@
...
+    Parameters
+    ----------
+    query_list: List[Query]
+        A list of Query objects, each of which must include a `subscriber` column
+
+    """
+
+    # Review question: Does this need to be 'subscriber' or can we make the column to select a parameter?
+
+    def __init__(self, query_list: List[Query]):
```
Thingus, line 28: Is it worth making the params for this `(self, query_list, column)`and letting it just return unique values in whatever we pass it?

flowmachine/flowmachine/features/subscriber/unique_subscribers_from_queries.py
```diff
@@ -0,0 +1,52 @@
...
+        super().__init__()
+
+    @property
+    def column_names(self) -> List[str]:
+        return ["subscriber"]
+
+    def _make_query(self):
+
+        union_stack = "\nUNION ALL\n".join(
+            [
```
greenape, line 39: Don't actually need the []s here

flowmachine/flowmachine/features/subscriber/unique_subscribers_from_queries.py
greenape: Sounds reasonable to me (but change the name to UniqueFromQueries or similar)

jc-harrison: Yes, I think that's a good idea. We could even accept a list of `columns` - that way we could use the query to get the set of all distinct (subscriber, location) pairs from several queries, for example.

I'd suggest we should move this to 'flowmachine/features/utilities', since it's not specifically subscriber-related. And perhaps `UniqueValuesFromQueries` would be an appropriate name?

There's a separate question of how we want to _expose_ this through the API - we may decide that for now it will only be exposed with `column` hard-coded to `"subscriber"`, as an allowed input to other queries' `subscriber_subset` parameters. But that's a separate question, and can be handled in a separate PR.

greenape:
Changelog entry pls.

flowmachine/flowmachine/features/subscriber/unique_subscribers_from_queries.py
Thingus: Name changed, params updated

flowmachine/flowmachine/features/subscriber/unique_values_from_queries.py
```diff
@@ -0,0 +1,51 @@
...
+    Parameters
+    ----------
+    query_list: List[Query]
+        A list of Query objects, each of which must include a `subscriber` column
+
+    """
+
+    def __init__(self, query_list: List[Query], column_names: List[str]):
+        self.query_list = query_list
+        self._column_names: List[str] = column_names
```
jc-harrison, line 28: Probably a good idea to check here that all of the queries in `query_list` include the requested columns (based on their `column_names` properties).

Thingus: There a specific exception in FM for this situation, or just raise `ValueError`?

Thingus: Ah, just found `MissingColumnsError` seems like it could be a good fit

jc-harrison: Ah, good find! I couldn't find an example that raises anything other than `ValueError`, but `MissingColumnsError` looks good (although it looks like it's not an exception we're currently using anywhere in the codebase...)

flowmachine/flowmachine/features/subscriber/unique_subscribers_from_queries.py
jc-harrison: Thanks! Could you move it into 'flowmachine/features/utilities' as well, please?

flowmachine/flowmachine/features/subscriber/unique_values_from_queries.py
```diff
@@ -24,6 +25,11 @@ class UniqueValuesFromQueries(Query):
     """
 
     def __init__(self, query_list: List[Query], column_names: List[str]):
+
+        for query in query_list:
+            if False in [name in query.column_names for name in column_names]:
```
jc-harrison, line 30: Perhaps more readable to do `if any(name not in query.column_names for name in column_names)`,
or `if not all(name in query.column_names for name in column_names)`? (I'm just being picky)

Thingus: Nah, makes sense. I was casting around for that but blanked on it >_>

flowmachine/flowmachine/features/subscriber/unique_values_from_queries.py
jc-harrison: (Apparently I can't start a second comment thread against the same line of code, so I'll unresolve this one to add a new comment)

Might be good to do `self._column_names = list(column_names)` here, to enforce the same internal attribute regardless of the type of iterable passed in.

jc-harrison: Oh, wait, I can do it now! I'll move this comment to the appropriate place.

flowmachine/flowmachine/features/utilities/unique_values_from_queries.py
```diff
@@ -0,0 +1,57 @@
...
+    Class representing unique values in  across a set of queries
+
+    Parameters
+    ----------
+    query_list: List[Query]
+        A list of Query objects, each of which must include a `subscriber` column
+
+    """
+
+    def __init__(self, query_list: List[Query], column_names: List[str]):
```
jc-harrison, line 27:
```suggestion
    def __init__(self, query_list: List[Query], *, column_names: List[str]):
```
Good to make the 'column_names' arg keyword-only, I think

Thingus: Probably not a bad idea to do both

jc-harrison (dismissed):
Looking good. I think it would be nice to have a test that this de-dupes (ideally in both the single-query and multiple-queries cases), and also test that `MissingColumnsError` gets raised as expected.

flowmachine/flowmachine/features/utilities/unique_values_from_queries.py
jc-harrison: Yes, no harm in making them both keyword-only. I think you're right that it doesn't make sense to have a default for `column_names`.

flowmachine/flowmachine/features/utilities/unique_values_from_queries.py
```diff
@@ -0,0 +1,70 @@
...
+    """
+
+    def __init__(
+        self,
+        *,
+        query_list: Union[Query, List[Query]],
+        column_names: Union[str, List[str]],
+    ):
+
+        if issubclass(type(query_list), Query):
```
jc-harrison, line 34:
```suggestion
        if isinstance(query_list, Query):
```

flowmachine/flowmachine/features/utilities/unique_values_from_queries.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+class UniqueValuesFromQueries(Query):
+    """
+    Class representing unique values in  across a set of queries
+
+    Parameters
+    ----------
+    query_list: List[Query]
+        A list of Query objects, each of which must include a `subscriber` column
```
jc-harrison, line 23: Docstring needs updating to include `column_names` and remove reference to 'subscriber'

flowmachine/tests/test_unique_values_from_queries.py
```diff
@@ -0,0 +1,44 @@
...
+        "b80699f8af4fa963c6f75eb6990556e1",
+    ]
+    assert get_dataframe(two_column_query).columns.tolist() == ["imei", "location_id"]
+    # location_id should dedupe more than location_id and imei, which should in turn dedupe more that the intitial query
+    assert (
+        len(get_dataframe(one_column_query))
+        < len(get_dataframe(two_column_query))
+        < len(get_dataframe(EventTableSubset(start="2016-01-01", stop="2016-01-03")))
+    )
+
```
jc-harrison, line 37: I think it would be good to have a test that uses a list of two or more queries - ideally checking that rows appearing in both queries get de-duped, but as a minimum just checking that the query works and returns a result when operating on multiple queries.

jc-harrison (commented):
Just a few more comments

flowmachine/tests/test_unique_values_from_queries.py
Thingus: Done:
```
    two_column_query = UniqueValuesFromQueries(
        query_list=[
            EventTableSubset(start="2016-01-01", stop="2016-01-03"),
            EventTableSubset(start="2016-01-04", stop="2016-01-06"),
        ],
        column_names=["imei", "location_id"],
    )
```

should be equivalent to

```
    equivalent_query = UniqueValuesFromQueries(
        query_list=[EventTableSubset(start="2016-01-01", stop="2016-01-06")],
        column_names=["imei", "location_id"],
    )
```

jc-harrison (approved):
:+1:

Thingus:
`test_quickstart` is falling over, but is working happily on my machine. Could this be a holdover from #4449?

PR #4493, opened 2021-11-08
Fix cannot uncache novel objs

### LINKED ISSUE #4008: Cache cleanup fails if it hits a custom class
If the cache cleanup op hits a new query subclass, it will fail because it can't be unpickled.

greenape:
Closes #4008 

### I have:

### Description

Deals with queries which were defined outside the server process (e.g. interactive session, defined in an imported module in an interactive session) being un-invalidatable outside the defining session by unpickling them to a stub class which does nothing but preserve their dependency links and query id.

I was hoping to deal with this by pickling them using dill, but that can't pickle them either outside of a `__main__`. One could also pickle to a stub where not pickleable, but I think this is marginally more robust because it doesn't rely on everyone immediately using an up to date FlowMachine.

### DIFF
--- a/flowmachine/flowmachine/core/cache.py
+++ b/flowmachine/flowmachine/core/cache.py
@@ -36,7 +36,26 @@ import structlog
 logger = structlog.get_logger("flowmachine.debug", submodule=__name__)
 
 
-def get_obj_or_stub(connection, query_id):
+def get_obj_or_stub(connection: "Connection", query_id: str):
+    """
+    Get a query object by ID if the object can be recreated. If it cannot
+    then get a stub which keeps the query's dependency relationships as recorded
+    in the database.
+
+    Parameters
+    ----------
+    connection : Connection
+        DB connection to get the object from
+    query_id : str
+        Unique identifier of the query
+
+    Returns
+    -------
+    Query
+        The query object if it can be recreated, or a 'stub' class which records
+        the dependencies it had if not.
+
+    """
     from flowmachine.core.query import Query
 
     class QStub(Query):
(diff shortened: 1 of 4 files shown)


flowmachine/flowmachine/core/cache.py
```diff
@@ -36,6 +36,52 @@
...
+        def _make_query(self):
+            pass
+
+        @property
+        def column_names(self):
+            pass
+
+    try:
+        return get_query_object_by_id(connection, query_id)
+    except (EOFError, ModuleNotFoundError, AttributeError) as exc:
```
Thingus, line 76: Do we want `UnpicklingError` here as well?

flowmachine/tests/test_cache.py
```diff
@@ -249,6 +253,80 @@ def test_retrieve(get_dataframe):
...
+        def _make_query(self):
+            return "select 1 as value"
+
+    q = NestTestQuery()
+    q_id = q.query_id
+    nested_id = q.nested.query_id
+    q.store(store_dependencies=True).result()
+    Query._QueryPool.clear()
+    yield q_id, nested_id
+
```
Thingus, line 313: Only extra test I can think of here is nesting a novel in a not-novel or vice versa

Thingus: Wait, novel in not-novel would never happen (if I've understood correctly) and from what I've seen, nesting two novel queries encapsulates nesting a not-novel in a novel, so disregard.

jc-harrison (approved):
:+1:

Thingus (approved):

PR #4523, opened 2021-11-12
Initial implementation of `MajorityLocation` and test

### LINKED ISSUE #4522: MajorityLocation
A query returning, for each subscriber, the location that subscriber visited more than 50% of the time.

Query.numerical_subset() will be used, but is not proportional - and this may be useful in furhter analysis

Thingus:
Closes #4522

### I have:

### Description
Introduces a new query, MajorityLocation, that, when supplied with a subscriber-location table or query, returns for each subscriber the location that subscriber appears more than 50% of the time. If there is no such location, the subscriber is dropped.


flowmachine/flowmachine/features/location/majority_location.py
```diff
@@ -0,0 +1,43 @@
+from typing import List
+
+from flowmachine.core import Query
+from flowmachine.features.utilities.subscriber_locations import BaseLocation
+
+
+# One of the spatial mixins here?
+class MajorityLocation(BaseLocation, Query):
```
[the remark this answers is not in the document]
Thingus: Makes the SQL a lot less convoluted as well.
Do we want it to return all of what goes in, or just subscriber and location?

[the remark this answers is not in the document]
jc-harrison: Good question. I think probably just subscriber and location - that way it's analogous to a daily location or modal location query.

flowmachine/flowmachine/features/location/majority_location.py
```diff
@@ -0,0 +1,35 @@
...
+
+    @property
+    def column_names(self) -> List[str]:
+        # This should be a reduction, so no new columns
+        return ["subscriber"] + self.subscriber_subset.spatial_unit.location_id_columns
+
+    def _make_query(self):
+        loc_id = ",".join(self.subscriber_subset.spatial_unit.location_id_columns)
+        sql = f"""
+WITH subscriber_subset AS (
```
greenape, line 24: So interesting to note that until quite recently (pg 12) CTEs were an optimization fence, and often to be avoided in favour of a subquery. Now, what it'll do is rewrite it as a subquery for you prior to planning in a case like this where there's only one use of the subquery - see https://www.postgresql.org/docs/12/queries-with.html for more (when we wrote very early versions of fk, we didn't know about the CTE behaviour, so we used it liberally, and then removed _all_ uses of it cos it generally led to dreadfully slow query plans).

flowmachine/flowmachine/features/location/majority_location.py
```diff
@@ -0,0 +1,35 @@
...
+from flowmachine.core.mixins import GeoDataMixin
+from flowmachine.features.utilities.subscriber_locations import BaseLocation
+
+
+class MajorityLocation(BaseLocation, Query):
+    def __init__(self, subscriber_subset: Query, weight_column):
+        if weight_column not in subscriber_subset.column_names:
+            raise ValueError("weight_column must exist in subscriber_subset")
+        self.subscriber_subset = subscriber_subset
+        self.weight_column = weight_column
```
jc-harrison, line 13: I think it would be worth doing `self.spatial_unit = subscriber_subset.spatial_unit` here - other queries that accept this one as a sub-query will often look for that attribute.

flowmachine/flowmachine/features/location/majority_location.py
```diff
@@ -0,0 +1,35 @@
+from typing import List
+
+from flowmachine.core import Query
+from flowmachine.core.mixins import GeoDataMixin
+from flowmachine.features.utilities.subscriber_locations import BaseLocation
+
+
+class MajorityLocation(BaseLocation, Query):
+    def __init__(self, subscriber_subset: Query, weight_column):
```
jc-harrison, line 9: I'd suggest something like 'subscriber_location_weights' instead of 'subscriber_subset'

jc-harrison:
I'm wondering whether it's worth adding an optional `include_unlocatable` argument - with `include_unlocatable=True`, the query would return a row for every subscriber in `subscriber_location_weights`, with nulls in the location columns when a subscriber has no majority location. Could be achieved by right-joining to the distinct subscribers in `subscriber_location_weights`, for example.

My thinking is that this might be useful if we want to include "unlocatable -> location A" in OD matrices, for example.

flowmachine/flowmachine/features/location/majority_location.py
```diff
@@ -0,0 +1,56 @@
...
+
+class MajorityLocation(BaseLocation, Query):
+    def __init__(
+        self,
+        subscriber_location_weights: Query,
+        weight_column,
+        include_unlocatable=False,
+    ):
+        if weight_column not in subscriber_location_weights.column_names:
+            raise ValueError("weight_column must exist in subscriber_subset")
```
jc-harrison, line 16: Would be good to have a test for this.

flowmachine/flowmachine/features/location/majority_location.py
```diff
@@ -0,0 +1,56 @@
...
+        self.include_unlocatable = include_unlocatable
+        self.spatial_unit = subscriber_location_weights.spatial_unit
+        super().__init__()
+
+    @property
+    def column_names(self) -> List[str]:
+        # This should be a reduction, so no new columns
+        return [
+            "subscriber"
+        ] + self.subscriber_location_weights.spatial_unit.location_id_columns
```
jc-harrison, line 28: Could use `self.spatial_unit.location_id_columns` here (not that it really makes any difference)

flowmachine/flowmachine/features/location/majority_location.py
```diff
@@ -0,0 +1,56 @@
...
+WITH subscriber_subset AS (
+    {self.subscriber_location_weights.get_query()}
+), summed_weights AS (
+    SELECT subscriber, sum({self.weight_column}) AS total_weight
+    FROM subscriber_subset
+    GROUP BY subscriber
+), seen_subs AS (
+    SELECT subscriber, {loc_id}
+    FROM summed_weights JOIN subscriber_subset USING(subscriber)
+    WHERE {self.weight_column} > total_weight/2.0
```
Thingus, line 44: A thought - I think that this would still work if `weight_column` is a float, but I'm not sure if there's an issue if it contains zeros.

flowmachine/tests/test_majority_location.py
```diff
@@ -0,0 +1,51 @@
...
+            daily_location("2016-01-01"),
+            daily_location("2016-01-02"),
+            daily_location("2016-01-03"),
+            daily_location("2016-01-04"),
+        )
+    )
+    yield lv
+
+
+def test_majority_location(get_dataframe, location_visits):
```
jc-harrison, line 24: Might be good to add a test that the `column_names` property matches the result columns

jc-harrison (commented):
Just a few minor comments here. Looking good - once the last few things are in place (docstring, type hints, changelog entry), I'm happy.

flowmachine/flowmachine/features/location/majority_location.py
jc-harrison: I think it should be fine. If `weight_column=0` then `weight_column <= total_weight/2.0` (provided there are no negative weights), so that row won't be included.

Could get some unexpected behaviour if there are negative weights (including getting multiple output rows per subscriber), but that would be a strange way to use this query. I'm not sure it's worth explicitly checking that all weights are >=0, though.

flowmachine/flowmachine/features/location/majority_location.py
```diff
@@ -0,0 +1,75 @@
...
+)
+"""
+
+        if self.include_unlocatable:
+            sql += f"""
+SELECT subscriber, seen_subs.{loc_id}
+FROM seen_subs RIGHT OUTER JOIN summed_weights USING(subscriber)
+            """
+        else:
+            sql += f"""SELECT subscriber, {loc_id} FROM seen_subs"""
```
greenape, line 73: Marginally nice to write this as "SELECT subscriber, seen_subs.{loc_id} FROM seen_subs" in the main body, then you can write:


```
if self.include_unlocatable:
  return f"{sql} RIGHT OUTER JOIN summed_weights USING(subscriber)"
return sql
```

?

flowmachine/flowmachine/features/location/majority_location.py
greenape: Can't check the weights are all >=0 ahead of time without executing the sql, right? You could coerce them to be so, or equally add a warning to the docstring?

jc-harrison: If there's a single obvious choice for how to coerce negative weights to be >=0 then I think I'd be in favour of that (along with a warning in the docstring). Setting negative weights equal to zero (effectively the same as just ignoring them) seems the most appropriate option, although using the absolute value is possibly an alternative.

flowmachine/flowmachine/features/location/majority_location.py
greenape: Ah, actually this'll go wrong anyway with a multicolumn spatial unit, no?

flowmachine/flowmachine/features/location/majority_location.py
Thingus: Easy route is to filter out using `WHERE weight >= 0`?

flowmachine/flowmachine/features/location/majority_location.py
Thingus: I think so, it'd shake out to `loc_id.[lat, lon]` (which I don't think is legal SQL?)

Thingus: I'm going to add a test using `exemplar_spatial_unit_param` to catch this

flowmachine/flowmachine/features/location/majority_location.py
Thingus: Sorry, read the discussion properly now. I do think that an explicit check is worth it, as we're already visiting each row as part of the summing-up - I don't think that this would add much to the execution time.

jc-harrison: I'm not sure I'm in favour of that. It's not so much the extra cost of the check, but the fact that the check wouldn't happen until you try to get the result. You could create the query, submit it, wait for it to get to the top of the queue, and only then find out it's not valid. I think filtering out negative weights is preferable, provided there's a warning in the docstring.

Thingus: Agree with that - filter in place, and docstring.

flowmachine/flowmachine/features/location/majority_location.py
```diff
@@ -0,0 +1,82 @@
...
+        self,
+        subscriber_location_weights: Query,
+        weight_column: str,
+        include_unlocatable: bool = False,
+    ):
+        if "subscriber" not in subscriber_location_weights.column_names:
+            raise ValueError("`subscriber` not in subscriber_location_weights query")
+        if weight_column not in subscriber_location_weights.column_names:
+            raise ValueError("weight_column must exist in subscriber_subset")
+        if not getattr(subscriber_location_weights, "spatial_unit"):
```
jc-harrison, line 42: More correct to use 'hasattr' here. I can't imagine we'd ever have a spatial unit that evaluates as falsy, but if we did it would mess up this check.

flowmachine/flowmachine/features/location/majority_location.py
```diff
@@ -0,0 +1,82 @@
...
+            raise AttributeError(
+                "subscriber_location_weights needs a spatial_unit column"
```
jc-harrison, line 48:
```suggestion
                "subscriber_location_weights query needs a spatial_unit attribute"
```

flowmachine/flowmachine/features/location/majority_location.py
```diff
@@ -0,0 +1,86 @@
...
+    FROM subscriber_subset
+    WHERE {self.weight_column} >= 0
+    GROUP BY subscriber
+), seen_subs AS (
+    SELECT subscriber, {loc_id}
+    FROM summed_weights JOIN subscriber_subset USING(subscriber)
+    WHERE {self.weight_column} > total_weight/2.0
+)
+"""
+
```
jc-harrison, line 77: One consequence of ignoring negative weights is that if a subscriber appears in `subscriber_location_weights` with _only_ negative weights, they won't appear in `summed_weights` and therefore won't appear in the query output even if `include_unlocatable=True`. But given that's a weird corner case where the expected behaviour is unclear anyway, I'm not too concerned about this.

flowmachine/flowmachine/features/location/majority_location.py
```diff
@@ -0,0 +1,86 @@
...
+        The query object containing subscribers, locations, and weights.
+    weight_column: str
+        The column, when summed, that will produce the count used to threshold the majority
+    include_unlocatable: bool default False
+        If `True`, returns every unique subscriber in the `subscriber_location_weights` query, with
+        the location column as `NULL` if no majority is reached.
+        If `False`, returns only subscribers that have achieved a majority location
+
+    Notes
+    -----
```
jc-harrison, line 32: Probably good to make the args keyword-only

flowmachine/flowmachine/features/location/majority_location.py
Thingus: Lets you filter out negative subs if they only have invalid weights, but not if they have a mix of valid and invalid weights - not a bug, a feature :wink:

flowmachine/tests/test_majority_location.py
```diff
@@ -0,0 +1,82 @@
...
+            ["37J9rKydzJ0mvo0z", "524 4 12 62"],
+            ["3XKdxqvyNxO2vLD1", "524 4 12 62"],
+            ["81B6q0K8k325XWmn", "524 1 03 13"],
+            ["8dXPM6JXj05qwjW0", "524 4 12 62"],
+            ["bKZLwjrMQG7z468y", "524 4 12 62"],
+            ["g9D5KEK9BQzOa8z0", "524 4 11 57"],
+        ],
+        columns=["subscriber", "pcod"],
+    )
+    assert (ml.column_names == target.columns).all()
```
jc-harrison, line 47: Again, being extremely pedantic here - I think I'd compare `ml.column_names` to `out.columns`, rather than `target.columns`. That way, we're asserting that `column_names` is self-consistent with the returned result, and _then_ checking that the returned result is what we expect it to be, which is a better separation of the two issues.

flowmachine/tests/test_majority_location.py
```diff
@@ -0,0 +1,82 @@
...
+        weight_column="dl_count",
+        include_unlocatable=True,
+    )
+    out_ml = get_dataframe(ml)
+    out_lv = get_dataframe(lv)
+    assert out_lv.subscriber.nunique() == len(out_ml)
+    assert len(out_ml.pcod.dropna()) == 15
+
+
+def test_weight_column_missing(location_visits):
```
jc-harrison, line 64: Equivalent tests for the 'subscriber' and 'spatial_unit' errors should get the coverage up, to keep automerge happy.

jc-harrison (approved):
Last few minor comments, then we should be good to merge.

PR #4525, opened 2021-11-12
Combine first

### LINKED ISSUE #4524: CombineFirst class/method
We would like to be able to combine the outputs of two subscriber location queries, such that a subscriber's location is assigned from the first query if they appear in that query result, or the second query otherwise (i.e. `SELECT subscriber, COALESCE(location_from_first_query, location_from_second_query)`).

I'd suggest a useful way to implement this would be to add a new query class that takes two queries, joins on specified `join_columns`, and coalesces values in specified `combine_columns`. This is somewhat similar to the `pandas.DataFrame.combine_first` method, and I propose to name it accordingly.

We could also add a `Query.combine_first` method to wrap this class (along the lines of the existing `join`, `subset`, `union` methods).

jc-harrison:
Closes #4524 

### I have:

### Description

Adds a `CombineFirst` query class, and also `CoalescedLocation` (a wrapper around `CombineFirst` for the specific case of joining on subscriber and coalescing location, adding `BaseLocation` inheritance and `spatial_unit` attribute).


flowmachine/flowmachine/core/combine_first.py
```diff
@@ -0,0 +1,98 @@
...
+from typing import Collection, List, Union
+
+from flowmachine.core.query import Query
+from flowmachine.core.errors import MissingColumnsError
+
+
+class CombineFirst(Query):
+    """
+    Given two queries 'first_query' and 'other_query', fill null or missing values in the
+    result of 'first_query' using those in the result of 'other_query'.
```
Thingus: Is there a definintion of null that doesn't encompass 'not present in `first_query`, like a nodata value or similar?

jc-harrison: PostgreSQL 'NULL' values in first_query will be replaced by non-null values from other_query. Is that what you're asking?

Thingus: Yep, that was what I wanted to know.

flowmachine/flowmachine/features/utilities/combine_first.py
```diff
@@ -0,0 +1,128 @@
...
+    (i.e. nulls in column 'a' of first_query are filled with values from
+    column 'a' of other_query)
+    """
+
+    def __init__(
+        self,
+        *,
+        first_query: Query,
+        other_query: Query,
+        join_columns: Union[str, Collection[str]],
```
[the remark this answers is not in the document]
jc-harrison: Main reason is so we can simplify this query by enforcing that combine column names are the same on both sides (i.e. nulls in 'col1' of first_query are filled with values from 'col1' of other_query). A `Join` would have to use `left_append` and `right_append` to give the columns different names, and then the `combine_columns` parameter here would have to be a collection of pairs of column names. Would make this query more general, but more complicated to use in the case when column names are the same on both sides.

flowmachine/flowmachine/features/utilities/combine_first.py
```diff
@@ -0,0 +1,137 @@
...
+        FROM ({self.first_query.get_query()}) l
+        FULL OUTER JOIN ({self.other_query.get_query()}) r
+        USING ({join_columns_string})
+        """
+
+        return sql
+
+
+class CoalescedLocation(BaseLocation, CombineFirst):
+    """
```
Thingus, line 105: Do we want this in it's own file under `subscribers` or `locations`? Think it'd make the import make more sense.

flowmachine/flowmachine/features/utilities/combine_first.py
```diff
@@ -0,0 +1,137 @@
...
+               {coalesce_string}
+        FROM ({self.first_query.get_query()}) l
+        FULL OUTER JOIN ({self.other_query.get_query()}) r
+        USING ({join_columns_string})
+        """
+
+        return sql
+
+
+class CoalescedLocation(BaseLocation, CombineFirst):
```
greenape, line 104: Split to a separate file I think.

Thingus (approved):
Approving, with the small consideration of file location for `CoalescedLocation`

flowmachine/flowmachine/features/utilities/combine_first.py
jc-harrison: Yes, that's a good point. I'll move it into 'subscribers'.

CHANGELOG.md
```diff
@@ -12,7 +12,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Made hour of day slicing available via FlowAPI. [#3165](https://github.com/Flowminder/FlowKit/issues/3165)
 - Added visited on most days reference location query. [#4267](https://github.com/Flowminder/FlowKit/issues/4267)
 - Added unique value from query list query. [#4486](https://github.com/Flowminder/FlowKit/pull/4486)
-- Added mixing for exposing start_date and end_date internally as datetime objects [#4497](https://github.com/Flowminder/FlowKit/issues/4497) 
+- Added mixing for exposing start_date and end_date internally as datetime objects [#4497](https://github.com/Flowminder/FlowKit/issues/4497)
```
greenape, line 15:
```suggestion
- Added mixin for exposing start_date and end_date internally as datetime objects [#4497](https://github.com/Flowminder/FlowKit/issues/4497)
```

Might as well correct that in passing.

greenape (changes requested):
All good for me but I'd prefer not to have the two classes in one file.

greenape (approved):

PR #4593, opened 2021-11-23
Union many

### LINKED ISSUE #4565: Allow Union to take a list of queries
Currently, `Union` produces a union of two queries. Taking a union of more than two queries can be achieved by recursively applying Union (i.e. `a.union(b).union(c).union(d)`), but this results in a deeply-nested structure (both within the SQL query and in terms of the python `Query` objects), which may cause performance issues and could also result in excessive duplication of results in cache (e.g.`a.union(b).union(c).union(d).store(store_dependencies=True)` would produce cache tables for `a`, `b`, `aUb`, `c`, `aUbUc`, `d` and `aUbUcUd`.

It would be useful if `Union` could take an arbitrary number of queries as arguments (e.g. `Union(a, b, c, d, all=True)` would union all four queries in one shot). This could be achieved either through arbitrary `*args` (which would remain compatible with the current implementation provided `top` and `bottom` are always provided as positional args and `all` is always explicitly provided as a kwarg) or a single list argument.

We could make `Union` smarter, so that Union(Union(a, b), c) would flatten the inputs into a single non-nested sequence of unions. However, this would mean pre-cached unions couldn't be used as components of larger unions, which may be something we need sometimes (e.g. if a union is already computed but we later want to extend it with rows from another query, or if we have a very large number of sub-queries and the only feasible way to union them all is to do so in stages).

greenape:
Closes #4565
### I have:

### Description

Changes `Union` and `Query.union` such that they accept variable numbers of `Query` objects to union together.

### DIFF
--- a/flowmachine/flowmachine/core/union.py
+++ b/flowmachine/flowmachine/core/union.py
@@ -14,25 +14,26 @@ class Union(Query):
 
     Parameters
     ----------
-    top, bottom : flowmachine.Query object
-        First and second tables respectively
+    queries : flowmachine.Query object
+        Queries to union
+    all : bool, default: True
+        If True, queries will be deduped, if set to False, duplicate rows will be preserved.
     """
 
-    def __init__(self, top, bottom, all=True):
-        """"""
-
-        self.top = top
-        self.bottom = bottom
+    def __init__(self, *queries: Query, all: bool = True):
+        self.queries = queries
         self.all = all
+        column_set = set(tuple(query.column_names) for query in queries)
+        if len(column_set) > 1:
+            raise ValueError("Inconsistent columns.")
 
         super().__init__()
 
     @property
     def column_names(self) -> List[str]:
-        return list(self.top.column_names)
+        return list(self.queries[0].column_names)
 
     def _make_query(self):
-
-        return """({}) UNION {} ({})""".format(
-            self.top.get_query(), "ALL" if self.all else "", self.bottom.get_query()
+        return (" UNION ALL " if self.all else " UNION ").join(
+            f"({query.get_query()})" for query in self.queries
         )
--- a/flowmachine/flowmachine/core/union.py
+++ b/flowmachine/flowmachine/core/union.py
@@ -26,6 +26,8 @@ class Union(Query):
         column_set = set(tuple(query.column_names) for query in queries)
         if len(column_set) > 1:
             raise ValueError("Inconsistent columns.")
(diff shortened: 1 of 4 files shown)


flowmachine/flowmachine/core/union.py
```diff
@@ -14,25 +14,26 @@ class Union(Query):
...
+        If True, queries will be deduped, if set to False, duplicate rows will be preserved.
     """
 
-    def __init__(self, top, bottom, all=True):
-        """"""
-
-        self.top = top
-        self.bottom = bottom
+    def __init__(self, *queries: Query, all: bool = True):
+        self.queries = queries
```
jc-harrison, line 24: Is it worth raising an error here if only one query is provided? The query would still work in that situation, but would be an entirely pointless wrapper around the input query.

jc-harrison: In fact, it wouldn't necessarily "work" for a single query, in that with `all=False` this should dedupe, which won't be the case for a single query.

jc-harrison (commented):
Would be good to have a test that uses more than 2 queries

flowmachine/flowmachine/core/union.py
```diff
@@ -14,25 +14,26 @@ class Union(Query):
...
+    all : bool, default: True
+        If True, queries will be deduped, if set to False, duplicate rows will be preserved.
     """
 
-    def __init__(self, top, bottom, all=True):
-        """"""
-
-        self.top = top
-        self.bottom = bottom
+    def __init__(self, *queries: Query, all: bool = True):
```
jc-harrison, line 23: Is it worth sorting the list of queries (either sorting in `__init__`, or doing `self.queries = set(queries)` and sorting in `_make_query`)? That way, the query (and query ID) is the same regardless of the input order.

Thingus: That could mess up any queries that depend on an already-derived ordering prior to joining (which shouldn't happen, but could)

flowmachine/flowmachine/core/union.py
greenape: Worth an error on one query, yep.

flowmachine/flowmachine/core/union.py
greenape: I think not because the order is meaningful when deduping (e.g. I've ordered home locations by time).

jc-harrison (approved):
:+1:

Issue #4585, opened 2021-11-22
Expose `CoalescedLocation`


jc-harrison:
Once #4584 is addressed, we'd like to be able to use FlowAPI to run a query which uses a query of this new kind as the "fallback location" in a `CoalescedLocation` query. This requires us to Expose `CoalescedLocation` and the new query class from #4584 through the API - either as two separate query schemas (with the new query kind an allowed input to 'coalesced_location') or combined into a single exposed query. Add the new query schema(s) to `ReferenceLocationSchema`, and add corresponding flowclient query specs.

jc-harrison:
I suggest that for now we roll the new query from #4584 into the exposed `CoalescedLocation` endpoint, so the parameters would be:
- preferred_location (e.g. a `MajorityLocation` query)
- fallback_location (e.g. a different `MajorityLocation` query)
- subscriber_location_weights (e.g. a `LocationVisits` query)
- weight_threshold (a number)
The query would then join fallback_location to subscriber_location_weights, threshold to weight < weight_threshold, and use that as the fallback location in a `CoalescedLocation` query.

PR #4616, opened 2021-11-25
Exposing Majority Location and LocationCount

### LINKED ISSUE #4582: Expose `LocationVisits`
Expose `LocationVisits` through the API, and add a corresponding flowclient query spec. I don't think this would be appropriate as a parameter of any currently-exposed queries (although I suppose it could be used in place of `UniqueLocations`, perhaps), but it should be allowed as an input to `MajorityLocation` and `PerSubscriberAggregate` once those are exposed.

I'd suggest we should directly accept a list of daily/modal location queries for the 'locations' parameter, rather than separately exposing `DayTrajectories` and allowing that as an input here.

I also think we should rename the 'dl_count' column name of `LocationVisits` to 'value', for consistency with other metrics exposed through the API.

### LINKED ISSUE #4583: Expose `MajorityLocation`
Expose `MajorityLocation` through the API, and add a corresponding flowclient query spec. This should be added to `ReferenceLocationSchema` and `InputToFlowsSchema` (while we're here, I'm wondering whether there's any reason not to add `ReferenceLocationSchema` to `InputToFlowsSchema`, to avoid duplication here).

The 'subscriber_location_weights' parameter should accept any appropriate exposed query (i.e. those with subscriber, location and numeric 'value' columns) - for the moment, I think this will only be `LocationVisits`.

I think 'weight_column' should be hard-coded to 'value', which will require renaming the 'dl_count' column of `LocationVisits`.

Thingus:
Closes #4582
Closes #4583

### I have:

### Description

Exposes MajorityLocation - also deduplicates InputToFlowsSchema and ReferenceLocationSchema

(NOTE: There was so much overlap between this and #4617 that it wasn't worth keeping them separate anymore; I'll cancel that issue)


Thingus:
Blocked until #4582 closed

Thingus:
Merged in-progress from fix for #4582

Thingus:
Though it's still a draft, it just needs the flowclient method - would welcome comments on the schema implementation

flowmachine/flowmachine/core/server/query_schemas/flows.py
```diff
@@ -18,18 +18,14 @@
 __all__ = ["FlowsSchema", "FlowsExposed"]
 
 from .most_frequent_location import MostFrequentLocationSchema
+from .reference_location import ReferenceLocationSchema
 
 from .unique_locations import UniqueLocationsSchema
 
 
 class InputToFlowsSchema(OneOfSchema):
```
greenape, line 26: Could just remove this entirely and use `ReferenceLocationSchema` down on L60?

flowmachine/flowmachine/core/server/query_schemas/majority_location.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+class LocatableQueries(OneOfSchema):
+    type_field = "query_kind"
+    type_schemas = {"location_visits": LocationVisitsSchema}
+
+
+class MajorityLocationSchema(BaseSchema):
+    query_kind = fields.String(validate=OneOf(["majority_location"]))
+    subscriber_location_weights = fields.Nested(LocatableQueries)
```
jc-harrison, line 37:
```suggestion
    subscriber_location_weights = fields.Nested(LocatableQueries, required=True)
```

flowmachine/flowmachine/core/server/query_schemas/majority_location.py
```diff
@@ -0,0 +1,40 @@
...
+            include_unlocatable=self.include_unlocatable,
+        )
+
+
+class LocatableQueries(OneOfSchema):
+    type_field = "query_kind"
+    type_schemas = {"location_visits": LocationVisitsSchema}
+
+
+class MajorityLocationSchema(BaseSchema):
```
jc-harrison, line 35: As with `LocationVisitsSchema`, good to use `BaseQueryWithSamplingSchema` here.

For context, we tend to enable sampling for all subscriber-level queries. Not for aggregates, though, because it doesn't really make sense to randomly sample the output of an aggregate query (and, in any case, the user could do so outside of FlowKit if they really wanted).

flowmachine/flowmachine/core/server/query_schemas/reference_location.py
```diff
@@ -16,9 +19,11 @@
 
 class ReferenceLocationSchema(OneOfSchema):
     type_field = "query_kind"
+    # Should UniqueLocation be in here?
```
jc-harrison, line 22: No, I don't think so. These queries all return a single location per subscriber, whereas `UniqueLocations` can return multiple locations per subscriber.

flowclient/flowclient/aggregates.py
```diff
@@ -1550,3 +1550,24 @@ def histogram_aggregate(*, connection: Connection, **kwargs) -> APIQuery:
         Histogram aggregate query
     """
     return connection.make_api_query(parameters=histogram_aggregate_spec(**kwargs))
+
+
+def majority_location_spec(*, subscriber_location_weights: Dict):
```
jc-harrison, line 1555: This needs an `include_unlocatable` argument

jc-harrison (changes requested):

flowmachine/flowmachine/core/server/query_schemas/location_visits.py
```diff
@@ -0,0 +1,39 @@
...
+    type_schemas = {
+        "daily_location": DailyLocationSchema,
+        "modal_location": ModalLocationSchema,
+    }
+
+
+class LocationVisitsSchema(BaseSchema):
+    query_kind = fields.String(validate=OneOf(["location_visits"]))
+    day_trajectories = fields.List(fields.Nested(VisitableLocation))
+    __model__ = LocationVisitsExposed
```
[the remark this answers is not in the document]
Thingus: I've open a new issue for this; keeping it here for future reference #4639

flowmachine/flowmachine/core/server/query_schemas/location_visits.py
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+
+class LocationVisitsSchema(BaseQueryWithSamplingSchema):
+    query_kind = fields.String(validate=OneOf(["location_visits"]))
+    locations = fields.List(fields.Nested(VisitableLocation), validate=Length(min=1))
+
+    @validates("locations")
+    def validate_locations(self, values):
+        if len(set(value.aggregation_unit for value in values)) > 1:
```
jc-harrison, line 44: This won't work if any of the locations are modal locations, because they won't have an `aggregation_unit` attribute. I think this could be fixed by doing `self.aggregation_unit = locations[0].aggregation_unit` in `ModalLocationExposed` - that should make the `aggregation_unit` attribute available here, without requiring the user to provide it as a parameter to `ModalLocationSchema`.

While we're here, it would be good to add an equivalent `@validates` method in `ModalLocationSchema` as well. It would also be good to define `aggregation_unit` attributes on `LocationVisitsExposed` and `MajorityLocationExposed` too - I think we'll need them when we get to #4585

jc-harrison: Also good to add a test that includes a modal location in `locations`, to catch this.

flowmachine/tests/test_query_object_construction.py
```diff
@@ -278,13 +279,60 @@ def test_construct_query(diff_reporter):
                 "method": "last",
             },
             "to_location": {
-                "query_kind": "unique_locations",
+                "query_kind": "most_frequent_location",
```
jc-harrison, line 282: Why has this changed? Given that "unique_locations" is an extra option for Flows (in addition to those in ReferenceLocationSchema), i think it's useful to have "unique_locations" used in one of the tests.

flowmachine/flowmachine/core/server/query_schemas/location_visits.py
```diff
@@ -0,0 +1,53 @@
...
+            raise ValidationError("All locations must have the same spatial unit")
```
jc-harrison, line 51:
```suggestion
            raise ValidationError("All locations must have the same aggregation unit")
```

flowmachine/tests/test_query_object_construction.py
```diff
@@ -355,3 +403,30 @@ def test_invalid_sampling_params_raises_error(sampling, message):
     with pytest.raises(ValidationError, match=message) as exc:
         _ = FlowmachineQuerySchema().load(query_spec)
     print(exc)
+
+
+@pytest.mark.parametrize("query_type", ["daily_location", "modal_location"])
+def test_unmatching_spatial_unit_raiese_error(query_type):
+    query_spec = {
+        "query_kind": query_type,
```
jc-harrison, line 411: 'query_kind' here should always be "location_visits" - I'm guessing you intended to parametrize the 'locations'?

jc-harrison (changes requested):
Just one test to sort out, then I think this should be ready.

flowmachine/tests/test_query_object_construction.py
```diff
@@ -355,3 +403,98 @@ def test_invalid_sampling_params_raises_error(sampling, message):
     with pytest.raises(ValidationError, match=message) as exc:
         _ = FlowmachineQuerySchema().load(query_spec)
     print(exc)
+
+
+@pytest.fixture
+def location_list_unmatched_agg():
```
jc-harrison, line 409: Looks like this new fixture isn't actually used?

Thingus: Oh yeah - turned out not to be useful.

flowmachine/tests/test_query_object_construction.py
```diff
@@ -355,3 +403,78 @@ def test_invalid_sampling_params_raises_error(sampling, message):
     with pytest.raises(ValidationError, match=message) as exc:
         _ = FlowmachineQuerySchema().load(query_spec)
     print(exc)
+
+
+def test_unmatching_spatial_unit_raises_error_daily(location_list_unmatched_agg):
```
jc-harrison, line 408: These tests still reference the now-removed fixture (although they don't use it)

jc-harrison (approved):
A couple of references to the fixture that still need to be removed, but otherwise this looks good.

PR #4617, opened 2021-11-25
Expose LocationVisits query

### LINKED ISSUE #4582: Expose `LocationVisits`

Thingus:
Closes #4582

### I have:

### Description
Exposes LocationVisits query and adds a FlowClient method

### DIFF
--- a/flowmachine/flowmachine/core/server/query_schemas/location_visits.py
+++ b/flowmachine/flowmachine/core/server/query_schemas/location_visits.py
@@ -1,0 +1,23 @@
+from marshmallow import fields
+from marshmallow.validate import OneOf
+
+from flowmachine.core.server.query_schemas import BaseExposedQuery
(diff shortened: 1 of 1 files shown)


Thingus:
NOTE: There was so much overlap between this and #4616 that it wasn't worth keeping them separate anymore; further work is carrying on there

PR #4626, opened 2021-11-26
Visit count filtered reference location

### LINKED ISSUE #4584: Add a new flowmachine query class that can handle "`MajorityLocation` joined to `LocationVisits`"
We would like to be able to define a query that returns a specified reference location (we are particularly interested in `MajorityLocation` for each subscriber, only if they visited that location on more than 'threshold' days. This could be achieved with something like:
```
SELECT subscriber, location
FROM {reference_locations}
INNER JOIN {subscriber_locations_subset}
USING (subscriber, location)
```
where the parameters could be `reference_locations=MajorityLocation(...)` and `subscriber_locations_subset=LocationVisits(...).numeric_subset(low=threshold, high=max_days)`. Alternatively we could use `INTERSECT` or `INTERSECT ALL` instead of an inner join - either would treat duplicates differently from the join, and the choice would presumably have performance implications.

This whole step could be achieved using existing methods, using:
```
reference_locations_query.join(
    location_visits_query.numeric_subset(low=threshold, high=max_days)),
    how="inner",
    left_on=('subscriber', <location_id_columns>),
    right_on=('subscriber', <location_id_columns>),
```
but this is not ideal because:
- we'd like to use this query as an input to `CoalescedLocation` (and quite possibly other queries), which would expect to recognise it as a subscriber location query (i.e. having a `spatial_unit` attribute), which is not the case if the query is a generic `Join` query,
- exposing generic `Join` through the API wouldn't help, for much the same reason - only specific joins would be suitable parameters for the queries that could consume this one, so we'd need a specific marshmallow schema for those special-case joins, which would mirror the flowmachine query we're proposing to add here.

greenape:
Closes #4584

### I have:

### Description

Implements #4584 essentially as suggested in the issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Added `join_type` param to `Flows` class. [#4539](https://github.com/Flowminder/FlowKit/issues/4539)
 - Added `PerSubscriberAggregate` query. [#4559](https://github.com/Flowminder/FlowKit/issues/4559)
 - Added FlowETL QA checks 'count_imeis', 'count_imsis', 'count_locatable_location_ids', 'count_null_imeis', 'count_null_imsis', 'count_null_location_ids', 'max_msisdns_per_imei', 'max_msisdns_per_imsi', 'count_added_rows_outgoing', 'count_null_counterparts', 'count_null_durations', 'count_onnet_msisdns_incoming', 'count_onnet_msisdns_outgoing', 'count_onnet_msisdns', 'max_duration' and 'median_duration'. [#4552](https://github.com/Flowminder/FlowKit/issues/4552)
-- Added `VisitCountFilteredReferenceLocation` query, which returns only rows where a subscriber visited a reference location the required number of times. [#4584](https://github.com/Flowminder/FlowKit/issues/4584) 
+- Added `FilteredReferenceLocation` query, which returns only rows where a subscriber visited a reference location the required number of times. [#4584](https://github.com/Flowminder/FlowKit/issues/4584) 
 
 ### Changed
 - Harmonised FlowAPI parameter names for start and end dates. They are now all `start_date` and `end_date`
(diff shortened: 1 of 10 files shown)


flowmachine/flowmachine/features/subscriber/visit_count_filtered_reference_location.py
```diff
@@ -0,0 +1,61 @@
...
+    lower_bound : int
+        Minimum number of visits to allow a location to be included
+    upper_bound : int, default Inf
+        Maximum number of visits to allow a location to be included. Defaults to unbounded.
+    """
+
+    def __init__(
+        self,
+        reference_locations_query: BaseLocation,
+        location_visits_query: LocationVisits,
```
jc-harrison, line 34: Does this need to be a `LocationVisits` query? E.g. I think this would work equally well with `CallDays` - or any query with (subscriber, location, value) columns.

flowmachine/flowmachine/features/subscriber/visit_count_filtered_reference_location.py
```diff
@@ -0,0 +1,61 @@
...
+                "subscriber",
+                *reference_locations_query.spatial_unit.location_id_columns,
+            ),
+            on_right=(
+                "subscriber",
+                *reference_locations_query.spatial_unit.location_id_columns,
+            ),
+            right_append="_visit_count",
+        )
+        self.spatial_unit = reference_locations_query.spatial_unit
```
jc-harrison, line 53: Probably worth raising an error here if `reference_locations_query.spatial_unit` and `location_visits_query.spatial_unit` don't match

CHANGELOG.md
```diff
@@ -18,6 +18,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Added `join_type` param to `Flows` class. [#4539](https://github.com/Flowminder/FlowKit/issues/4539)
 - Added `PerSubscriberAggregate` query. [#4559](https://github.com/Flowminder/FlowKit/issues/4559)
 - Added FlowETL QA checks 'count_imeis', 'count_imsis', 'count_locatable_location_ids', 'count_null_imeis', 'count_null_imsis', 'count_null_location_ids', 'max_msisdns_per_imei', 'max_msisdns_per_imsi', 'count_added_rows_outgoing', 'count_null_counterparts', 'count_null_durations', 'count_onnet_msisdns_incoming', 'count_onnet_msisdns_outgoing', 'count_onnet_msisdns', 'max_duration' and 'median_duration'. [#4552](https://github.com/Flowminder/FlowKit/issues/4552)
+- Added `VisitCountFilteredReferenceLocation` query, which returns only rows where a subscriber visited a reference location the required number of times. [#4584](https://github.com/Flowminder/FlowKit/issues/4584)
```
jc-harrison, line 21:
```suggestion
- Added `FilteredReferenceLocation` query, which returns only rows where a subscriber visited a reference location the required number of times. [#4584](https://github.com/Flowminder/FlowKit/issues/4584) 
```

jc-harrison (approved):
:+1:

Issue #4615, opened 2021-11-25
Disaggregate subscriber counts by mobility type


jc-harrison:
We would like to be able to disaggregate spatially-aggregated counts of home locations according to categories defined from their home locations over multiple months, in the following way:
![Screenshot 2021-11-25 at 13 37 51](https://user-images.githubusercontent.com/8958797/143451619-85b837fd-4fd4-4bf8-a48a-af11505603dd.png)

While we could do this using the existing subscriber subsetting mechanism, this would require calculating home locations once for all subscribers (as input for the categorisation) and then again separately for each category (since subscriber subsets are always applied to the events tables before further computation, so pre-computed intermediate results for a superset of the requested subset cannot be re-used - see #3808).

I would suggest the following:
- [x] #4666
- [x] #4668
- [x] #4679
- [x] #4671
- [x] #4669

PR #4672, opened 2021-12-06
Mobility classification

### LINKED ISSUE #4666: Implement MobilityClassification
Define a new flowmachine Query class, `MobilityClassification`, which assigns labels to subscribers according to the definition in #4615. For now this can be a single, specialised query - it may be appropriate to generalise the query or split into multiple sub-queries at a later stage.

jc-harrison:
Closes #4666

### I have:

### Description

Adds a new flowmachine query class -  `MobilityClassification` - that assigns "mobility type" labels to subscribers based on a sequence of locations.

I've changed the label names slightly from those in the diagram in #4615, but the categories themselves are unchanged (there's no explicit label for "inactive" - that's implied by absence of that subscriber in the query result). Happy to accept alternative suggestions for label names - the criteria I'm aiming to meet are that the label names should be:
- as short as possible
- unambiguous (i.e. it's clear from the label name which category it refers to)
- different from all other labels in the first character (this one's not essential, but should help make grouping by label as efficient as possible, as well as making labels more easily distinguishable at a glance)


flowmachine/flowmachine/features/subscriber/mobility_classification.py
```diff
@@ -0,0 +1,124 @@
...
+        SELECT
+            subscriber,
+            count(*) < {len(self.locations)} AS sometimes_inactive,
+            {count_unlocatable} < {len(self.locations)} AS sometimes_unlocatable
+        FROM ({locations_union}) AS locations_union
+        GROUP BY subscriber
+        """
+
+        # Find stay lengths using gaps-and-islands approach
+        long_term_mobility = f"""
```
greenape, line 88: I think I would consider breaking this in particular out into a separate query class.

flowmachine/flowmachine/features/subscriber/mobility_classification.py
```diff
@@ -0,0 +1,124 @@
...
+    @property
+    def column_names(self) -> List[str]:
+        return ["subscriber", "value"]
+
+    def _make_query(self) -> str:
+        loc_cols_string = ", ".join(self.locations[0].spatial_unit.location_id_columns)
+
+        # Workaround to count rows where all location columns are null, if there are multiple location columns
+        if len(self.locations[0].spatial_unit.location_id_columns) > 1:
+            count_unlocatable = f"sum((NOT (({loc_cols_string}) IS NULL))::int)"
```
greenape, line 65: This is actually `count_locatable` rather than _un_ locatable, right?

greenape: Might be more readable to use `num_nulls()`?

greenape: Cos then you can do:

`sum(num_nulls({loc_cols_string}))` for both cases, and check for above zero below.

jc-harrison: Yes, you're right, this is actually `count_locatable`.

I'm not sure `num_nulls` helps to make this cleaner - I think we want to count the rows where _all_ of the location columns are null, which would mean counting the rows with `num_nulls({loc_cols_string}) = <number_of_location_cols>`. (I don't think there's any situation where the location columns in one row might have a mix of null and non-null values, but if that ever did occur I expect we'd want to treat that as a known location rather than "unlocatable").

Perhaps flipping this round to `count_unlocatable = 0` instead of `count_locatable < {len(self.locations)}` is a better approach. We could then do `bool_or(({loc_cols_string}) IS NULL) AS sometimes_unlocatable`.

greenape: "location columns in one row might have a mix of null and non-null values" cell with longitude or latitude missing?

jc-harrison: Hmm, I hadn't considered that. I was thinking of 'unlocatable' in the sense used in MajorityLocation - i.e. "we couldn't assign this subscriber to a single location". Whereas a cell with missing longitude/latitude would be "we were able to assign this subscriber to a single cell, but we don't know where that cell _is_".

I think if we're using this query with versioned-cell-level locations, the first definition makes more sense - the subscriber is still locatable to a specific versioned cell. If we wanted to exclude cells with unknown locations, we'd be better off excluding those earlier so that they don't appear in the outputs of the 'locations' sub-queries.

flowmachine/flowmachine/features/subscriber/mobility_classification.py
```diff
@@ -0,0 +1,103 @@
...
+    subscribers absent from the query result can be assumed to fall into a
+    sixth category: "not active in the most recent period").
+
+    Parameters
+    ----------
+    locations : list of BaseLocation
+        List of reference location queries, each returning a single location
+        per subscriber (or NULL location for subscribers that are active but
+        unlocatable). The list is assumed to be sorted into ascending
+        chronological order.
```
Thingus, line 43: Would it matter if it's not, and if not, is it worth adding a sort to the `subscriber_stay_lengths` query?

flowmachine/flowmachine/features/subscriber/mobility_classification.py
```diff
@@ -0,0 +1,103 @@
...
+
+    @property
+    def column_names(self) -> List[str]:
+        return ["subscriber", "value"]
+
+    def _make_query(self) -> str:
+        loc_cols_string = ", ".join(self.locations[0].spatial_unit.location_id_columns)
+
+        # Note: in some ways it would be nicer to use a DayTrajectories query instead of a list of location queries,
+        # but DayTrajectories only works with queries that have 'start' and 'stop' attributes
```
Thingus, line 71: I guess, given the custom name of the column, using a Join object is out of the question here? (I guess it might improve caching, but I don't know how much impact it'd have)

flowmachine/flowmachine/features/subscriber/mobility_classification.py
greenape: Yeah, I can see your point. I too have difficulty conjuring a case where we'd be only partially assigning to be honest.

Thingus (approved):
Some minor comments, but only thoughts/suggestions.

flowmachine/flowmachine/features/subscriber/mobility_classification.py
jc-harrison: A `Union` object wouldn't work here because of the need to add the 'ordinal' column. We could use the same hack as in DayTrajectories, and wrap this up in a CustomQuery object, but I don't really like that hack in the place it's already used so I'm reluctant to use it again here.

There could be a benefit to caching this, since the same union is used in the SubscriberStayLengths query, but it's not guaranteed to help (for the same reason CTEs can lead to bad performance - if this bit of SQL is included in the full query then postgres might be able to do some optimisations that wouldn't be possible if it's pre-calculated and just gets read from disk). And if we do cache the union then we're doubling the disk usage, because the results of the location queries will all be cached separately and then cached again in the unioned query.

PR #4677, opened 2021-12-07
Implementing labelling disaggregation

### LINKED ISSUE #4615: Disaggregate subscriber counts by mobility type

### LINKED ISSUE #4668: Labelled spatial aggregates
Add equivalents of `SpatialAggregate` and `Flows`, which disaggregate counts according to per-subscriber categories defined in a specified sub-query (e.g. `MobilityClassification` - see #4666). The aggregate queries should take a `subscriber_labels` argument in addition to the subscriber-locations queries, join on subscriber, and group by label as well as location.

In the case of `Flows`, we could consider allowing two separate sets of labels (i.e. subscribers labelled differently at the "from" and "to" locations). It's not clear whether that would be useful in practice, though (and if it was, the same could be achieved by using a single `subscriber_labels` query that assigns a label-pair to each subscriber), so in the first instance a single set of labels should be sufficient.

This could be achieved using new classes `LabelledSpatialAggregate` and `LabelledFlows`, or by adding an optional `subscriber_labels` parameter to the existing `SpatialAggregate` and `Flows` classes.

Thingus:
Closes #4668

### I have:

### Description

Adds paramater to `spatial_aggregate` that takes a query containing subscribers + labels to disaggregate by

### DIFF
--- a/flowmachine/flowmachine/features/location/labelled_spatial_aggregate.py
+++ b/flowmachine/flowmachine/features/location/labelled_spatial_aggregate.py
@@ -1,3 +1,6 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
 from typing import List
 
 from flowmachine.core import Query
--- a/flowmachine/flowmachine/features/location/labelled_spatial_aggregate.py
+++ b/flowmachine/flowmachine/features/location/labelled_spatial_aggregate.py
@@ -16,6 +16,13 @@ class LabelledSpatialAggregate(GeoDataMixin, Query):
         subscriber_labels: Query,
         label_columns: List[str] = ("value",),
     ):
+        if any(
+            label_column not in subscriber_labels.column_names
+            for label_column in label_columns
+        ):
+            raise ValueError(
+                f"One of {label_columns} not a column of {subscriber_labels}"
+            )
         self.locations = locations
         self.subscriber_labels = subscriber_labels
         self.label_columns = list(label_columns)
--- a/flowmachine/flowmachine/features/location/labelled_spatial_aggregate.py
+++ b/flowmachine/flowmachine/features/location/labelled_spatial_aggregate.py
@@ -50,13 +50,9 @@ class LabelledSpatialAggregate(GeoDataMixin, Query):
         label_columns: List[str] = ("value",),
     ):
 
-        if any(
-            label_column not in subscriber_labels.column_names
-            for label_column in label_columns
-        ):
-            raise ValueError(
-                f"One of {label_columns} not a column of {subscriber_labels}"
-            )
+        for label_column in label_columns:
+            if label_column not in subscriber_labels.column_names:
+                raise ValueError(f"{label_column} not a column of {subscriber_labels}")
         if "subscriber" not in locations.column_names:
             raise ValueError(f"Locations query must have a subscriber column")
         if not hasattr(locations, "spatial_unit"):
--- a/flowmachine/flowmachine/features/location/labelled_spatial_aggregate.py
+++ b/flowmachine/flowmachine/features/location/labelled_spatial_aggregate.py
@@ -109,7 +109,7 @@ class LabelledSpatialAggregate(GeoDataMixin, Query):
                 {aggregate_cols}, {label_select}, count(*) AS value
             FROM 
                 ({self.locations.get_query()}) AS agg
-            JOIN
+            LEFT JOIN
                 ({self.subscriber_labels.get_query()}) AS labels USING (subscriber)
             GROUP BY
                 {aggregate_cols}, {label_group}
--- a/flowmachine/flowmachine/features/location/labelled_spatial_aggregate.py
+++ b/flowmachine/flowmachine/features/location/labelled_spatial_aggregate.py
@@ -7,7 +7,8 @@ from flowmachine.features.location.spatial_aggregate import SpatialAggregate
 
 class LabelledSpatialAggregate(GeoDataMixin, Query):
     """
-    Class represeneting a disaggregation of a SpatialAggregate by some label
+    Class representing a disaggregation of a SpatialAggregate by some label or set of labels
+    Returns a query with
     """
 
     def __init__(
@@ -19,25 +20,35 @@ class LabelledSpatialAggregate(GeoDataMixin, Query):
         self.locations = locations
         self.subscriber_labels = subscriber_labels
         self.label_columns = list(label_columns)
+        self.spatial_unit = locations.spatial_unit
+        self.label_columns = label_columns
+        self.out_label_columns = [
+            f"label_{label_col}" for label_col in self.label_columns
+        ]
         super().__init__()
 
     @property
     def column_names(self) -> List[str]:
         return (
-            self.locations.spatial_unit.location_id_columns
+            list(self.spatial_unit.location_id_columns)
             + [f"label_{label}" for label in self.label_columns]
             + ["value"]
         )
 
+    @property
+    def out_label_columns_as_string_list(self):
+        return ",".join(self.out_label_columns)
+
     def _make_query(self):
 
         aggregate_cols = ",".join(
-            f"agg.{agg_col}"
-            for agg_col in self.locations.spatial_unit.location_id_columns
+            f"agg.{agg_col}" for agg_col in self.spatial_unit.location_id_columns
         )
         label_select = ",".join(
-            f"labels.{label_col} AS label_{label_col}"
-            for label_col in self.label_columns
+            f"labels.{label_col} AS {out_label_col}"
+            for label_col, out_label_col in zip(
+                self.label_columns, self.out_label_columns
+            )
         )
         label_group = ",".join(
             f"labels.{label_col}" for label_col in self.label_columns
--- a/flowmachine/flowmachine/features/location/labelled_spatial_aggregate.py
+++ b/flowmachine/flowmachine/features/location/labelled_spatial_aggregate.py
@@ -21,7 +21,8 @@ class LabelledSpatialAggregate(GeoDataMixin, Query):
     label_columns: List[str]
         A list of columns in subscriber_labels to aggregate on
 
-    Example:
+    Examples
+    --------
 
     >>> locations = locate_subscribers(
     ...     "2016-01-01",
--- a/flowmachine/flowmachine/features/location/labelled_spatial_aggregate.py
+++ b/flowmachine/flowmachine/features/location/labelled_spatial_aggregate.py
@@ -8,6 +8,35 @@ from flowmachine.features.location.spatial_aggregate import SpatialAggregate
 class LabelledSpatialAggregate(GeoDataMixin, Query):
     """
     Class representing a disaggregation of a SpatialAggregate by some label or set of labels
+
+    Parameters
+    ----------
+    locations: Query
+        Any query with a subscriber and location columns
+    subscriber_labels: Query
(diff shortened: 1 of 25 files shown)


flowmachine/flowmachine/features/location/spatial_aggregate.py
```diff
@@ -18,31 +18,45 @@ class SpatialAggregate(GeoDataMixin, Query):
     Parameters
     ----------
     locations : subscriber location query
+    subscriber_label: A query with a  column to further disaggregate by
+    label_column: Name of the column containing the label: defaults to 'value'
```
greenape, line 22:
```suggestion
    label_column : str, default 'value'
        Name of the column containing the label: defaults to 'value'
```

(Might have the ident wrong there)

Thingus:
Does this need an ADR for the redaction strategy?

flowmachine/flowmachine/features/location/labelled_spatial_aggregate.py
```diff
@@ -0,0 +1,116 @@
...
+    Parameters
+    ----------
+    locations: Query
+        Any query with a subscriber and location columns
+    subscriber_labels: Query
+        Any query with a subscriber column
+    label_columns: List[str]
+        A list of columns in subscriber_labels to aggregate on
+
+    Example:
```
jc-harrison, line 24:
```suggestion
    Examples
    --------
```
(as per [numpydoc style](https://numpydoc.readthedocs.io/en/latest/format.html#examples))

flowmachine/flowmachine/features/location/labelled_spatial_aggregate.py
```diff
@@ -0,0 +1,117 @@
...
+        if any(
+            label_column not in subscriber_labels.column_names
+            for label_column in label_columns
+        ):
+            raise ValueError(
+                f"One of {label_columns} not a column of {subscriber_labels}"
+            )
+        if "subscriber" not in locations.column_names:
+            raise ValueError(f"Locations query must have a subscriber column")
+        if "spatial_unit" not in dir(locations):
```
jc-harrison, line 62:
```suggestion
        if not hasattr(locations, "spatial_unit"):
```

flowmachine/flowmachine/features/location/labelled_spatial_aggregate.py
```diff
@@ -0,0 +1,117 @@
...
+        subscriber_labels: Query,
+        label_columns: List[str] = ("value",),
+    ):
+
+        if any(
+            label_column not in subscriber_labels.column_names
+            for label_column in label_columns
+        ):
+            raise ValueError(
+                f"One of {label_columns} not a column of {subscriber_labels}"
```
jc-harrison, line 58: Could move this inside the loop and include the specific missing column name in the error message

flowmachine/flowmachine/features/location/labelled_spatial_aggregate.py
```diff
@@ -0,0 +1,117 @@
...
+        label_group = ",".join(
+            f"labels.{label_col}" for label_col in self.label_columns
+        )
+
+        sql = f"""
+            SELECT
+                {aggregate_cols}, {label_select}, count(*) AS value
+            FROM 
+                ({self.locations.get_query()}) AS agg
+            JOIN
```
jc-harrison, line 112: Might want to use a left join here, I think - which would lead to an additional group (label=NULL) for subscribers that appear in the location query but not the label query.

flowmachine/flowmachine/features/location/labelled_spatial_aggregate.py
```diff
@@ -0,0 +1,117 @@
...
+     1  524 3 08 44       Smart     28
+     2  524 4 12 62     Feature     44
+     3  524 4 12 62       Smart     19'
+
+    """
+
+    def __init__(
+        self,
+        locations: Query,
+        subscriber_labels: Query,
```
jc-harrison, line 49: I'd imagine that in practice, if we wanted to use multiple labels they would usually come from multiple queries. But I suppose there's always the option of `join`ing the queries first, to get a single query to use as the `subscriber_labels`, so I think this covers that use-case too.

flowmachine/flowmachine/features/location/labelled_spatial_aggregate.py
```diff
@@ -0,0 +1,117 @@
...
+        self.out_label_columns = [
+            f"label_{label_col}" for label_col in self.label_columns
+        ]
+        super().__init__()
+
+    @property
+    def column_names(self) -> List[str]:
+        return (
+            list(self.spatial_unit.location_id_columns)
+            + [f"label_{label}" for label in self.label_columns]
```
jc-harrison, line 81:
```suggestion
            + self.out_label_columns
```

flowmachine/flowmachine/features/location/redacted_labelled_spatial_aggregate.py
```diff
@@ -0,0 +1,62 @@
...
+        If any labels within a location reveal less than this number of subscribers, that location is dropeed
+    """
+
+    def __init__(
+        self,
+        *,
+        labelled_spatial_aggregate: LabelledSpatialAggregate,
+        redaction_threshold: int = 15,
+    ):
+
```
jc-harrison, line 32: You'll need `self.spatial_unit = labelled_spatial_aggregate.spatial_unit` here as well, for the get-result-as-geojson stuff to work.

flowmachine/flowmachine/features/location/redacted_labelled_spatial_aggregate.py
```diff
@@ -0,0 +1,62 @@
...
+        self.labelled_spatial_aggregate = labelled_spatial_aggregate
+        self.redaction_threshold = redaction_threshold
+        super().__init__()
+
+    @property
+    def column_names(self) -> List[str]:
+        return self.labelled_spatial_aggregate.column_names
+
+    def _make_query(self):
```
jc-harrison, line 41: We discussed having this return the total (non-disaggregated) count for a location if any of the disaggregated counts are below the threshold, rather than returning nothing at all for that location. But actually, that's no different to using this approach (drop all rows for locations with any below-threshold counts) alongside a second, non-disaggregated, SpatialAggregate query

flowmachine/flowmachine/features/location/labelled_spatial_aggregate.py
```diff
@@ -0,0 +1,117 @@
...
+    @property
+    def column_names(self) -> List[str]:
+        return (
+            list(self.spatial_unit.location_id_columns)
+            + self.out_label_columns
+            + ["value"]
+        )
+
+    @property
+    def out_label_columns_as_string_list(self):
```
jc-harrison, line 82: Is it worth having this as a property? Given that this is only used in `RedactedLabelledSpatialAggregate`, I'd be inclined to do the list-to-string bit there.

Thingus: I think it could be useful for building queries based on labels or propagating labels through queries for the future

jc-harrison: Yes, possibly. OK, that seems fair enough.

flowmachine/flowmachine/features/location/redacted_labelled_spatial_aggregate.py
jc-harrison: Yes, agreed. It would also require a second redaction step to drop all locations where the total count is below the redaction threshold, which I think would end up being a bit messy.

flowmachine/tests/test_labelled_spatial_aggregate.py
```diff
@@ -0,0 +1,89 @@
...
+from flowmachine.core.dummy_query import DummyQuery
+from flowmachine.features import TotalLocationEvents
+from flowmachine.core import make_spatial_unit
+from flowmachine.features import SubscriberHandsetCharacteristic
+from flowmachine.features.location.labelled_spatial_aggregate import (
+    LabelledSpatialAggregate,
+)
+from flowmachine.features.subscriber.daily_location import locate_subscribers
+
+
```
jc-harrison, line 12: I think a couple of other tests would be good:
- One test that this works with multiple `label_columns`
- And a test that re-aggregating gives the expected result (i.e. summing counts over all labels gives the same result as the equivalent un-labelled SpatialAggregate query)

jc-harrison:
> Does this need an ADR for the redaction strategy?

Yes, that's not a bad idea. We're likely to need to make similar decisions again, so recording the reasoning behind this choice would be good.

flowmachine/tests/test_labelled_spatial_aggregate.py
```diff
@@ -28,6 +28,41 @@ def test_labelled_spatial_aggregate(get_dataframe):
     assert len(df) == 50
     assert len(df.pcod.unique()) == 25
     assert len(df.label_value.unique()) == 2
+    # Total number of values should equal the initial number of subscribers
+    assert sum(map(int, df.value.tolist())) == len(get_dataframe(metric))
```
jc-harrison, line 32: I think it would make more sense to compare to `len(get_dataframe(locations))` here - in the event that the set of subscribers differed between `locations` and `metric`, we'd still expect each row in the `locations` query to be counted once.

Also, is there a reason not to use `df.value.sum()` for the left-hand side of this check?

jc-harrison: Could also consider testing that `df.groupby("pcod").value.sum()` is the same as the result of `SpatialAggregate(locations=locations)` (i.e. counts per location sum to the correct values).

flowmachine/tests/test_labelled_spatial_aggregate.py
```diff
@@ -28,6 +28,41 @@ def test_labelled_spatial_aggregate(get_dataframe):
...
+    labelled = LabelledSpatialAggregate(
+        locations=locations,
+        subscriber_labels=metric,
+        label_columns=["value_hnd_type", "value_model"],
+    )
+    df = get_dataframe(labelled)
+    foo = labelled.get_query()
+    assert all(
+        df.columns == ["pcod", "label_value_hnd_type", "label_value_model", "value"]
+    )
```
jc-harrison, line 64: Would be good to also check that the `column_names` property is accurate here (i.e. `df.columns == labelled.column_names`)

flowmachine/tests/test_labelled_spatial_aggregate.py
```diff
@@ -28,6 +28,41 @@ def test_labelled_spatial_aggregate(get_dataframe):
...
+        left_append="_hnd_type",
+        right_append="_model",
+    )
+    labelled = LabelledSpatialAggregate(
+        locations=locations,
+        subscriber_labels=metric,
+        label_columns=["value_hnd_type", "value_model"],
+    )
+    df = get_dataframe(labelled)
+    foo = labelled.get_query()
```
jc-harrison, line 61: I don't think this line is doing anything?

Thingus: Debug statement that got left in <_<

jc-harrison (approved):
:+1:

PR #4696, opened 2021-12-13
Labelled flows

### LINKED ISSUE #4679: Labelled flows
Issue for labelled flows (see linked)

Thingus:
Closes #4679 

### I have:

### Description

Introduces `LabelledFlows`, a disaggregation of the `Flows` class. Also modifies the `Flows` associated classes (`FlowsLike, BaseInOutFlow, InFlow, OutFlow`) to handle objects with a `label_columns` property.


flowmachine/flowmachine/features/location/flows.py
```diff
@@ -62,15 +62,56 @@ def _geo_augmented_query(self):
...
-        loc_cols_from_string = ",".join([f"{col}_from" for col in loc_cols])
-        loc_cols_to_string = ",".join([f"{col}_to" for col in loc_cols])
-        loc_cols_from_aliased_string = ",".join(
-            [f"{col}_from AS {col}" for col in loc_cols]
-        )
-        loc_cols_to_aliased_string = ",".join(
-            [f"{col}_to AS {col}" for col in loc_cols]
-        )
+
+        def _build_json_agg_clause(loc_cols, direction, label_cols=[]):
```
jc-harrison, line 71: Can this be split out as a separate method, instead of a function defined within `_geo_augmented_query`?

Thingus: Done + test added to keep covbot happy

flowmachine/flowmachine/features/location/flows.py
```diff
@@ -49,6 +49,48 @@ def inflow(self):
 
         return InFlow(self)
 
+    def _build_json_agg_clause(self, loc_cols, direction, label_cols=[]):
```
jc-harrison, line 52: Can probably get rid of the `loc_cols` and `label_cols` args now, and infer them from `self`

flowmachine/flowmachine/features/location/flows.py
```diff
@@ -193,12 +226,18 @@ def __init__(self, flow):
         cols = self.flow.column_names
         self.loc_from = ",".join([c for c in cols if c.endswith("_from")])
         self.loc_to = ",".join([c for c in cols if c.endswith("_to")])
+        self.label = ",".join(c for c in cols if c.endswith("_label"))
```
jc-harrison, line 231: Can this use `flow.out_label_columns` instead of filtering on column name? (And the same for `self.loc_from` and `self.loc_to` I suppose, using `flow.spatial_unit.location_id_columns`, although those haven't been added in this PR)

flowmachine/flowmachine/features/location/labelled_flows.py
```diff
@@ -0,0 +1,137 @@
...
+
+        if loc1.spatial_unit != loc2.spatial_unit:
+            raise ValueError("Flows must have the same spatial unit")
+        for label_column in label_columns:
+            if label_column not in labels.column_names:
+                raise ValueError(f"{label_column} not present in {labels.column_names}")
+
+        self.loc_from = loc1
+        self.loc_to = loc2
+        self.labels = labels
```
jc-harrison, line 90: `LabelledSpatialAggregate` has a `subscriber_labels` parameter, whereas this calls the parameter `labels`. I don't mind which we choose, but it would be good to stay consistent.

Thingus:
We'll need redaction in this branch as well - I'll need to implement that.

flowmachine/flowmachine/features/location/labelled_flows.py
```diff
@@ -0,0 +1,137 @@
...
+    def __init__(
+        self, *, loc1, loc2, labels, label_columns=("value",), join_type="inner"
+    ):
+
+        if loc1.spatial_unit != loc2.spatial_unit:
+            raise ValueError("Flows must have the same spatial unit")
+        for label_column in label_columns:
+            if label_column not in labels.column_names:
+                raise ValueError(f"{label_column} not present in {labels.column_names}")
+
```
jc-harrison, line 87: `LabelledSpatialAggregate` has a check that `"subscriber"` is not in `label_columns` - that would also be applicable here.

flowmachine/tests/test_labelled_flows.py
```diff
@@ -0,0 +1,174 @@
...
+    out = labelled_flows.to_geojson_string()
+    dict = loads(out)
+    assert all(
+        test in dict["features"][0]["properties"]["outflows"].keys()
+        for test in ["Smart", "Feature"]
+    )
+    assert all(
+        test in dict["features"][0]["properties"]["inflows"].keys()
+        for test in ["Smart", "Feature"]
+    )
```
jc-harrison, line 165: Would be good to check that the locations are nested within the labels here (e.g. by checking for a particular pcod in `dict["features"][0]["properties"]["inflows"]["Feature"].keys()`)

jc-harrison (changes requested):

flowmachine/flowmachine/features/location/labelled_flows.py
```diff
@@ -0,0 +1,137 @@
...
+        return (
+            [f"{col}_from" for col in loc_cols]
+            + [f"{col}_to" for col in loc_cols]
+            + self.out_label_columns
+            + ["value"]
+        )
+
+    @property
+    def out_label_columns(self) -> List[str]:
+        return [f"{col}_label" for col in self.label_columns]
```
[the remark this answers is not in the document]
Thingus: Stuck with `{col}_label` - as, at present, we can specify suffixes on `join`s but not prefixes. I think it might be worth adjusting this later, but it's not urgent.

flowmachine/flowmachine/features/location/labelled_flows.py
Thingus: Changed to `labels`; there's nothing special about subscriber labels as opposed to flow labels, and it might be worth doing if we break it out into a mixin later.

flowmachine/flowmachine/features/location/labelled_flows.py
```diff
@@ -0,0 +1,145 @@
...
+
+    def _make_query(self):
+
+        group_cols = ",".join(self.joined.column_names[1:])
```
jc-harrison, line 132: Can we explicitly specify the group columns (i.e. `[f"{col}_from" for col in loc_cols] + [f"{col}_to" for col in loc_cols] + self.out_label_columns`)? Currently this will group by all non-'subscriber' columns in `labels`, regardless of whether or not they are specified as label columns.

jc-harrison: And in light of this, ideally it would be good to have a test that columns in `labels` are not grouped-by if they're not specified in `label_columns`.

Thingus: I've moved the new property `out_spatial_columns` to this branch from `redacted_spatial_aggregate` - seems appropriate here.

flowmachine/tests/test_labelled_flows.py
```diff
@@ -0,0 +1,237 @@
...
+        right_append="_brand",
+    )
+    lf = LabelledFlows(
+        loc1=loc_1,
+        loc2=loc_2,
+        labels=labels,
+        label_columns=["value_hnd_type"],
+    )
+    assert "value_brand" not in lf.out_label_columns
+    assert "value_brand" not in get_dataframe(lf).columns
```
jc-harrison, line 207: If the column  _was_ there, it would be called "value_brand_label", so this test would still pass.

Perhaps better to check that only the expected columns _are_ present, rather than checking that the extra column is not.

jc-harrison (approved):
:+1:

PR #4754, opened 2022-01-04
Use Union's ability to take variadic args

### LINKED ISSUE #4753: Internal uses of Union should be variadic
Now that #4565 is merged, internal uses of `Union` should use the variadic form where possible, i.e. `ModalLocation`, `DayTrajectories` and `TotalActivePeriods`

greenape:
Closes #4753 

### I have:

### Description

Makes the handful of uses of `Union` internally variadic instead of reduces. Haven't added a changelog for this because it is really only internal.

### DIFF
--- a/flowmachine/flowmachine/core/union.py
+++ b/flowmachine/flowmachine/core/union.py
@@ -26,8 +26,8 @@ class Union(Query):
         column_set = set(tuple(query.column_names) for query in queries)
         if len(column_set) > 1:
             raise ValueError("Inconsistent columns.")
-        if len(queries) < 2:
-            raise ValueError("Union requires at least two queries.")
+        if len(queries) < 1:
+            raise ValueError("Union requires at least one query.")
 
         super().__init__()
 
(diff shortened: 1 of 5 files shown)


flowmachine/flowmachine/core/union.py
```diff
@@ -26,8 +26,8 @@ def __init__(self, *queries: Query, all: bool = True):
         column_set = set(tuple(query.column_names) for query in queries)
         if len(column_set) > 1:
             raise ValueError("Inconsistent columns.")
-        if len(queries) < 2:
-            raise ValueError("Union requires at least two queries.")
+        if len(queries) < 1:
+            raise ValueError("Union requires at least one query.")
```
jc-harrison, line 30: If we're allowing just one query, it's probably worth adding a warning if a user passes a single query with `all=False`, to warn that the output won't be de-duped (or, alternatively, handle this as a special case such that it _does_ de-dupe).

greenape: Former I think - it would be weirdly different behaviour to dedupe a single query.

jc-harrison (approved):
:+1:

PR #4793, opened 2022-01-13
Csv retrieval

Thingus:
Closes #ISSUE_NUMBER

### I have:

### Description

Implements the initial draft of the long-term storage -> flowdb pipeline (pictured, ignore the colours)
image
This adds the above facilities to create the above DAG to flowetl. The DAG is designed to be run on daily data, and the operators are as follows:
 - `Mount{x]`; Mounts that CSV as a foreign data wrapper
 - `CreateAndFillStagingTable`; Creates a table containing data common to all events and populates it from the foreign data wrappers
 - `DefaultLocationMapping`; Creates a table that maps cell tower IDs to themselves. May be replaced in the future by more sophisticated mappings.
 - `ApplyMappingToStagedEvents`; Replaces the cell IDs in the staged events table with their counterparts from the location mapping table.
 - `CreateSightingsTable`; Creates the reduced.sightings table (as designed by James) if it does not exist, to be filled by partitions.
 - `CreateAndFillDaySightingsTable`; Reworks the data in the staging table into the sightings format
 - `AppendSightingsToMainTable`; Attaches the daily table to the reduced.sightings table as a partition
 - `CleanupStagingTable`; Drops the staging table from `CreateAndFillStagingTable` and the foreign data wrappers from `Mount[x]`

It is built with the following assumptions (all env vars are in the flowetl container):
 - That a set of CSVs of the naming format [yyyymmdd]_[event_type] are accessible to the flowdb container local filesystem
 - That the CSVs have headings as set out in flowetl/flowetl/flowetl/operators/staging/event_columns.py
 - That the path to these is set in the environment variable `FLOWDB_CSV_DIR`
 - That the date to start running the DAG is set in `FLOWETL_CSV_START_DATE`, in the format yyyy-mm-dd
 - That the events to be staged are in `FLOWETL_EVENT_TYPES`, as a comma-seperated list.

### What has been added
 - A new subset of operators, `flowetl.operators.staging` that are relevent to building this DAG. These are mainly wrappers around a set of Jinja-templated queries found in `flowetl.operators.staging.sql`, with the following exceptions:
   -`mount_event_operator_factory` contains a function for creating the Mount operators - not every MNO will provide all of them.
   -`create_and_fill_staging_table` creates a Jinja template for mounting a set of event types.
 - `utils.create_staging_dag`; a function that takes a start date and a list of event types, and returns a DAG defining the pipeline for that set of event types.
 - Unit tests (`test_staging_operators.py`) and fixtures (added to `unit/conftest.py`). These are built on the postgresql_db fixture for building temporary databases, and a dummy one-task DAG.
 - An integration test of the new DAG, using a similar framework to existing DAG integration tests.
 - Static correctly formatted CSVs for testing

### What has changed
 - Workaround for the missing postgres public key in the flowetl Dockerfile
 - pytest-psql added to dev requirements

### Open questions
 - How should this DAG be triggered? Daily, on CSV creation, or by the other half of flowetl?
 - The location mapping step has some holes; there's no facility to update the entire reduced table if the cell tower mapping is changed, and it might just be not needed
 - How should we handle missing or empty CSVs? I'm leaning towards passing this problem off to the monitoring system.

### Still to do
 - Add the opt-out step

(there's a load of files from the benchmarking branch that have ended up in here, I'll sort them out tomorrow and get rid of this line)


flowetl.Dockerfile
```diff
@@ -38,20 +43,23 @@ WORKDIR /${SOURCE_TREE}/flowetl
 
 COPY . /${SOURCE_TREE}/
 
-RUN apt-get update && \
+
+RUN     apt-get update && \
```
greenape, line 47: Extra spaces?

flowetl.Dockerfile
```diff
@@ -38,20 +43,23 @@ WORKDIR /${SOURCE_TREE}/flowetl
...
         apt-get install -y --no-install-recommends git build-essential && \
         pip install --no-cache-dir pipenv && pipenv install --clear --deploy --system && \
         apt-get -y remove git build-essential && \
         apt purge -y --auto-remove && \
         rm -rf /var/lib/apt/lists/*
+RUN echo $PWD && ls $PWD && ls $PWD/flowetl
```
greenape, line 53: Remove.

flowetl.Dockerfile
```diff
@@ -38,20 +43,23 @@ WORKDIR /${SOURCE_TREE}/flowetl
...
 RUN apt-get update && \
         apt-get install -y --no-install-recommends git && \
         cd flowetl && \
         python setup.py install && \
         apt-get -y remove git && \
         apt purge -y --auto-remove && \
         rm -rf /var/lib/apt/lists/* && \
-        mv /${SOURCE_TREE}/flowetl/entrypoint.sh /
+        mv /${SOURCE_TREE}/flowetl/entrypoint.sh / && \
+	cp /${SOURCE_TREE}/flowetl/mounts/dags/ /opt/airflow/dags/
```
greenape: Why's this here?

flowetl.Dockerfile
```diff
@@ -38,20 +43,23 @@ WORKDIR /${SOURCE_TREE}/flowetl
...
 RUN apt-get update && \
         apt-get install -y --no-install-recommends git && \
         cd flowetl && \
         python setup.py install && \
         apt-get -y remove git && \
         apt purge -y --auto-remove && \
         rm -rf /var/lib/apt/lists/* && \
-        mv /${SOURCE_TREE}/flowetl/entrypoint.sh /
+        mv /${SOURCE_TREE}/flowetl/entrypoint.sh / && \
+	cp -r /${SOURCE_TREE}/flowetl/mounts/dags/ /opt/airflow/dags/
```
greenape, line 62: Why's this here? (May become apparent as I work thru)

flowetl/Pipfile
```diff
@@ -9,13 +9,13 @@ sqlalchemy = "<=1.3.23"
 click = "*"
 apache-airflow-upgrade-check = "*"
 wtforms = "<3.0.0"
+pytest-pgsql = "*"
```
greenape, line 12: Why's this here, and why not in dev?

flowetl/flowetl/flowetl/operators/staging/create_and_fill_staging_table.py
```diff
@@ -0,0 +1,52 @@
...
+
+    Parameters
+    ----------
+    event_type_list:List[str]
+        A list of event types; any or all of 'call','sms','location','mds','topup'
+    """
+
+    def __init__(self, *args, event_type_list: List[str], **kwargs):
+
+        for event_type in event_type_list:
```
greenape, line 23: Could do this with a set difference and then show all the bad'uns?

flowetl/flowetl/flowetl/operators/staging/create_and_fill_staging_table.py
```diff
@@ -0,0 +1,52 @@
...
+    and contact event type (defined in sql/event_type_enums.csv) from foreign tables mounted by operators
+    produced by mount_event_operator_factory.
+
+    Parameters
+    ----------
+    event_type_list:List[str]
+        A list of event types; any or all of 'call','sms','location','mds','topup'
+    """
+
+    def __init__(self, *args, event_type_list: List[str], **kwargs):
```
greenape, line 21: What's the `*args` about?

flowetl/flowetl/flowetl/operators/staging/create_sightings_table.py
```diff
@@ -0,0 +1,20 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+from airflow.operators.postgres_operator import PostgresOperator
+
+
+class CreateSightingsTable(PostgresOperator):
```
greenape, line 8: Should be built into flowdb really I think.

flowetl/flowetl/flowetl/operators/staging/mount_event_operator_factory.py
```diff
@@ -0,0 +1,43 @@
...
+from flowetl.operators.staging.event_columns import event_column_mappings
+
+
+def create_mount_event_operator(*, event_type: str):
+    """
+    A class factory function that produces an Airflow operator for mounting a foreign table containing data of
+    event_type.
+    Parameters
+    ----------
+    event_type:str
```
greenape, line 11: Should be spaces around the colon, and for a limited valid set use `{'call', 'sms'..}` instead of `str`

flowetl/flowetl/flowetl/operators/staging/mount_event_operator_factory.py
```diff
@@ -0,0 +1,43 @@
...
+        One of call,sms,location,mds,topup.
+    """
+    if event_type not in event_column_mappings.keys():
+        raise KeyError(f"{event_type} must be one of {event_column_mappings.keys()}")
+
+    columns = ",\n".join(
```
greenape, line 17: Could just `try` this and raise the error rather than checking above?

flowetl/flowetl/flowetl/operators/staging/sql/append_sightings_to_main_table.sql
```diff
@@ -0,0 +1,12 @@
+BEGIN;
+
+-- Partitions are inclusive on lower, exclusive on upper
+ALTER TABLE reduced.sightings
+ATTACH PARTITION reduced.sightings_{{ds_nodash}} FOR VALUES FROM (date('{{ds_nodash}}')) TO (date('{{ds_nodash}}')+ 1) ;
+
+ALTER TABLE reduced.sightings_{{ds_nodash}}
+DROP CONSTRAINT sightings_{{ds_nodash}}_check;
```
greenape, line 8: Why's the constraint being dropped out of curiosity?

flowetl/flowetl/flowetl/operators/staging/sql/append_sightings_to_main_table.sql
```diff
@@ -0,0 +1,12 @@
...
+
+ANALYZE reduced.sightings;
```
greenape, line 10: I would split this out I think, doesn't need to be in the same transaction.

flowetl/flowetl/flowetl/operators/staging/sql/apply_mapping_to_staged_events.sql
```diff
@@ -0,0 +1,9 @@
+BEGIN;
+
+UPDATE staging_table_{{ds_nodash}}
```
greenape, line 3: I think dispense with this altogether? The `UPDATE` is going to relatively super expensive and a source of errors.

flowetl/flowetl/flowetl/operators/staging/sql/cleanup_staging_table.sql
```diff
@@ -0,0 +1,12 @@
+BEGIN;
+
+DROP TABLE staging_table_{{ds_nodash}};
+DROP FOREIGN TABLE IF EXISTS call_table_{{ds_nodash}};
+DROP FOREIGN TABLE IF EXISTS sms_table_{{ds_nodash}};
+DROP FOREIGN TABLE IF EXISTS location_table_{{ds_nodash}};
+DROP FOREIGN TABLE IF EXISTS mds_table_{{ds_nodash}};
+DROP FOREIGN TABLE IF EXISTS topup_table_{{ds_nodash}};
+
+-- Vacuum or other upkeep here may be sensible?
```
greenape, line 10: You _should_ get an autovacuum triggered anyway, and also won't need a vacuum to reclaim space after a table drop.

flowetl/flowetl/flowetl/operators/staging/sql/create_and_fill_day_sightings_table.sql
```diff
@@ -0,0 +1,54 @@
...
+           array_agg(event_type ORDER BY date_time) AS cons_events
+    FROM group_ranked
+    GROUP BY msisdn, cell_id, group_rank
+    ORDER BY msisdn, row_id
+)
+INSERT INTO reduced.sightings_{{ds_nodash}}
+    SELECT '{{ds_nodash}}', convert_to(msisdn, 'LATIN1'), row_id, cell_id, cons_dates, cons_events
+    FROM aggregated;
+
+-- Clustering needs indexing; I think we leave it out for the time being
```
greenape, line 46: Move to a separate operator.

flowetl/flowetl/flowetl/operators/staging/sql/create_and_fill_day_sightings_table.sql
```diff
@@ -0,0 +1,54 @@
...
+--CLUSTER reduced.sightings_{{ds_nodash}}
+--USING subscriber_id, row_id;
+
+ANALYZE reduced.sightings_{{ds_nodash}};
```
greenape, line 50: Move to separate operator. (Or just use the existing analyse operator)

flowetl/flowetl/flowetl/operators/staging/sql/create_and_fill_day_sightings_table.sql
```diff
@@ -0,0 +1,54 @@
...
+                               ORDER BY min(date_time), max(date_time)
+                             ) AS row_id,
+           cell_id,
+           array_agg(date_time ORDER BY date_time) AS cons_dates,
+           array_agg(event_type ORDER BY date_time) AS cons_events
+    FROM group_ranked
+    GROUP BY msisdn, cell_id, group_rank
+    ORDER BY msisdn, row_id
+)
+INSERT INTO reduced.sightings_{{ds_nodash}}
```
greenape, line 42: I _think_ the issue here is that this will trigger WAL, which can be skipped if using a `CREATE TABLE AS ...`.

flowetl/flowetl/flowetl/operators/staging/sql/create_sightings_table.sql
```diff
@@ -0,0 +1,10 @@
+CREATE SCHEMA IF NOT EXISTS reduced;
+CREATE TABLE IF NOT EXISTS reduced.sightings(
+    sighting_date DATE ,  -- Actually partition
+    sub_id bytea NOT NULL, -- change to bytea
+    sighting_id INTEGER NOT NULL,
+    location_id text, --NOT NULL REFERENCES reduced.cell_location_mapping(location_id),  -- change to bytea
+    event_times TIME[],
+    event_types smallint[] -- change to enum
+) PARTITION BY RANGE (sighting_date);
```
greenape, line 9: Move to flowdb, but also please see https://github.com/Flowminder/FlowKit/blob/master/flowdb/bin/build/0030_schema_interactions.sql

and consider whether this should be using the dimension tables in there?

flowetl/flowetl/flowetl/operators/staging/sql/event_type_enums.csv
```diff
@@ -0,0 +1,8 @@
+event_type_string, enum
```
greenape, line 1: `interactions.d_event_type` may cover this?

flowetl/flowetl/flowetl/operators/staging/sql/opt_out.sql
```diff
@@ -0,0 +1,20 @@
+/*
+Drops every row from staging_table containing an msisdn from
+the list at opt_out_path.
+Quary args:
+-date
+-opt_out_path
+*/
+
+BEGIN;
```
greenape, line 9: I think remove at this time rather than add in something not fully realised?

flowetl/flowetl/flowetl/util.py
```diff
@@ -390,3 +390,96 @@ def create_dag(
         ]
     globals()[dag_id] = dag
     return dag
+
+
+def create_staging_dag(start_date: datetime, event_types: List[str], end_date=None):
```
greenape, line 395: See other dag creator for additional kwargs we would need to accept here.

flowetl/flowetl/flowetl/util.py
```diff
@@ -390,3 +390,96 @@ def create_dag(
...
+
+    Returns
+    -------
+    dag: DAG
+        The staging DAG.
+    """
+    from airflow import DAG
+    import os
+
+    template_folder = (
```
greenape, line 417: Use pathlib?

flowetl/flowetl/flowetl/util.py
```diff
@@ -390,3 +390,96 @@ def create_dag(
...
+        "/"
+        + os.getenv("SOURCE_TREE")
+        + "/flowetl/flowetl/flowetl/operators/staging/sql"
+    )
+
+    with DAG(
+        # How do we want to trigger this? Ask James./Jono
```
greenape, line 424: Should run daily but with retries, after some thought. But actually the function should allow you to configure this.

flowetl/flowetl/flowetl/util.py
```diff
@@ -390,3 +390,96 @@ def create_dag(
...
+        from flowetl.operators.staging.cleanup_staging_table import CleanupStagingTable
+        from flowetl.operators.staging.mount_event_operator_factory import (
+            create_mount_event_operator,
+        )
+
+        event_mount_operator_classes = [
+            create_mount_event_operator(event_type=event_type)
+            for event_type in event_types
+        ]
+        event_mount_instantiations = [op() for op in event_mount_operator_classes]
```
greenape, line 463: As mentioned, I think the dag failing if one event type is missing on one day is an issue. I've thought it over, and my feeling is that while the monitoring should of course report this, creating a bogus file to enable this to pass or whatever isn't really tenable and will cause pain in future, as will attaching a bunch of empty partitions.

flowetl/mounts/dags/append_sightings_to_main_table_dag.py
```diff
@@ -0,0 +1,15 @@
...
+from airflow import DAG
+
+from datetime import datetime
+
+from flowetl.util import create_staging_dag
+
+
+dag = create_staging_dag(
+    start_date=datetime.strptime(os.getenv("FLOWETL_CSV_START_DATE"), "%Y-%m-%d"),
+    event_types=os.getenv("FLOWETL_EVENT_TYPES", "call,location,sms,mds,topup").split(
```
greenape, line 12: Shouldn't need to kill a docker container to change the event types I think.

flowetl/tests/integration/conftest.py
```diff
@@ -369,10 +375,11 @@ def wait_for_container(
         flowdb_container()
         flowetl_db_container()
         logger.info("Started FlowETL container")
+
         yield container
 
         save_airflow_logs = (
-            os.getenv("FLOWETL_INTEGRATION_TESTS_SAVE_AIRFLOW_LOGS", "FALSE").upper()
+            os.getenv("FLOWETL_INTEGRATION_TESTS_SAVE_AIRF" "LOW_LOGS", "FALSE").upper()
```
greenape, line 382: Typo?

flowetl/tests/integration/conftest.py
```diff
@@ -9,9 +9,11 @@
 import os
 import random
 import string
+import tarfile
 import warnings
 from pathlib import Path
 from subprocess import Popen, run
+from tempfile import NamedTemporaryFile
```
greenape, line 16: For why?

flowetl/tests/integration/conftest.py
```diff
@@ -9,9 +9,11 @@
 import os
 import random
 import string
+import tarfile
```
greenape, line 12: What for?

flowetl/tests/unit/conftest.py
```diff
@@ -20,3 +42,78 @@ def unload_airflow():
         for module in list(sys.modules.keys()):
             if "airflow" in module or "flowetl" in module:
                 del sys.modules[module]
+
+
+# In hindsight, these could have been a factory or partial. Ho hum.
+@pytest.fixture()
+def dummy_db_conn(postgresql_db, monkeypatch):
+    postgresql_db.install_extension("file_fdw")
```
greenape, line 50: This is so the tests can be run against not-flowdb?

flowetl/tests/unit/test_csv_ingestion.py
```diff
@@ -0,0 +1,85 @@
...
+import tarfile
+from tempfile import NamedTemporaryFile
+from pathlib import Path
+
+
+import pytest
+
+
+@pytest.fixture
+def flowmachine_env(monkeypatch):
```
greenape, line 12: Best handled outside the tests, no? pipenv does this itself, or one can use pytest-dotenv or similar?

greenape (changes requested):
I've had a preliminary look and made a few comments. Can you strip out the stuff not relevant to this PR as well?

flowetl/flowetl/flowetl/operators/staging/sql/append_sightings_to_main_table.sql
jc-harrison: Constraint isn't necessary once the partition is attached (it's just duplicating what the declarative partitioning is doing). But it's required earlier to prevent postgres needing to do a full table scan when attaching the partition.

flowetl.Dockerfile
Thingus: Airflow goes looking for it's DAGs in /opt/airflow/dags by default - instead of messing with the Airflow config, I thought it better (closer to the airflow, one less configuration to mess about with) to copy the DAGs on build.

flowetl/tests/unit/conftest.py
Thingus: Yep - testing the operators individually, not the flowdb integration.

flowetl/flowetl/flowetl/operators/staging/create_and_fill_staging_table.py
Thingus: Given this is basically a Postgres operator with a special case, I wasn't sure how Airflow calls or passes it around internally, or if there's a convention I've missed - hence *args and **kwargs. Now there's unit tests in place, I can take them out and see if they still work.

flowetl/flowetl/flowetl/operators/staging/sql/cleanup_staging_table.sql
Thingus: :+1:

flowetl/flowetl/flowetl/operators/staging/sql/append_sightings_to_main_table.sql
Thingus: You reckon different `begin .. commit` block, or separate operator?

flowetl/flowetl/flowetl/operators/staging/sql/apply_mapping_to_staged_events.sql
Thingus: Yeah, we can put it back later.

flowetl/tests/integration/conftest.py
Thingus: Once there was a hairy test fixture that pushed files to a running container, but it's gone now. Will drop.

flowetl/tests/integration/conftest.py
Thingus: Ditto

flowetl/tests/unit/test_csv_ingestion.py
Thingus: We can drop this entire file, actually - this is a holdover from before Airflow that I forgot to drop.

flowetl/flowetl/flowetl/operators/staging/mount_event_operator_factory.py
```diff
@@ -0,0 +1,48 @@
+from typing import Literal
+
+from airflow.operators.postgres_operator import PostgresOperator
+from flowetl.operators.staging.event_columns import event_column_mappings
+
+
+def create_mount_event_operator(
```
jc-harrison, line 7: I'd prefer it if this was an operator class with a parameterised sql string rather than a factory function.

I.e. use `{{ params.event_type }}` in the sql. Could probably define a macro to handle the columns (i.e. `{{ columns_string(params.event_type) }}`).

flowetl/flowetl/flowetl/operators/staging/append_sightings_to_main_table.py
```diff
@@ -0,0 +1,20 @@
...
+class AppendSightingsToMainTable(PostgresOperator):
+    """
+    Attaches reduced.sightings_DATE to the main reduced.sightings table as a partition, dropping the date
+    constraint as it does so.
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(
+            *args,
+            task_id="AppendSightingsToMainTable",
```
jc-harrison, line 17: I think setting the task ID should really be left to the DAG definition, rather than hard-coded in the operator class (i.e. set the `task_id` argument in `create_staging_dag`, not here).

flowetl/flowetl/flowetl/operators/staging/sql/create_and_fill_day_sightings_table.sql
Thingus: I think that'd lose us the constraint-at-table-creation, which is needed for the partitioning. Maybe we can disable the WAL with `CREATE UNLOGGED TABLE`, given this is only a transient table created as part of the DAG?

flowetl/flowetl/flowetl/operators/staging/append_sightings_to_main_table.py
jc-harrison: In fact, I'm not entirely convinced it's worth defining these custom operator classes at all, if they're just wrapping a sql script - could use `PostgresOperator`s directly in `create_staging_dag` (i.e. `append_sightings_to_main_table = PostgresOperator(task_id="AppendSightingsToMainTable", sql="append_sightings_to_main_table.sql")`).

flowetl/flowetl/flowetl/operators/staging/sql/event_type_enums.csv
jc-harrison: It doesn't quite, as things stand at the moment, because we want to identify incoming and outgoing calls/sms as different event types. But I think it would make sense to be able to use `interactions.d_event_type` here.

flowetl.Dockerfile
jc-harrison: The DAGs dir should be bind-mounted in so that users can define their own DAGs (and customise the standard DAGs if necessary) - as such, I don't think it can make sense to copy anything into the DAGs dir on build.

I don't think we should need to copy any flowetl internals into the DAGs dir, as long as users can import the required functions/classes from flowetl in their own DAG definitions. The DAGs defined in `flowetl/mounts/dags/` are useful for integration testing, and as a reference for defining production DAGs, but I think it's better if the actual python files that define the DAGs in a FlowKit deployment are user-supplied.

flowetl/flowetl/flowetl/operators/staging/create_and_fill_staging_table.py
```diff
@@ -0,0 +1,50 @@
...
+            f"""
+            SELECT
+                MSISDN,
+                IMEI,
+                IMSI,
+                TAC,
+                CELL_ID,
+                DATE_TIME,
+                EVENT_TYPE
+            FROM {event_type}_table_{{{{ds_nodash}}}}
```
jc-harrison, line 37: It would be nice to have a static sql template that references `{{ params.event_type_list }}` - seems cleaner to me to have the event types set via params than to hard-code them into the sql when the DAG is defined. Could be achieved via a jinja for loop, I suspect.

jc-harrison (commented):
I'm wondering whether it's worth splitting the DAG into two - one to ingest from the csv files into the "full" events tables (this DAG doesn't technically do that bit at the moment, but it does create the "staging table" as an intermediate step, which is largely the same thing), and a second to populate the "sightings" table from the events tables.

The first DAG can then run independently for each event type, and the second DAG doesn't need to be concerned about event-type-specific columns because it only need s the columns that are common to all event types. We could create a combined-events table that inherits from all the per-event-type event tables (only the common columns), which could then take the place of the staging table.

My thinking is that we may want to create the sightings table based on only a subset of event types, or other choices e.g. the choice of spatial mappings. I.e. the exact way in which the sightings table is populated might vary from project to project (in situations where a FlowKit deployment is being used predominantly by one project at a time), whereas the ingestion into the events tables is unlikely to change.

jc-harrison (commented):
Something for a later PR perhaps, but I think it would be worth adding a step that invalidates any cache tables that depend on the newly-ingested day of data (in case we're re-ingesting a day that was already previously ingested, but for which we have received new data). Better to do this on re-ingest than on drop, I think, because we might want to keep cached results around after the data they were calculated from has been dropped (but we wouldn't want to keep cached results around if re-running the query would produce a different result, and it seems reasonable to assume that if we've re-ingested a day of data then the data for that day have changed).

flowetl/flowetl/flowetl/operators/staging/create_and_fill_staging_table.py
Thingus: :+1:

flowetl/flowetl/flowetl/operators/staging/append_sightings_to_main_table.py
Thingus: I was wondering it while I was making them - the thinking was that any preprocessing of parameters before applying the template can live in the custom operators, but those turned out not to be needed in pretty much every case, or Jinja could take care of it. I'll take them out.

Thingus:
Re @jc-harrison 's comments about splitting the DAG - would this eventually lead to three DAGs?
CSV -> 'staging' (always used, can be run with subset of events)
'staging' -> 'sighting' (used for 'lite' queries, that don't need event-specific information, always happens)
'staging' -> 'full' (used for event-specific queries, only happens if that analysis is needed)

flowetl/tests/integration/conftest.py
Thingus: Yep

Thingus:
Done:
  Broken the CSV DAG into two
  Added 'analyse' tasks to the end of both DAGs
  Moved sightings table schema into flowdb
  Rewrote sighting table schema to add a constraint post-CREATE
  Changed schemas and test data to merge the 'caller' and 'callee' event types, and added a bool column to the CSV for is_incoming
Still to do:
  A constraint to the dimension table in flowdb
  A staging sensor for the DAG
DAG tests

PR #4940, opened 2022-03-07
Upgrade airflow

greenape:
This updates Airflow to 2.2.3.

It dispenses with most of our startup script for FlowETL, and of necessity will be a breaking change release because airflow updates are not reversible.

Need to add some update notes for this, because:

- Existing airflow DBs should be backed up before starting, or any upgrade cannot be rolled back
- The way in which airflow should be run has changed (previously all in one container, this is still possible but requires extra steps and should be avoided)


flowetl/tests/unit/test_qa_check_discovery.py
```diff
@@ -121,7 +121,7 @@ def test_module_path_added_to_dag_template_locations():
     dag = DAG("DUMMY_DAG", start_date=datetime.now())
     get_qa_checks(dag=dag)
     assert (
-        str(Path(__file__).parent.parent.parent / "flowetl" / "flowetl" / "qa_checks")
+        Path(__file__).parent.parent.parent / "flowetl" / "flowetl" / "qa_checks"
```
Thingus, line 124: Do we need the extra layer of 'flowetl' in the tree, or can that be flattened out?

Thingus (approved):
Looks good - I'd say it's worth editing the compose files so we have flowetl_scheduler, flowetl_webserver (both just flowetl with the appropriate entry point), but that's a related PR I think, Just needs codecov to be happy.

flowetl/flowetl/flowetl/mixins/wrapping_sql_mixin.py
```diff
@@ -7,7 +7,13 @@
 
 class WrappingSQLMixin:
     def prepare_template(self) -> None:
-        self.sql = self.wrapper_sql.format(sql=self.sql)
+        try:
+            # See if the sql to be wrapped lives in a file
+            jinja_env = self.dag.get_template_env()
+            sql = jinja_env.loader.get_source(jinja_env, self.sql)[0]
+        except:
```
Thingus, line 14: Do we need an explicit exception here?

flowetl/flowetl/flowetl/util.py
```diff
@@ -261,12 +271,6 @@ def create_dag(
     from flowetl.sensors.table_flux_sensor import TableFluxSensor
 
     # Choose flux sensor
-    if use_file_flux_sensor is not None:
-        message = "The 'use_file_flux_sensor' argument is deprecated."
-        if use_flux_sensor == True:
-            use_flux_sensor = "file" if use_file_flux_sensor else "table"
-            message = f"{message} Set use_flux_sensor='{use_flux_sensor}' instead."
-        warnings.warn(message, DeprecationWarning)
```
jc-harrison, line 269: If we're removing the deprecation warning, can we also remove the `use_file_flux_sensor` argument, please?

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Added
 
 ### Changed
+- __Action Needed__ Airflow updated to version 2.3.3; **backup flowetl_db before applying update** [#4940](https://github.com/Flowminder/FlowKit/pull/4940)
```
jc-harrison, line 12: I think it would be worth adding this to the docs as well - perhaps a warning [here](https://flowkit.xyz/administrator/deployment/) for anyone upgrading from version 1.16.0 or earlier.

flowetl/flowetl/flowetl/util.py
Thingus: :+1:

Thingus: Done and changelog entry added

jc-harrison (approved):
:+1:

PR #4954, opened 2022-03-11
Non-blocking unstored dependencies graph

### LINKED ISSUE #4949: unstored_dependencies_graph should not block while queries are running/resetting
At present `unstored_dependencies_graph` blocks while queries are in a mutating state (running, resetting). It should not, because this means returns from running a query via api can be excessively slow.

jc-harrison:
Closes #4949

### I have:

### Description

Modifies `unstored_dependencies_graph` to exclude a query's dependencies from the graph if the query is stored _or queued or executing_, instead of waiting for the query to finish executing before checking whether it is stored.

This way, a query's dependencies are still always added to the queue before the query itself. One consequence, though, is that if a dependency gets cancelled before the parent query starts executing, that dependency will not get re-run.


greenape (approved):
Be good to have a test really, but I think sensible.

PR #5012, opened 2022-03-21
Airflow celery

### LINKED ISSUE #4950: CelerySchedulder for Airflow
Airflow webserver, scheduler and executors are all presently running on the same container - they should be spun out.

Change the docker compose files + switch to CeleryExecutor so this is the case

Thingus:
Closes #4950

### I have:

### Description
docker-compose
(image generated with https://github.com/pmsipilot/docker-compose-viz)
Replaces components for flowetl with the needed infrastructure for the worker-based Celery backend.
Celery is a brokering service for splitting tasks between Docker worker containers
New containers are:
 - flowetl_redis: Basic Redis container as a Celery broker
 - flowetl_db: Basic Postgres container as an Airflow and Celery backend
 - flowetl_init: Flowetl container with an overriden default entrypoint - runs `airflow db init` and creates webserver user (admin:admin, at present). Runs command then exits.
 - flowetl_webserver: Flowetl container running the Airflow web frontend
 - flowetl_triggerer: Flowetl container running the DAG triggering service
 - flowetl_scheduler: Flowetl container running the DAG scheduler
 - flowetl_flower: Flowetl container running the Celery web frontend for worker management.
 - flowetl_worker: Celery worker containers - number spun up is set by `FLOWETL_WORKER_COUNT`

Startup sequence is as follows:
```mermaid
sequenceDiagram
  participant db
  participant redis
  participant init
  participant webserver
  participant scheduler
  participant triggerer
  participant workers
  participant flower

    Note over db,redis: Start
    db->>+init: started
    redis->>+init: started
    Note over init: Run init
    init->>db: Initialise database
    init->>db: Add default Airflow user
    Note over init: Stop init container
    init ->> webserver: start
    init ->> scheduler: start
    init ->> triggerer: start
    init ->> workers: start
    init ->> flower: start
```

New environment variables introduced:
`FLOWETL_WORKER_COUNT`: Number of worker containers to create. Defaults to 1 in compose file.
`FLOWETL_CELERY_PORT`: External port used for Flower access
`CELERY_CONN`: Connection string for Celery backend


greenape:
Applicable to the [secrets quick-start](https://github.com/Flowminder/FlowKit/blob/master/secrets_quickstart/flowetl.yml) as well? (

flowstef (commented):
Overall it looks good. Split ETL into multiple containers connected via a network seems like a good idea. Startup sequence makes sense too.
Not sure how I can test this.

flowetl/Pipfile
```diff
@@ -4,13 +4,15 @@ verify_ssl = true
 name = "pypi"
 
 [packages]
-apache-airflow = {extras = ["crypto"],version = "==2.2.3"}
```
greenape, line 7: crypto extra is no longer required?

flowetl.Dockerfile
```diff
@@ -24,10 +24,10 @@ ENV SOURCE_VERSION=${SOURCE_VERSION}
 ENV SOURCE_TREE=FlowKit-${SOURCE_VERSION}
 WORKDIR /${SOURCE_TREE}/flowetl
 
-COPY --chown=airflow . /${SOURCE_TREE}/
-
+COPY --chown=airflow:1 . /${SOURCE_TREE}/
```
greenape, line 27: Why's the gid 1?

docker-compose.yml
```diff
@@ -7,15 +7,49 @@
...
+  environment:
+    &flowetl-envs
+    AIRFLOW__CORE__EXECUTOR: CeleryExecutor
+    AIRFLOW__CORE__SQL_ALCHEMY_CONN: ${SQL_ALCHEMY_CONN:?Must set SQL_ALCHEMY_CONN env var}
+    AIRFLOW_CONN_FLOWDB: postgresql://$POSTGRES_USER:$POSTGRES_PASSWORD@flowdb:5432/flowdb
+    AIRFLOW__CORE__FERNET_KEY: ${FLOWETL_FERNET_KEY:?Must set FLOWETL_FERNET_KEY env var}
+    AIRFLOW__CELERY__RESULT_BACKEND: ${CELERY_CONN:?Must set CELERY_CONN env var}
+    AIRFLOW__CELERY__BROKER_URL: redis://:@flowetl_redis:6379/0
+    POSTGRES_HOST: flowetl_db
+    FLOWETL_AIRFLOW_ADMIN_USERNAME: "admin"
```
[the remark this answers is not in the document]
Thingus: These are default/shared between flowetl instances except for the `FLOWETL_AIRFLOW_ADMIN_USERNAME` which I think can be removed

docker-compose.yml
```diff
@@ -144,48 +178,104 @@ services:
...
+    container_name: flowetl_webserver
+    <<: *flowetl-common
+    ports:
+      - ${FLOWETL_PORT:?Must set FLOWETL_PORT env var}:8080
+    command: webserver
     networks:
+      - flowetl
+    environment:
+      <<: *flowetl-envs
+      _AIRFLOW_WWW_USER_USERNAME: admin
```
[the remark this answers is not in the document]
Thingus: Ditto above

flowetl.Dockerfile
Thingus: Airflow image issue - it dropped the group in the v2 image. Might've been fixed in the rc1 revisiion - I'll check

flowetl/Pipfile
Thingus: It's included in the airflow image itself - various airflow requirements are put in here to expose them to the graph builder and test suite.

docker-compose.yml
```diff
@@ -267,7 +263,7 @@ services:
   flowetl_init:
     container_name: flowetl_init
     <<: *flowetl-common
-    entrypoint: /bin/bash -c "airflow db init && airflow users create --role Admin --username admin --email admin --firstname admin --lastname admin --password admin"
+    entrypoint: /bin/bash -c "airflow db init && airflow users create --role Admin --username admin --email admin --firstname admin --lastname admin --password ${FLOWETL_WEBSERVER_PASSWORD:?Must set FLOWETL_WEBSERVER_PASSWORD env var}"
```
greenape, line 266: Tricky question, but where will this get substituted? Will compose do it, or will it happen inside the container in bash? Cos if compose you'll be able to see it by inspecting the container, right?

Also should this be `command` rather than entrypoint, or do we need to deliberately override the entrypoint for some reason?

Thingus: Compose subs at run - they're basically preprocessor directives here.
This does need to be `entrypoint` as `command` goes through the Airflow entrypoint by default - this then continues to do Airflow things, the service never finishes and the compose does not continue.
The container gets stopped on completion, so the only place that this should appear is in it's `~/,history`, and that can be skipped with a leading space before the command

greenape: Really? `docker run apache/airflow db init` seems to terminate? Are we doing something in our entry point to break that behaviour?

Also, there's an `init.sh` in our image which does the db init and creates a user, no?

Thingus: As far as I can tell, we're not calling `init.sh` directly, but I think that it's responsibilities are better split up between the 'init' container, the webserver container and wherever the pool setup lives (I suspect the scheudler). Am investigating what's going on with `flowetl db init` now

Thingus: No, it's terminates fine now. Weird; entrypoint removed, replaced with `bash -c "blabla"`

flowetl/entrypoint.sh
```diff
@@ -30,19 +30,28 @@ done
...
-: "${AIRFLOW__WEBSERVER__SECRET_KEY:=${AIRFLOW__WEBSERVER__SECRET_KEY:=$(python -c "import os; print(os.urandom(16))")}}"
+## Defaults and back-compat
+# These could mask missing
+#: "${AIRFLOW__CORE__FERNET_KEY:=${FERNET_KEY:=$(python -c "from cryptography.fernet import Fernet; FERNET_KEY = Fernet.generate_key().decode(); print(FERNET_KEY)")}}"
+#: "${AIRFLOW__CORE__EXECUTOR:=${EXECUTOR:-Sequential}Executor}"
+#: "${AIRFLOW__WEBSERVER__SECRET_KEY:=${AIRFLOW__WEBSERVER__SECRET_KEY:=$(python -c "import os; print(os.urandom(16))")}}"
+
+# Aliasing flowetl-specific secrets to their general use
+
+: "REDIS_PASSWORD"=${FLOWETL_REDIS_PASSWORD:?FLOWETL_REDIS_PASSWORD not defined; check secrets}
```
greenape, line 41: Given the container can still actually be run with local scheduler, can these be required only if celery is being used?

Thingus:
Would we need to update the integration tests to include checks for Celery?

greenape (approved):

PR #5117, opened 2022-04-29
Flows api endpoint

### LINKED ISSUE #2029: Inflows/outflows methods in FlowClient?
We could consider adding FlowClient helper methods to calculate in/outflows from a flows query result. (It's not a complicated calculation, so I'm not sure how useful this would be in practice. Just a thought).

Thingus:
Closes #2029 

### I have:

### Description
 
Exposes redacted inflow and outflow methods


flowmachine/flowmachine/core/server/query_schemas/flows.py
```diff
@@ -58,3 +81,21 @@ class FlowsSchema(BaseSchema):
     join_type = fields.String(validate=OneOf(Join.join_kinds), missing="inner")
 
     __model__ = FlowsExposed
+
+
+class InflowsSchema(BaseSchema):
+    query_kind = fields.String(validate=OneOf(["flows"]))
```
greenape, line 87: Can just subclass `FlowsSchema` and override `query_kind` to fields.String(validate=OneOf(["inflows"])) and `__model__ = FlowsExposed`?

Thingus: Done - also applied similar structure to InflowsExposed and Outflows Exposed

flowmachine/flowmachine/core/server/query_schemas/flows.py
```diff
@@ -50,6 +50,18 @@ def _flowmachine_query_obj(self):
         return RedactedFlows(flows=Flows(loc1, loc2, join_type=self.join_type))
 
 
+class InflowsExposed(FlowsExposed):
```
greenape, line 53: Ok, next step is to make this be the redacted inflow of the unredacted flows obj.

Thingus: Done now

flowmachine/flowmachine/features/location/redacted_flows.py
```diff
@@ -5,13 +5,14 @@
...
-from flowmachine.features.location.flows import FlowLike
+from flowmachine.features.location.flows import FlowLike, BaseInOutFlow
 from flowmachine.features.location.redacted_location_metric import (
     RedactedLocationMetric,
 )
 
 
-class RedactedFlows(RedactedLocationMetric, FlowLike, Query):
+# As far as
+class RedactedFlowsLike(RedactedLocationMetric, FlowLike, Query):
```
greenape, line 15: Can't combine these I think because an in/out flow is not flows-like anymore (no from-to), so better two separate classes.

Thingus: Thought I could get away with it as the queries produced are identical - but better to have a RedactedInOutFlows class

flowmachine/flowmachine/features/location/redacted_flows.py
```diff
@@ -22,9 +22,20 @@ class RedactedFlows(RedactedLocationMetric, FlowLike, Query):
...
-    def __init__(self, *, flows: Flows):
-
+    def __init__(self, *, flows: FlowLike):
         self.redaction_target = flows
         # self.spatial_unit is used in self._geo_augmented_query
         self.spatial_unit = flows.spatial_unit
         super().__init__()
+
+
+class RedactedInOutFlow(RedactedLocationMetric, BaseInOutFlow, Query):
```
greenape, line 32: Move to own file I think.

flowmachine/flowmachine/features/location/redacted_flows.py
```diff
@@ -22,9 +22,20 @@ class RedactedFlows(RedactedLocationMetric, FlowLike, Query):
...
+    """
+    An object representing the summation
```
greenape, line 34: Add rest of docstring - Params etc.

flowmachine/tests/test_redacted_flows.py
```diff
@@ -17,3 +17,35 @@ def test_all_above_threshold(get_dataframe):
...
     )
+
+
+def test_in_above_threshold(get_dataframe):
+    """
+    Test that inflows redacts correctly
+    """
+    assert all(
+        get_dataframe(
+            RedactedFlows(
```
greenape, line 28: These need to be updated, right?

flowmachine/flowmachine/core/server/query_schemas/flows.py
```diff
@@ -50,6 +53,26 @@ def _flowmachine_query_obj(self):
...
+    @property
+    def _flowmachine_query_obj(self):
+        loc1 = self.from_location._flowmachine_query_obj
+        loc2 = self.to_location._flowmachine_query_obj
+        return RedactedInOutFlow(
+            flows=Flows(loc1, loc2, join_type=self.join_type).inflow()
+        )
+
+
+class OutflowsExposed(FlowsExposed):
```
greenape, line 66: Maybe nicer to move these to their own files as well?

greenape (changes requested):
Nice. Couple of suggestions, one thing to add, one to fix.

I think once those are done we merge, and then open a new issue to cover the suggestion from #2029 of optionally excluding stays.

flowmachine/flowmachine/features/location/redacted_in_out_flows.py
```diff
@@ -0,0 +1,23 @@
...
+    -----
+    in_out_flow: BaseInOutFlow
+        The underlying Inflow or Outflow to redact; normally created from `flows.inflow` or `flows.outflow`
+    """
+
+    def __init__(self, *, in_out_flows: BaseInOutFlow):
+        self.redaction_target = in_out_flows
+        # self.spatial_unit is used in self._geo_augmented_query
+        self.spatial_unit = in_out_flows.spatial_unit
+        super().__init__(in_out_flows.flow)
```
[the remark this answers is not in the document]
Thingus: BaseInOutFlow needs a Flow as a constructor argument. Reckon we do 
```suggestion
        super().__init__(flow = in_out_flows.flow)
```

flowmachine/flowmachine/features/location/redacted_in_out_flows.py
```diff
@@ -0,0 +1,23 @@
+from flowmachine.core import Query
+from flowmachine.features.location.flows import BaseInOutFlow
+from flowmachine.features.location.redacted_location_metric import (
+    RedactedLocationMetric,
+)
+
+
+class RedactedInOutFlow(RedactedLocationMetric, BaseInOutFlow, Query):
```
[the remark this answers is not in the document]
greenape: Note you'll also lose the index columns (although one could fix that in `RedactedLocationMetric` as well

greenape (changes requested):

flowclient/flowclient/aggregates.py
```diff
@@ -750,6 +778,57 @@ def flows(*, connection: Connection, **kwargs) -> APIQuery:
...
+@merge_args(outflows_spec)
+def outflows(*, connection: Connection, **kwargs) -> APIQuery:
+    """
+    Given a set of from- and to- locations, returns a count of every user that travelled *from* each *origin* location.
+
+    Parameters
+    ----------
+    connection : Connection
+        FlowKit API connection
+    from_location: dict
```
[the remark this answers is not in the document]
Thingus:
```suggestion
    from_location : dict
```

flowclient/flowclient/aggregates.py
```diff
@@ -750,6 +778,57 @@ def flows(*, connection: Connection, **kwargs) -> APIQuery:
...
+        Query which maps individuals to single location for the "origin" period of interest.
+    to_location: dict
```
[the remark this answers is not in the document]
Thingus:
```suggestion
    to_location : dict
```

flowmachine/flowmachine/features/location/redacted_in_out_flows.py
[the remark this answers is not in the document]
jc-harrison: Just catching up on this conversation. It looks to me like there's currently no need for `RedactedInOutFlow` to inherit from `BaseInOutFlow`, but fixing https://github.com/Flowminder/FlowKit/issues/5132 will require overriding `_geo_augmented_query` in `BaseInOutFlow`, and that will also be required here. So the current inheritance makes sense in the long run.

flowmachine/tests/test_redacted_flows.py
```diff
@@ -17,3 +18,35 @@ def test_all_above_threshold(get_dataframe):
...
+    assert all(
+        get_dataframe(
+            RedactedInOutFlow(
+                in_out_flows=Flows(
+                    daily_location("2016-01-01"), daily_location("2016-01-01")
+                ).outflow()
+            )
+        ).value
+        > 15
+    )
```
jc-harrison, line 52: Might be nice to have a test that checks the summation happens before redaction (as opposed to redacting the flows first and then summing the redacted values).

jc-harrison (commented):
I've added a suggestion for an additional test that might be nice to have, but otherwise I'm happy if @greenape is happy.

flowmachine/flowmachine/features/location/redacted_in_out_flows.py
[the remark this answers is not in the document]
greenape: Hmm. I disagree - the in/outflows shouldn't be retaining the from/to suffix to the location columns, rather than needing an override of the geojsoning.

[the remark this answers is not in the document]
jc-harrison: Ah, true. Without the from/to suffix the in/outflows are just like the result of any other spatial aggregate, so then there's no special treatment required and redacted in/outflows should just need the `RedactedLocationMetric` and `GeoDataMixin` mixins, the same as other metrics.

There's the issue that `BaseInOutFlow` also handles labelled in/outflows, and the default geojson representation isn't ideal for that (each geography will be duplicated `n_labels` times in the output).  But that's an issue more widely than just for in/outflows (e.g. `LabelledSpatialAggregate` has the same issue), and it isn't relevant for this PR.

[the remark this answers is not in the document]
greenape: Yup, https://github.com/Flowminder/FlowKit/commit/c5a780163e92cad7152d651181f11c3b258fbe0f is my proposed solution to that.

greenape (approved):

PR #5158, opened 2022-05-19
Pg median

### LINKED ISSUE #4511: Update median implementation
We define a `median` aggregate function in FlowDB ([here](https://github.com/Flowminder/FlowKit/blob/a247bf65184c5cdccd5fde0af501651c2368889f/flowdb/bin/build/0030_utilities.sql#L130)). The definition is taken from https://wiki.postgresql.org/wiki/Aggregate_Median, which now starts with the following message:

> **_NOTE: This page is obsolete._**
> 
> To get the median in PostgreSQL, use `percentile_cont(0.5) WITHIN GROUP (ORDER BY num_value)`. See [the documentation](https://www.postgresql.org/docs/current/sql-expressions.html#SYNTAX-AGGREGATES) for more details.

We should update the median implementation in FlowDB.

greenape:
Closes #4511

### I have:

### Description

Removes flowdb's median function and updates usages in flowmachine to use the Postgres builtin.

### DIFF
--- a/flowmachine/flowmachine/features/subscriber/displacement.py
+++ b/flowmachine/flowmachine/features/subscriber/displacement.py
@@ -150,7 +150,7 @@ class Displacement(SubscriberFeature):
         sql = f"""
         SELECT 
             subscriber,
-            {self.statistic:{f'COALESCE(value_dist, 0) * {multiplier}'}} as value
+            {self.statistic:COALESCE(value_dist, 0) * {multiplier}} as value
         FROM 
             ({self.joined.get_query()}) _
         GROUP BY 
--- a/flowmachine/flowmachine/features/subscriber/displacement.py
+++ b/flowmachine/flowmachine/features/subscriber/displacement.py
@@ -12,7 +12,7 @@ from flowmachine.features.spatial import DistanceMatrix
 from .metaclasses import SubscriberFeature
 from ..utilities.subscriber_locations import SubscriberLocations, BaseLocation
 from flowmachine.core import Query
-from flowmachine.utils import standardise_date
+from flowmachine.utils import standardise_date, get_stat
 
 valid_stats = {"sum", "avg", "max", "min", "median", "stddev", "variance"}
 
@@ -159,7 +159,7 @@ class Displacement(SubscriberFeature):
         sql = f"""
         SELECT 
             subscriber,
-            {self.statistic}(COALESCE(value_dist, 0) * {multiplier}) as value
+            {get_stat(self.statistic, f'COALESCE(value_dist, 0) * {multiplier}')} as value
         FROM 
             ({self.joined.get_query()}) _
         GROUP BY 
--- a/flowmachine/flowmachine/features/subscriber/displacement.py
+++ b/flowmachine/flowmachine/features/subscriber/displacement.py
@@ -37,9 +37,8 @@ class Displacement(SubscriberFeature):
         The set of home locations from which to calculate displacement.
         If not given then ModalLocation Query wil be created over period
         start -> stop.
-    statistic : str
-        the statistic to calculate one of 'sum', 'avg', 'max', 'min',
-        'median', 'stddev' or 'variance'
+    statistic : Statistic, default Statistic.AVG
+        The statistic to calculate.
     unit : {'km', 'm'}, default 'km'
         Unit with which to express the answers, currently the choices
         are kilometres ('km') or metres ('m')
@@ -83,7 +82,7 @@ class Displacement(SubscriberFeature):
         start: str,
         stop: str,
         reference_location: BaseLocation,
-        statistic: str = "avg",
+        statistic: Statistic = Statistic.AVG,
         unit: str = "km",
         hours: Union[str, Tuple[int, int]] = "all",
         table: Union[str, List[str]] = "all",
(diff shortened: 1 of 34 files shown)


flowmachine/flowmachine/features/subscriber/handset_stats.py
```diff
@@ -4,7 +4,7 @@
 
 # -*- coding: utf-8 -*-
 
-valid_stats = {"count", "sum", "avg", "max", "min", "median", "stddev", "variance"}
+valid_stats = {"count", "sum", "avg", "max", "min", "stddev", "variance"}
```
jc-harrison, line 7: Would be nice to support weighted median here, I suppose. But that's out of scope for this PR, and quite possibly something that would never actually be used.

jc-harrison: Could do with updating the docstring accordingly (I'd add a 'suggestion' comment if GitHub would let me)

flowmachine/flowmachine/features/subscriber/topup_balance.py
```diff
@@ -12,15 +12,14 @@
 
 from ..utilities.sets import EventsTablesUnion
 from .metaclasses import SubscriberFeature
-from flowmachine.utils import standardise_date
+from flowmachine.utils import standardise_date, get_stat
 
 valid_stats = {
```
jc-harrison, line 17: Again, good to update the docstring to reflect that 'median' is not a valid stat here.

flowmachine/flowmachine/utils.py
```diff
@@ -17,6 +17,31 @@
 from typing import Union, Tuple
 
 
+def get_stat(stat: str, column_name: str):
```
jc-harrison, line 20: Is it worth sticking a `valid_stats` set alongside this and importing it into all the queries that use `get_stat`, rather than duplicating the set of `valid_stats` everywhere?

flowmachine/flowmachine/features/subscriber/handset_stats.py
jc-harrison: Hang on, no, I think median _is_ supported here. `_make_query` is written in an exceedingly convoluted way, but I think the final `return sql` (only reached if all the `if self.statistic in ...` are `False`) is the weighted median. The same is true in `TopupBalance`.

Docstring could still do with an update, though - it lists 'mode', which is not supported here (although there's no reason it couldn't be; it's implemented in `TopupBalance`).

flowmachine/tests/test_subscriber_handset_stats.py
```diff
@@ -19,7 +19,6 @@
         ("width", "avg", "JZoaw2jzvK2QMKYX", 28.0178288001543),
         ("width", "max", "DELmRj9Vvl346G50", 45.9),
         ("width", "min", "9vXy462Ej8V1kpWl", 5.19),
-        ("width", "median", "KXVqP6JyVDGzQa3b", 17.1),
```
jc-harrison, line 22: As noted above, I think this actually is valid (it was the presence of this test, and the fact it hasn't been failing, that made me go back and look at the query implementation more closely).

flowmachine/tests/test_subscriber_topup_balance.py
```diff
@@ -19,7 +19,6 @@
         ("avg", "JZoaw2jzvK2QMKYX", 224.419_246),
         ("max", "DELmRj9Vvl346G50", 529.73),
         ("min", "9vXy462Ej8V1kpWl", 442),
-        ("median", "KXVqP6JyVDGzQa3b", 435.94),
```
jc-harrison, line 22: As above

jc-harrison (changes requested):
Looks good on the whole; just some minor points.

flowmachine/flowmachine/features/subscriber/handset_stats.py
greenape: Oh good lord you're right. Well spotted. I'll reinstate and add a comment to make that more obvious I think.

flowmachine/flowmachine/utils.py
greenape: I was actually half wondering about making it an enum or something, but felt that might be overkill.

jc-harrison: I think it might be worth making it an enum. Particularly as several of the API-exposed query schemas use a `Statistic` field with a hard-coded list of valid stats, so it would be helpful to enforce consistency across the underlying flowmachine objects.

flowmachine/flowmachine/features/subscriber/handset_stats.py
```diff
@@ -157,7 +154,8 @@ def _make_query(self):
             """
             return sql
 
-        sql = f"""
+        # Weighted median
+        return f"""
```
jc-harrison, line 158: Might be safer to stick this in a `if self.statistic in {"median"}`. As things stand, if we add a new entry to the `Statistic` enum then `HandsetStats` will accept the new stat without complaint, but will calculate the median.

flowmachine/flowmachine/features/subscriber/displacement.py
```diff
@@ -159,7 +150,7 @@ def _make_query(self):
         sql = f"""
         SELECT 
             subscriber,
-            {self.statistic}(COALESCE(value_dist, 0) * {multiplier}) as value
+            {self.statistic:{f'COALESCE(value_dist, 0) * {multiplier}'}} as value
```
jc-harrison, line 153:
```suggestion
            {self.statistic:COALESCE(value_dist, 0) * {multiplier}} as value
```

jc-harrison (commented):
A few f-strings that I think can be simplified a bit, and another case where I think we don't want to default to calculating weighted median. Otherwise this looks good.

PR #5172, opened 2022-05-24
Remove Model

### LINKED ISSUE #5168: Remove Model
We should, I think, remove `Model` and `ModelResult` from flowmachine.

They're a neat trick, but in practice have not been used. We would be better off dealing with 'I want to run a python thing and put it in the db' through a more considered route (e.g. just run the damn python in the db). They also introduce a bunch of 'clever' code that is a maintenance headache.

Would also make unpicking #5140 a little easier.

greenape:
Closes #5168 

### I have:

### Description

Removes `Model`, `ModelResult` and the only actual model, `Louvain`.


flowmachine/flowmachine/core/cache.py
```diff
@@ -88,14 +89,54 @@ def column_names(self):
...
+        try:
+            ddl_op_result = connection.execute(ddl_op)
+        except Exception as e:
+            logger.error(f"Error executing SQL: '{ddl_op}'. Error was {e}")
+            raise e
+        try:
+            ddl_op_results.append(ddl_op_result.fetchall())
+        except ResourceClosedError:
+            pass  # Nothing to do here
+        for ddl_op_result in ddl_op_results:
```
jc-harrison, line 123: I haven't looked at this in detail before, and I'm not sure I understand what it's doing. We're looping over all DDL ops and appending the results to `ddl_op_results`, but within the loop we're looping over `ddl_op_results` and adding the execution time s to the overall `plan_time`. So execution times of earlier ops will be counted multiple times, if there are multiple ops?

jc-harrison (commented):
I'm happy with the removal of Model. Seems a shame to remove the Louvain stuff, but it's not worth maintaining a very different execution model just for that.

So my only concern is one that's only peripherally related to this PR (see comment).

flowmachine/flowmachine/core/cache.py
greenape: Good catch, I've screwed up an indent while moving it out to a function I think.

PR #5189, opened 2022-05-31
Make flowmachine table cache owner

### LINKED ISSUE #4714: flowmachine should own all cache tables
The flowmachine user in Flowdb needs to be the owner of all cache tables in cache schema, otherwise the flowmachine server will be unable to properly clean the cache and error out in trying to do so. Should either explicitly set the owner of the table to flowmachine user, or, there should be a trigger set in Flowdb to do this.

greenape:
Closes #4714

### I have:

### Description

1. Makes the `flowmachine` user always named `flowmachine` (you can create other users, but there shall always be a `flowmachine` user
2. Adds, as a post action on writing to cache, an ownership change of the newly created table to `flowmachine`.


flowdb/bin/build/9000_last_create_roles.sh
```diff
@@ -199,7 +193,7 @@ psql --dbname="$POSTGRES_DB" -c "
           obj record;
```
jc-harrison, line 193: Presumably this trigger is no longer required?

jc-harrison (commented):
Can also remove the `FLOWMACHINE_FLOWDB_USER` env from 'development_environment', and the secret from `secrets-quickstart.sh`.

flowdb/bin/build/9000_last_create_roles.sh
greenape: I've kept it in to allow for the possibility of tables getting created thru routes other than flowmachine that we'd still want to be managed. That said, I'm not sure there's _really_ an eventuality where that arises and they're also in the cache table.

greenape:
> Can also remove the `FLOWMACHINE_FLOWDB_USER` env from 'development_environment', and the secret from `secrets-quickstart.sh`.

Need to keep these, because _flowmachine_ still has a configurable username, it just happens to be `flowmachine` in these examples. Perhaps we should mandate it to `flowmachine` in container though?

flowdb/bin/build/9000_last_create_roles.sh
jc-harrison: I'm not sure we would necessarily want to change ownership of tables created through routes other than flowmachine, since the ownership change is specifically so that all flowmachine instances can access flowmachine-generated cache tables. Without an obvious situation where this might occur, I think it's just as plausible that we'd want the opposite - that tables created through some unforeseen route should remain owned by the user that created them.

jc-harrison:
> > Can also remove the `FLOWMACHINE_FLOWDB_USER` env from 'development_environment', and the secret from `secrets-quickstart.sh`.
> 
> Need to keep these, because _flowmachine_ still has a configurable username, it just happens to be `flowmachine` in these examples. Perhaps we should mandate it to `flowmachine` in container though?

Ah, I see. I suppose the question is: would we ever want the flowmachine container to use a postgres user other than `flowmachine` (perhaps, if individual users wanted to spin up their own flowmachine server and API?). In that situation, the flowmachine service would need a different `FLOWMACHINE_FLOWDB_PASSWORD` as well.

But for a standard deployment I think we'd always want the flowmachine container to use the `flowmachine` user. Maybe we could set `flowmachine` as the default in the container but allow overriding via the env?

flowdb/bin/build/9000_last_create_roles.sh
greenape: Yeah, I agree.

jc-harrison:
I've been thinking about the exception-catching business, and whether catching `BaseException` is the right thing to do here. I wonder whether it's more appropriate to `cancel` a query on `KeyboardInterrupt` than to `error` it, given that an `ERRORED` query indicates something went wrong when executing the query and it therefore can't be re-run without a manual reset, whereas `CANCELLED` indicates query execution was stopped but not due to the query itself, and the query can be re-run without manual intervention. So perhaps:

```
try:
    <run query>
except Exception as exc:
    q_state_machine.raise_error()
    raise exc
finally:
    if not q_state_machine.is_finished_executing:
        q_state_machine.cancel()
```

(i.e. if the query is still in a blocking state when it should have finished, but no `Exception` was raised, cancel it).

jc-harrison (approved):
:+1:

PR #5228, opened 2022-06-16
No more threadlocal

### LINKED ISSUE #5140: Deprecated thread local strategy
We're stuck on an old version of pandas when using flowmachine, because new pandas uses new sqlalchemy which deprecated threadlocal strategy.

We assume threadlocal throughout flowmachine's codebase at present in a way which is quite challenging to unpick, because one of the side effects of it is that one can be in a transaction in a very concealed way. For example, everything in `write_query_to_cache` is in a single transaction, but this is not immediately obvious.

We're quite inconsistent about when we pass in a connection (and what a connection is when passed), and when we summon one from context using `get_db()`, which also makes this a bit difficult to resolve.

greenape:
Closes #5140

### I have:

### Description

Turns off threadlocal strategy, and deals with the consequences.


flowmachine/flowmachine/core/cache.py
```diff
@@ -226,7 +227,9 @@ def write_query_to_cache(
 
 
 def write_cache_metadata(
-    connection: "Connection", query: "Query", compute_time: Optional[float] = None
+    connection: sqlalchemy.engine.Connection,
```
greenape, line 230: Should be a transaction object, need to find the right type

flowmachine/flowmachine/core/cache.py
```diff
@@ -261,14 +265,14 @@ def write_cache_metadata(
                 logger.debug(f"Can't pickle ({e}), attempting to cache anyway.")
                 pass
 
-        with con.begin():
+        with connection.begin():
```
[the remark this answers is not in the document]
greenape: It actually is, but better to rewrite it with the `as transaction` and then use that instead of the implicit version I think.

Thingus (commented):
I'm concerned that the the transaction strategy will fall foul of this: https://docs.sqlalchemy.org/en/14/core/connections.html#connections-avoid-nesting.
Fwiw, I like the transaction nesting pattern and think it's the Right Way to do things here (from a Python perspective, at least - there could be SQL problems that I'm unaware of). Wondering if either 
a) keep for now and put in an issue tracker, then deal with when SQLlchemy 2.0 depreciates it 
or 
b)  do the following (as recommended in the link):

```python
import contextlib

@contextlib.contextmanager
def transaction(connection):
    if not connection.in_transaction():
        with connection.begin():
            yield connection
    else:
        yield connection
```

then replace each `with con.begin()` with `transaction(con)`

greenape:
Where I've found them, I've attempted to refactor out any nesting of transactions - if you spot them, comment on them so we can get rid I think.

flowmachine/flowmachine/core/query.py
```diff
@@ -819,13 +819,13 @@ def invalidate_db_cache(self, name=None, schema=None, cascade=True, drop=True):
...
-                with con.begin():
-                    con.execute(
+                with con.begin() as trans:
+                    trans.execute(
                         "DELETE FROM cache.cached WHERE query_id=%s", (self.query_id,)
                     )
                     log("Deleted cache record.")
                     if drop:
-                        con.execute(
+                        trans.execute(
```
Thingus, line 828: Nested trans here

flowmachine/flowmachine/core/query.py
```diff
@@ -165,8 +165,8 @@ def __format__(self, fmt=""):
     def __iter__(self):
         con = get_db().engine
         qur = self.get_query()
-        with con.begin():
-            self._query_object = con.execute(qur)
+        with con.begin() as trans:
+            self._query_object = trans.execute(qur)
```
Thingus, line 169: This was my main concern for nested transactions, but it could be I've not read it through right - now I think about it, `__iter__` will only be called on top-level queries, and not by sub-queries?

Thingus:
> Where I've found them, I've attempted to refactor out any nesting of transactions - if you spot them, comment on them so we can get rid I think.

I gave it a good look, but I can't see any obvious ones - thought that some might be hiding in he exception handling, but it looks fine.

Thingus (approved):
Can't find any nested transactions, looks good :+1:

PR #5337, opened 2022-08-12
Upgrade versioneer

### LINKED ISSUE #5336: Setuptools-versioneer incompatibility
Setuptools has recently introduced PEP-660 editable installs, and in doing so has broken versioneer (https://github.com/pypa/setuptools/issues/3501, https://github.com/python-versioneer/python-versioneer/issues/314).

Looks like we can work around the issue in the immediate term with [this suggested patch](https://github.com/pypa/setuptools/issues/3501#issuecomment-1212476025):
```
diff --git i/versioneer.py w/versioneer.py
index 1040c21..5f57a8b 100644
--- i/versioneer.py
+++ w/versioneer.py
@@ -1559,6 +1559,8 @@ def get_cmdclass(cmdclass=None):
             cfg = get_config_from_root(root)
             versions = get_versions()
             _build_py.run(self)
+            if self.editable_mode:
+                return
             # now locate _version.py in the new build/ directory and replace
             # it with an updated value
             if cfg.versionfile_build:
```

But longer-term we should look out for an official patch from versioneer, and/or a compatibility update from setuptools.

jc-harrison:
Closes #5336

### I have:

### Description

Upgrades versioneer to 0.23.

Also switches to quiet (`-q`) pip install in the examples Dockerfile because we appear to be encountering this issue: https://github.com/pypa/pip/issues/10995 (I haven't investigated why the issue is arising here, but it's related to the Rich progress bar used by pip, which isn't used in a quiet install).


greenape (approved):

Thingus (approved):
:+1:

PR #5464, opened 2022-09-20
Bump deps

jc-harrison:
### I have:

### Description

- Re-lock all Pipfiles
- Bump Apache Airflow to 2.3.4
- Downgrade FlowETL base airflow image to 2.3.4 (for consistency with flowetl Pipfile)
- Configure dependabot to ignore airflow version in Dockerfile


jc-harrison:
Currently hitting a number of issues, none of which I've managed to diagnose yet:
- In flowetl integration tests, airflow is complaining about needing to run 'airflow db upgrade' (on what should be an empty or newly-initialised db). I initially thought this was related to psycopg2 issues that arose in airflow 2.4.0 (apache-airflow-providers-postgres 5.2.1), which is why I've only bumped airflow to 2.3.4, but the issue remains so perhaps once it's resolved we'll be able to upgrade to airflow 2.4.0 after all.
- When building the docs, example notebooks are complaining that `TOKEN` is undefined - implies the `notebook_preamble.py` script is not running.
- FlowAPI is returning a 500 error instead of the API spec

PR #5520, opened 2022-10-17
Fix deadlock

greenape:
Fixes a possible deadlock where the connection pool is smaller than the number of queries to be run.

### DIFF
--- a/flowmachine/flowmachine/core/cache.py
+++ b/flowmachine/flowmachine/core/cache.py
@@ -209,7 +209,7 @@ def write_query_to_cache(
                             trans,
                             query,
                             compute_time=plan_time,
-                            executed_sql="\n".join(query_ddl_ops),
+                            executed_sql=";\n".join(query_ddl_ops),
                         )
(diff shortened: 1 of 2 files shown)


flowmachine/flowmachine/core/cache.py
```diff
@@ -205,7 +205,12 @@ def write_query_to_cache(
                     raise exc
                 if schema == "cache":
                     try:
-                        write_cache_metadata(trans, query, compute_time=plan_time)
+                        write_cache_metadata(
+                            trans,
+                            query,
+                            compute_time=plan_time,
+                            executed_sql="\n".join(query_ddl_ops),
```
jc-harrison, line 212:
```suggestion
                            executed_sql=";\n".join(query_ddl_ops),
```

jc-harrison (approved):
I think we need a semicolon to separate the DDL ops, but otherwise this looks good. I'm happy to merge this without a regression test, because I think it's a good change even aside from the deadlock issue.

PR #5580, opened 2022-11-04
Rebuild rolescopepicker

Thingus:
### I have:

- [ ] Brought the branch up to date with ~master~ permission_rework

### Description

This re-implements `RoleScopePicker`, reimplementing `CascadePicker` within Flowauth. There is also a Cypress component test.

### DIFF
--- a/flowauth/frontend/src/RoleScopePicker.cy.jsx
+++ b/flowauth/frontend/src/RoleScopePicker.cy.jsx
@@ -1,102 +1,110 @@
-import RoleScopePicker from './RoleScopePicker'
+/* eslint-disable no-undef */
+/* eslint-disable react/react-in-jsx-scope */
+import RoleScopePicker from "./RoleScopePicker"
 
 //Need a dummy set of server scopes
 //and a dummy set of existing role scopes
 //
 
 
-describe('<RoleScopePicker>', () => {
-	it('mounts', () => {
-		cy.intercept({
-				method: 'GET',
-				url: '/admin/servers/1/scopes', 
-			},
-			[
-				{
-					"enabled": true,
-					"id": 1,
-					"name": "admin0:dummy_query:dummy_nested_query"
-				},
-				{
-					"enabled": true,
-					"id": 3,
-					"name": "admin0:dummy_query:inner_dummy_query"
-				},
-				{
-					"enabled": true,
-					"id": 5,
-					"name": "get_results"
-				}
-			]
-		).as('getServerScopes');
+describe("<RoleScopePicker>", () => {
+  it("mounts", () => {
+    cy.intercept({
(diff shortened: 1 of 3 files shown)


flowauth/frontend/src/RoleScopePicker.jsx
```diff
@@ -2,109 +2,219 @@
...
-
-  return (
-    <RightsCascade
-      options={rightsChoices}
-      value={selectedRights}
-      onChange={handleChange}
-      // label={"Scopes"}
-    />
-  );
+    // This mess takes every scope that is
```
greenape, line 131: Easier to just return that form from the backend?

Thingus: It needs both fully-resolved scope paths and the grouped rep; it'd need to be mangling whatever's returned in one direction or the other.

flowauth/frontend/src/RoleScopePicker.cy.jsx
```diff
@@ -1,102 +1,140 @@
...
+      .find("[data-cy=nested-dummy_query]")
+      .find("[data-cy=checkbox]").first()
+      .invoke("prop", "checked", false)
+    cy.get("[data-cy=nested-admin0]")
+      .find("[data-cy=nested-other_query]")
+      .find("[data-cy=checkbox]").first()
+      .invoke("prop", "checked", false)
+  })
 
+})
```
flowstef, line 140: No newline?! 😱
I thought the code formatter takes care of these things?

flowauth/frontend/src/RoleScopePicker.jsx
```diff
@@ -2,109 +2,227 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+/* eslint-disable react/prop-types */
```
flowstef, line 5: What happens if we don't disable this?
If it really makes no sense perhaps we can disable in eslint's config globally?

flowstef (approved):
Nothing horribly wrong jumps out at me.

flowauth/frontend/src/RoleScopePicker.cy.jsx
Thingus: Knew I'd forgotten to do something...

flowauth/frontend/src/RoleScopePicker.jsx
Thingus: It's something we _should_ be doing (according to eslint), but haven't - I'd like to put it in in future, but I think it adds a dependency.

greenape (approved):
I'm happy to merge as well I think.

PR #5716, opened 2022-12-21
Bump python deps

### LINKED ISSUE #5704: Pin pglast version in setup.py
I think we should really pin the pglast version, because it is used for query id generation in some cases, and means that an upgrade to pglast could change cache ids unexpectedly.

jc-harrison:
Closes #5704

### I have:

### Description

- Re-locks all Pipfiles
- Upgrades FlowETL Airflow base image to match Airflow version in Pipfile
- Tells dependabot to stop upgrading Airflow - the version in the Pipfile needs to be consistent with the docker image version, and I don't know of any way to tell dependabot to respect that constraint
- Pins `pglast` in flowmachine's `setup.py`, as different `pglast` versions may result in different flowmachine query IDs (at least for `CustomQuery`).


greenape (approved):
👍👍

Issue #5721, opened 2022-12-23
Allow setting minimum total weight in MajorityLocation


jc-harrison:
The `MajorityLocation` query class is sufficiently general to accept any query with subscriber, location and weight columns as the `subscriber_location_weights` parameter, but the original use-case for this query kind is to assign to each subscriber the location that was their "daily home location" on the majority of days in a specified period (i.e. where the `subscriber_location_weights` parameter is a `LocationVisits` query that counts the number of times each location was a subscriber's "home" in a sequence of home location queries).

In this situation, `MajorityLocation` only works as intended if all subscribers are active (i.e. have a "daily home location") on each of the days. E.g. if we're looking at a 7-day period, but a subscriber only has a daily location on 3 of those days, then the threshold for that subscriber to achieve a "majority" would be just 2 days at the same location, not 4 days (which would be required for a majority over the entire 7-day period).

One solution would be to set the "majority" threshold using the length of the sequence of daily locations instead of the summed weight per subscriber, but this would make the query class less general  (this approach would not work if one used a different kind of query for the `subscriber_location_weights`, e.g. `CallDays` or `PerLocationSubscriberCallDurations`).

An alternative would be to add a `minimum_total_weight` parameter to `MajorityLocation` - then the threshold for assigning a location as the "majority" would be for that location's weight to be greater than `max(summed_weight, minimum_total_weight)/2`, rather than `summed_weight/2`.

PR #5722, opened 2022-12-23
Changes to MajorityLocation

### LINKED ISSUE #5720: Decompose MajorityLocation to improve cache re-use
If a user wants to run a `MajorityLocation` query excluding unlocatable subscribers, and also the same `MajorityLocation` query including unlocatable subscribers, these would be two independent queries. But the second case actually requires running the first query, and then joining this to the set of all subscribers with weighted locations. I think it would be beneficial to break down `MajorityLocation` a little so that a query with `include_unlocatable=True` uses the corresponding query with `include_unlocatable=False` as a sub-query, to improve cache re-use in this situation.

_Originally posted by @jc-harrison in https://github.com/Flowminder/FlowKit/issues/5128#issuecomment-1121384239_

jc-harrison:
Closes #5720 , #5721

### I have:

### Description

- Adds a `minimum_total_weight` parameter to `MajorityLocation` (and to the exposed `majority_location`)
- Removes the `include_unlocatable` parameter from `MajorityLocation`, and instead introduces a `MajorityLocationWithUnlocatable` query class which takes a `MajorityLocation` sub-query and augments its result with rows for all the subscribers in the `subscriber_location_weights` query that have no majority location
- Adds a `majority_location` function with the same arguments as the previous `MajorityLocation` constructor, which returns either a `MajorityLocation` or `MajorityLocationWithUnlocatable` query object


flowmachine/flowmachine/features/subscriber/majority_location.py
```diff
@@ -63,26 +72,113 @@ def column_names(self) -> List[str]:
...
+    Subscribers for whom there is no single location with an outright majority will either be excluded from the
+    query result (if `include_unlocatable==False`), or included in the result with `NULL` value in the location ID
+    column(s) (if `include_unlocatable==True`).
+
+    Parameters
+    ----------
+    subscriber_location_weights : Query
+        The query object containing subscribers, locations, and weights.
+    weight_column : str
+        The column, when summed, that will produce the count used to threshold the majority
```
Thingus, line 152:
```suggestion
        The column in `subscriber_location_weights`, when summed, that will produce the count used to threshold the majority
```

Thingus (approved):
Looks good to me, also happy to see the xfail gone.#
Only got the one docstring comment, but it's a minor one

PR #5799, opened 2023-01-24
Flask migrate for Flowauth

Thingus:
### I have:

### Description

This PR introduces Alembic (via `flask_migrate`) migrations to Flowauth, and includes a migration from v1.17.0 to v1.18.0.
It also includes an automated test environment for testing migrations without touching the main db.


Thingus:
A note on transactions: Alembic _does_ try to do them, except when it doesn't because of the backend. https://docs.python.org/2/library/sqlite3.html#controlling-transactions

flowauth/backend/migrations/env.py
```diff
@@ -0,0 +1,102 @@
...
+fileConfig(config.config_file_name)
+logger = logging.getLogger("alembic.env")
+
+
+def get_engine():
+    try:
+        # this works with Flask-SQLAlchemy<3 and Alchemical
+        return current_app.extensions["migrate"].db.get_engine()
+    except TypeError:
+        # this works with Flask-SQLAlchemy>=3
```
greenape, line 25: Do we need to support both?

flowauth/backend/migrations/env.py
```diff
@@ -0,0 +1,102 @@
...
+from alembic import context
+
+# this is the Alembic Config object, which provides
+# access to the values within the .ini file in use.
+config = context.config
+
+# Interpret the config file for Python logging.
+# This line sets up loggers basically.
+fileConfig(config.config_file_name)
+logger = logging.getLogger("alembic.env")
```
greenape, line 17: Better to use structlog as elsewhere I think

flowauth/backend/migrations/env.py
```diff
@@ -0,0 +1,102 @@
...
+            target_metadata=get_metadata(),
+            process_revision_directives=process_revision_directives,
+            **current_app.extensions["migrate"].configure_args
+        )
+
+        with context.begin_transaction():
+            context.run_migrations()
+
+
+if context.is_offline_mode():
```
greenape, line 99: So importing this file is sufficient to trigger a migration?

flowauth/backend/migrations/migration_tests/test_migration.py
```diff
@@ -0,0 +1,157 @@
...
+import sys
+from pathlib import Path
+
+import flask_migrate
+import git
+from git.util import stream_copy
+import pytest
+import sqlite3
+
+# from alembic import
```
greenape, line 13:
```suggestion
```

flowauth/backend/migrations/env.py
Thingus: Not that I've seen - that's done by either `$ flask db upgrade` or `>>> migration.upgrade`

Thingus: OK, on an investigation if you import `env` normally, it falls over at `alembic.context` (presumably because if you don't call it from an alembic-derived script, there is no `context` to work with)

flowauth/backend/migrations/env.py
Thingus: Probably not, but this is an auto-generated script - I'm inclined to leave it alone unless we need to change it.

flowauth/backend/migrations/env.py
Thingus: Done

flowauth/backend/flowauth/models.py
```diff
@@ -567,12 +569,11 @@ def init_db(force: bool = False) -> None:
         db.drop_all()
     db.create_all()
     current_app.config["DB_IS_SET_UP"].set()
+    # Since this is a fresh db, set the revision inside alembic to HEAD
+    flask_migrate.stamp(revision="head")
```
jc-harrison, line 573: Will this be executed if the db already exists (from previous deployment)? (Or if not, does it matter?) Not sure I've fully got my head around how Alembic determines revisions.

jc-harrison (approved):
I haven't tried running it, but it all looks correct to me, so as long as the test is working I think I'm happy.

Would be good to have a short summary somewhere of the steps required to do a migration, for future reference.

flowauth/backend/flowauth/models.py
Thingus: No, this _should_ be skipped if the db already exists - in that case, there should either be an existing revision number or we're starting from revision = None. In the case of a fresh db (if I've understood Alembic correctly), we want to set it's revision to `head`.

PR #5897, opened 2023-02-17
Look for COUNT instead of just first row in data_present_sensor

### LINKED ISSUE #5090: FlowETL DataPresentSensor trips over null fields
The FlowETL `DataPresentSensor` operator checks whether a file contains any data by attempting to select one row from the foreign data wrapper around the file. `DataPresentSensor` is an Airflow `SqlSensor` operator, whose success criterion is that the _first cell_ in the returned row is not in `(0, '0', '', None)`. This means that if the first field in the first row of the file is null (or an empty string), the `DataPresentSensor` operator will fail to detect data, so ingestion will not proceed.

Given that it's not necessarily the case that a null value in the first column (whatever that may be) invalidates the entire contents of the file, we should find a more robust way of handling this check.

Thingus:
Closes #5090 

### I have:

### Description
Instead of `SELECT` ing the first row in `staging_table` in `data_present_sensor`, this PR uses COUNT to check for data presence. This would still fail if no records are present (COUNT (*) would return 0 in the first cell), but would work if there's any data present in the first 100 records (arbitarily selected)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Fixed
+
+- Null values in first column of first row of ingested data no longer cause flowetl to skip ingestion [#5090](https://github.com/Flowminder/FlowKit/issues/5090)
+
 ## [1.18.2]
 
 ## Fixed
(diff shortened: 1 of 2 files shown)


jc-harrison (commented):
Would be more efficient to do "select exists (select * from {{staging_table}} limit 1)"

greenape (approved):

greenape:
Approved - add a changelog before sticking ready to merge label on

PR #5907, opened 2023-02-23
Allow setting max locks, make default much higher than normal pg

### LINKED ISSUE #5157: Review 'max_locks_per_transaction' setting in FlowDB
Postgres stores table locks in memory, with pre-allocated space for `max_locks_per_transaction * max_connections` locks. Attempting to take more locks than this at any one time will result in an out-of-memory error. In the FlowDB defaults we set `max_connections=200`, and we're not explicitly setting `max_locks_per_transaction` so that defaults to 64.

Queries that select from partitioned tables require an `AccessShareLock` on each partition, as well as on every index on each of those partitions (this is true even if the query only requires data from one partition - postgres needs to obtain a lock before checking constraints to assess whether a partition is required). The events tables have a partition per day, and typically a number of indexes on each partition, so this can result in a large number of locks required for each query.

E.g. if the `events.calls` table contains 2 years of data, with 3 indexes (on `datetime`, `msisdn` and `location_id`),  then an `EventTableSubset` query such as:
```
SELECT datetime, location_id, msisdn
FROM events.calls 
WHERE events.calls.datetime >= '2022-02-10 00:00:00' AND events.calls.datetime < '2022-02-11 00:00:00'
```
would require `365*2*(3+1)=2920` 'AccessShareLock' table locks. With the current default settings, just 5 such transactions (of a possible 200) would be sufficient to exhaust the memory allocated for `200*64=12800` locks.

From this I think it's clear that the default `max_locks_per_transaction=64` is insufficient for our heavy use of partitions, and we need to increase it. A suitable value would depend on the maximum span of data we expect to keep in the DB at any given time, along with the typical number of indexes we create on the events tables.

At the same time it's worth considering whether this is the right solution. As far as I can tell there's no major reason it would be a bad idea to allow significantly more table locks per transaction, but it's possible this issue is a sign we should be approaching something a bit differently.

greenape:
Closes #5157

### I have:

### Description

Adds a new env var that can control Postgres' max_locks_per_transaction setting to the configure script, and increases the default value.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1202306260479520

### DIFF
--- a/flowdb/bin/build/configurate.py
+++ b/flowdb/bin/build/configurate.py
@@ -69,6 +69,8 @@ use_jit = "off" if bool_env("NO_USE_JIT") else "on"
 stats_target = int(
     os.getenv("STATS_TARGET", 10000)
 )  # Default to higher than pg default
+max_locks = int(os.getenv("MAX_LOCKS", 365 * 5 * 4 * (1 + 4)))
+
 
 config_path = os.getenv(
     "AUTO_CONFIG_PATH", "/var/lib/postgresql/data/postgresql.configurator.conf"
@@ -89,6 +91,7 @@ with open("/docker-entrypoint-initdb.d/pg_config_template.conf") as fin:
         gendate=datetime.datetime.now().strftime("%Y-%m-%d %H:%M:%S"),
         stats_target=stats_target,
         use_jit=use_jit,
+        max_locks=max_locks,
     )
 
 print("Writing config file to", config_path)
--- a/flowdb/bin/build/configurate.py
+++ b/flowdb/bin/build/configurate.py
@@ -69,7 +69,7 @@ use_jit = "off" if bool_env("NO_USE_JIT") else "on"
 stats_target = int(
     os.getenv("STATS_TARGET", 10000)
 )  # Default to higher than pg default
-max_locks = int(os.getenv("MAX_LOCKS", 365 * 5 * 4 * (1 + 4)))
+max_locks = int(os.getenv("MAX_LOCKS_PER_TRANSACTION", 365 * 5 * 4 * (1 + 4)))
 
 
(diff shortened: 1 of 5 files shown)


flowdb/bin/build/configurate.py
```diff
@@ -69,6 +69,8 @@ def bool_env(var):
 stats_target = int(
     os.getenv("STATS_TARGET", 10000)
 )  # Default to higher than pg default
+max_locks = int(os.getenv("MAX_LOCKS", 365 * 5 * 4 * (1 + 4)))
```
Thingus, line 72: For future reference, what's the logic for the default here?

jc-harrison: https://github.com/Flowminder/FlowKit/issues/5157#issuecomment-1131735184

Thingus (approved):
:+1:

jc-harrison (commented):
A couple of things:

I'm not sure where's best to put this info (perhaps in the secrets quickstart; probably also somewhere in the docs), but I think it's important to note (given the non-negligible memory requirement for this many locks) that when running flowdb in a docker container, the size of the /dev/shm tmpfs mount needs to be larger than
```
SHARED_BUFFERS_SIZE + MAX_LOCKS * (max_connections + max_prepared_transactions) * (sizeof(LOCK) + sizeof(LOCKTAG))
```
(otherwise postgres may attempt to allocate more shared memory than is available).

Also, I think I'd prefer to call the env variable `MAX_LOCKS_PER_TRANSACTION` rather than `MAX_LOCKS` - then it's consistent with the name of the postgres config parameter, just to avoid any ambiguity about its purpose.

jc-harrison (approved):
👍

Issue #6006, opened 2023-03-29
Pinned dependencies getting out of sync between Pipfile and setup.py


jc-harrison:
**Describe the bug**
We are duplicating dependency constraints in the various FlowKit python packages (flowapi, flowauth, flowclient, flowkit-jwt-generator, flowmachine; no longer the case for flowetl since #5962) - constraints in `setup.py` and `Pipfile` must be manually kept in sync in each case.

This alone is an issue, since it runs the risk that if we update one but forget to update the other, the dependency versions used for testing will be different from those that would be installed via `pip install flow<package>`. Pipenv's [recommendation](https://pipenv-fork.readthedocs.io/en/latest/advanced.html#pipfile-vs-setup-py) when using both setup.py and Pipfile is to add the local package as an editable requirement in the Pipfile. This would be different from our current setup, though - currently a non-dev `pipenv install` will install all of the package dependencies but _not_ the package itself (editable or otherwise).

It becomes more of an issue, though, because dependabot automatically bumps pinned versions in the Pipfiles without changing the corresponding versions in setup.py. Because we editable-install the package in the dev dependencies, this inconsistency leads to an un-lockable env, but for whatever reason that doesn't stop dependabot from making the change (I think because dependabot doesn't correctly handle dependencies of local-path requirements), and it still produces a `pipenv install`-able Pipfile.lock. I don't _think_ this is a problem for the dev install - locked (upgraded) deps won't be overwritten by versions specified in setup.py, so unit tests will be correctly run using the dependabot-upgraded dependencies - but it _is_ a problem in the docker images (which therefore affects some of the integration tests). In the Dockerfiles we first do a non-dev pipenv install, and then run `pipenv run python setup.py install`, which will overwrite the pipenv-pinned versions with versions specified in setup.py. Similarly, non-containerised integration tests will use dependencies from the local-path installs of the packages in the integration tests env, which will pick up dependencies from each package's setup.py (not from that package's Pipfile). The result is that dependabot can bump version pins in a Pipfile, and integration tests will continue to use the older versions pinned in setup.py, so Pipfile pins can be "successfully" bumped to versions that would break the integration tests.

[...]

Issue #5867, opened 2023-02-10
flowmachine sqlachemy is unpinned


greenape:
This pr: https://github.com/Flowminder/FlowKit/pull/5864 updated the pin on sqlalchemy to 2.0.3, which should have triggered test failures because of the breaking changes made in sqlalchemy 2.

It didn't, because the version installed by pipenv is actually ignored because of the version specifier being set as <2 in setup.py. That means it gets replaced in the docker image when setup is run, replaced in the tests by the editable dev install of flowmachine, and is correctly pinned in the integration tests because the pin there comes from flowmachine's setup.py

jc-harrison:
I wonder whether the right thing to do here would be to stick `flowmachine = {editable = true, path = "."}` in the non-dev packages section of the Pipfile, and remove (from the Pipfile) all the packages that are specified in the setup.py requirements (which I suppose should be all of them, really) - then we're not duplicating requirements and version pins.

PR #6054, opened 2023-04-14
Don't overwrite pinned dependencies by re-installing local python packages

### LINKED ISSUE #6052: Flowmachine has incompatible sqlalchemy version
Flowmachine is incompatible with sqlalchemy >= 2.0, due to the direct use of strings in `transaction.execute()` (and quite possibly other reasons as well):
```
>>> flowmachine.core.context.get_db().has_table("calls", schema="events")
Traceback (most recent call last):
  File "/root/.local/share/virtualenvs/flowmachine-Lph78dnj/lib/python3.8/site-packages/sqlalchemy/engine/base.py", line 1410, in execute
    meth = statement._execute_on_connection
AttributeError: 'str' object has no attribute '_execute_on_connection'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/FlowKit-1.18.3/flowmachine/flowmachine/core/connection.py", line 174, in has_table
    return trans.execute(exists_query).fetchall()[0][0]
  File "/root/.local/share/virtualenvs/flowmachine-Lph78dnj/lib/python3.8/site-packages/sqlalchemy/engine/base.py", line 1412, in execute
    raise exc.ObjectNotExecutableError(statement) from err
sqlalchemy.exc.ObjectNotExecutableError: Not an executable object: "SELECT EXISTS(\n        SELECT * FROM information_schema.tables \n            WHERE table_name='calls'\n         AND table_schema='events')"
```

We added a sqlalchemy version constraint (https://github.com/Flowminder/FlowKit/commit/174b42e8da8c1879d559c032d124c1c7d73eced0) to work around this, but dependabot immediately bumped the version (https://github.com/Flowminder/FlowKit/commit/4f0f779410fd2fc7b6fd92b64de315d99911d023), and it appears our tests didn't catch this incompatibility. Hence the latest release of the flowmachine docker image (1.18.3) has an incompatible sqlalchemy version and is therefore broken.

jc-harrison:
Closes #6052, #5867

### I have:

### Description

This should help to prevent dependency bumps to incompatible versions slipping through CI without getting caught by the tests, by removing a few cases where we were installing FlowKit python packages _after_ running `pipenv install` - this could downgrade dependencies already installed from Pipenv.lock to different, un-pinned versions if the constraints in the relevant setup.py differed from those in the Pipfile (as can happen due to #6006).

- Removes unnecessary editable installs of flowmachine and flowclient in CI, which were overwriting pipenv-locked dependencies with unpinned versions
- Replaces `python setup.py install` with `pip install --no-deps .` in the FlowAPI and FlowMachine dockerfiles, also to avoid re-installing dependencies that should be pinned by pipenv
- Tells dependabot to ignore a couple of dependencies that will otherwise currently be bumped to incompatible versions without causing test failures
- Copies all constraints from flowmachine/setup.py into flowmachine/Pipfile (some were missing from the Pipfile)
- Bumps all python dependencies
- A few minor fixes for compatibility with pandas 2.0 (in flowclient, autoflow examples and flowmachine tests)

Note that this doesn't help if dependabot strips out the dependencies of one of our FlowKit packages from a Pipfile.lock (as we have seen happen due to dependabot's incorrect handling of relative-path Pipfile requirements in a multi-root repo) - in this case unpinned versions of the un-locked dependencies will still be installed by `pipenv install`.


jc-harrison:
Docs build and synth data tests are currently failing because the gadm server is down, so the flowdb-synthetic-data container won't run.

flowmachine/Pipfile
```diff
@@ -4,12 +4,12 @@ verify_ssl = true
 name = "pypi"
 
 [packages]
-sqlalchemy = "<3.0.0"
+sqlalchemy = "<2.0.0"
```
Thingus, line 7: Can you drop a comment in here explaining the pin?

flowmachine/Pipfile
```diff
@@ -19,8 +19,8 @@ structlog = "*"
 shapely = "*"
 finist = "*"
 python-rapidjson = "*"
-marshmallow = "*"
-marshmallow-oneofschema = "*"
+marshmallow = ">=3.0.0"
+marshmallow-oneofschema = ">=2.0.0"
```
Thingus, line 23: Ditto here

.circleci/config.yml
```diff
@@ -279,7 +279,7 @@ jobs:
           name: Install pipenv
           command: pip install --upgrade pip pipenv
       - restore_cache:
-          key: flowdb-deps-1-{{ checksum "flowdb/tests/Pipfile.lock"}}
+          key: flowdb-deps-6-{{ checksum "flowdb/tests/Pipfile.lock"}}
```
Thingus, line 282: Do we want this to be in lockstep with `flowmachine-deps`, on so we can bump them all at once if we need to with something like `s/deps-6-/deps-7-/g`?

Thingus (approved):
Looks alright to me, but I'd want Jono to give it a look

.circleci/config.yml
jc-harrison: Could do - but then we'd end up bumping all the caches any time one of them needs bumping, which seems excessive.

PR #6164, opened 2023-06-13
Location ids table and locatable events QA check

### LINKED ISSUE #5289: Locatable event count QA check
FlowETL includes a QA check that counts cell IDs for each day of data ingested, and a check that counts _locatable_ cell IDs (i.e. those that are present in the cells table). This is somewhat useful for detecting if the CDR contains a significant number of unrecognised cell IDs, but even with a large number of unrecognised cell IDs it's quite possible that these only appear in a small proportion of the CDR records. It would be useful to also have a QA check that counts locatable _events_ (i.e events corresponding to known cells).

### LINKED ISSUE #5376: Record first/last active date for cell IDs
Over time, new cells will be added to a mobile network and old cells removed. This means that the content of the `infrastructure.cells` table needs to be periodically updated to remain current.

Updating cell locations has a number of impacts:
- Cell locations may be provided for cells that were first active many months ago, meaning that previously-unlocatable CDR events from an indeterminate time ago may become locatable.
- Often, we may want to define cell clusters (or other cell groupings) based on only the cells that were active during a certain period. This information is not always readily available in `infrastructure.cells` (while we would hope to receive locations for all new cells that become active, we may not know when those cells were added, or that a previously-active cell has been removed from the network).
- Sometimes, it can be useful to identify short-lived cell IDs (those that only appear in the CDR over a short period) to help with quality-checking cell locations information.

For all of these reasons (and probably more), it would be useful to keep an up-to-date record of all cell IDs seen in the CDR (regardless of whether they correspond to known cells in `infrastructure.cells`), with the dates that each was first/last active. We could keep this information in an `events.cell_ids` table, and update this on CDR ingestion via a new task in FlowETL (I'd hope this update wouldn't be too slow, but even if it is I don't think it should need to block making the new CDR available for analysis).

jc-harrison:
Closes #5376
Closes #5289

### I have:

### Description

Should be two PRs really, but the changes are both small, and vaguely related...

- Adds a task in the FlowETL ingestion DAG to update a new table `events.location_ids`, which records the first and last date each location ID appeared in the CDR data (per event type).
- Adds a new QA check "count_locatable_events", which counts the number of added rows with location ID corresponding to a cell with known location (similar to the existing "count_locatable_location_ids", but counts the number of events at locatable cells rather than the number of distinct active locatable cells).

The `events.location_ids` table is created (if not exists) by the FlowETL task, rather than being defined in the FlowDB provisioning scripts. This is advantageous if we're updating FlowETL in an existing FlowKit deployment, since we don't have a formal migration process so would otherwise need to remember to manually create the table, but does mean the full specification of the FlowDB schema begins to spill out beyond the `flowdb` sub-repo, which is perhaps not the cleanest approach. I'm not sure whether this is the right decision.


greenape:
Maybe better to add it under the etl schema rather than events through this route? I'm also thinking if anywhere it belongs in the infrastructure schema. No objection in principle to the create occurring here, but shall we add the ddl for it to the flowdb schema as well?

jc-harrison:
> Maybe better to add it under the etl schema rather than events through this route? I'm also thinking if anywhere it belongs in the infrastructure schema. No objection in principle to the create occurring here, but shall we add the ddl for it to the flowdb schema as well?

My reasoning for putting the table in `events` rather than `infrastructure` is that it contains information specifically about the location IDs appearing in the CDR records, irrespective of whether they correspond to actual infrastructure (i.e. if we're debugging a discrepancy between CDR and cell info, this table will be providing the information from the events side).

Could put it in `etl`, but it doesn't really relate to the ETL, it's just populated through the ETL process. Although if it's not actually defined in the FlowDB schema and only created on-the-fly by FlowETL then arguably it should live in the `etl` schema.

I think you're right, we should add the DDL to the FlowDB schema. I'm not keen on having it in the FlowDB schema _and_ in the ETL task, because then there's a risk of the two getting out-of-sync. Perhaps it's best to have it in the FlowDB scripts only, and then if the task tries to run on a schema that's not updated it will fail. In lieu of a proper migration process, it's probably not actually very helpful to start scattering partial migration steps throughout the ETL.

greenape:
A follow on question would then be if we actually want a separate `events_metadata` schema maybe, but that doesn't feel required at this stage.

greenape (approved):
LGTM

PR #6258, opened 2023-08-09
Bumping airflow + hash to 2.6.3

Thingus:


jc-harrison:
You'll want to re-lock the FlowETL python dependencies as well, to match the ones in the upgraded base image:
- Change version to 2.6.2 in [.airflow-version](https://github.com/Flowminder/FlowKit/blob/bump-airflow/flowetl/.airflow-version)
- Run `make lock` in flowetl dir

Thingus:
> You'll want to re-lock the FlowETL python dependencies as well, to match the ones in the upgraded base image:
> 
>     * Change version to 2.6.3 in [.airflow-version](https://github.com/Flowminder/FlowKit/blob/bump-airflow/flowetl/.airflow-version)
> 
>     * Run `make lock` in flowetl dir

Thanks, that was causing me some pain

PR #6433, opened 2024-01-09
Cell info table

### LINKED ISSUE #6184: Changes to infrastructure.cells table
A few suggested additional columns for the `infrastructure.cells` table:
- `date_added` (might need a better name): date that the cell was added to the cells table (not necessarily the same as the date that the cell was added to the network)
- `date_modified`: Sometimes incorrect cell info may get corrected - it would be useful to be able to keep track of when this happens. Not sure whether we'd want to track changes to location separately from other changes.
- `additional_metadata`: json field to capture additional information (cell info files from MNOs often include additional fields that don't fit into the existing columns in the cells table, which would be useful to have available in FlowDB)
- `to_exclude` (or equivalently `to_include`): boolean field to record whether CDR events at this cell should be included in mobility analysis. We'd want to change flowmachine's `JoinToLocation` accordingly, to exclude these cells from the join. This would be useful when we suspect a cell location is incorrect, for example. It would be good to be able keep track of these excluded cells in the cells table (especially when updating the cell info), rather than needing to record them elsewhere.

Alternatively, instead of a `date_modified` column we could treat the cells table as append-only, and modify a cell record by adding a new row with incremented `version` and marking the old record as `to_exclude`. This would have the advantage of keeping a record of the history. We'd perhaps still want a `date_excluded` column, which would make it possible to reconstruct the contents of the cells table as it was on a previous date.

It would also be good to have an `EXCLUDE` constraint to ensure no two (non-excluded) rows with the same ID have overlapping `date_of_first_service`-`date_of_last_service` ranges. This would be facilitated by replacing the `date_of_first_service` and `date_of_last_service` fields with a single date range (or timestamp range) field.

jc-harrison:
Closes #6184

### I have:

### Description

Creates a new table `infrastructure.cell_info`, designed to contain not only the current set of cell location/metadata records but also all previous versions and any cells that we have excluded due to data quality issues.

Differences from `infrastructure.cells` are:
- I've renamed a few fields to make their intention clearer (at least to me)
- Separate 'longitude' and 'latitude' fields instead of 'geom_point' (purely because I'm thinking of this table as being closer to the source cell info files than `infrastructure.cells` - no strong reason we couldn't have 'geom_point' instead / as well)
- 'date_of_first_service' and 'date_of_last_service' are combined into a 'dates_of_service' `TSTZRANGE` field, which allows us to add an exclude constraint to ensure each CDR record (i.e. each (cell_id, timestamp)) can never map to more than one valid cell
- 'to_include' boolean field - cells with `to_include = FALSE` should not be used in analysis, which allows us to keep records of excluded cells (which can help, for example, to ensure we don't re-ingest previously-excluded cells). If any cell locations or metadata change from one cell info update file to the next, the intention would be to mark the old cell as excluded, and add a new row (with same cell ID and incremented version) with the new location/metadata.
- 'first_received'/'last_received' fields to record when each cell was first/last included in files provided by the MNO
- 'ingested_at' field (should be set as the time at which the record was added to the table) and 'excluded_at' field (should be set as the time at which 'to_include' was first set to False) - these fields allow us to extract the set of cells that were "included" (and would have been used in queries) at any previous point in time
- 'notes' field, for adding helpful human-readable notes about a cell record (e.g. reason for exclusion)
- 'additional_metadata' JSON field, for catching any fields provided in the files form the MNO that don't fit into the structure of `infrastructure.cells`
- (cell_id, version) are set as the primary key, instead of an additional internally-generated auto-incrementing ID field

When ingesting new cell info, the intention would be to update `infrastructure.cell_info` first, and then re-populate `infrastructure.cells` from `infrastructure.cell_info`. I'll add a FlowETL workflow for ingesting new cell info in a later PR. `infrastructure.cells` can be re-populated like this:
```
DELETE FROM infrastructure.cells;

INSERT INTO infrastructure.cells ( id,
                                   version,
                                   site_id,
                                   name,
                                   type,
                                   msc,
                                   bsc_rnc,
                                   antenna_type,
                                   status,
                                   lac,
                                   height,
                                   azimuth,
                                   transmitter,
                                   max_range,
                                   min_range,
                                   electrical_tilt,
                                   mechanical_downtilt,
                                   date_of_first_service,
                                   date_of_last_service,
                                   geom_point
                                 )
SELECT cell_id AS id,
       version,
       site_id,
       cell_name AS name,
       technology AS type,
       msc,
       bsc_rnc,
       antenna_type,
       status,
       lac,
       height,
       azimuth,
       transmitter,
       max_range,
       min_range,
       electrical_tilt,
       mechanical_downtilt,
       lower(dates_of_service) AS date_of_first_service,
       upper(dates_of_service) AS date_of_last_service,
       ST_SetSRID( ST_Point( longitude, latitude), 4326) AS geom_point
FROM infrastructure.cell_info
WHERE to_include;
```

In the future we could consider making `infrastructure.cells` a materialized view on `infrastructure.cell_info` to aid the update process. Alternatively, we could use `infrastructure.cell_info` directly, if an appropriate index on `to_include` was sufficient to alleviate any performance impacts. Although from an ETL point of view there's something to be said for ingesting into a separate table and then updating `infrastructure.cells` in one hit at the end, to avoid race conditions without having to squeeze the entire ETL process into a single transaction.

### DIFF
--- a/flowdb/bin/build/0020_schema_infrastructure.sql
+++ b/flowdb/bin/build/0020_schema_infrastructure.sql
@@ -182,7 +182,7 @@ CREATE SCHEMA IF NOT EXISTS infrastructure;
         ingested_at TIMESTAMPTZ NOT NULL DEFAULT CURRENT_TIMESTAMP,
         source_filename TEXT,
         file_hash TEXT,
-        file_date DATE,
+        file_date DATE
     );
 
     -- Table to keep records of all cells (including those excluded due to data quality issues, and old versions of cells that have moved/changed).
--- a/flowdb/bin/build/0020_schema_infrastructure.sql
+++ b/flowdb/bin/build/0020_schema_infrastructure.sql
@@ -176,14 +176,24 @@ CREATE SCHEMA IF NOT EXISTS infrastructure;
 
         );
 
+    -- Table to record each time the cell info is updated
+    CREATE TABLE IF NOT EXISTS infrastructure.cells_table_versions (
+        id INTEGER PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
+        ingested_at TIMESTAMPTZ NOT NULL DEFAULT CURRENT_TIMESTAMP,
+        source_filename TEXT,
+    );
+
     -- Table to keep records of all cells (including those excluded due to data quality issues, and old versions of cells that have moved/changed).
     -- Keeping all cell info in this table allows us to reconstruct the cells table as it was at a previous point in time, if necessary.
     CREATE TABLE IF NOT EXISTS infrastructure.cell_info(
-        cell_id TEXT, -- 'id' in infrastructure.cells
-        version INTEGER,
+        id INTEGER PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
+        cells_table_version INTEGER NOT NULL REFERENCES infrastructure.cells_table_versions (id),
+        mno_cell_id TEXT NOT NULL, -- 'id' in infrastructure.cells
+        dates_of_service TSTZRANGE NOT NULL, -- [date_of_first_service,date_of_last_service) in infrastructure.cells
+        to_include BOOLEAN NOT NULL, -- Flag to indicate whether or not cell should be used in analysis
         longitude DOUBLE PRECISION, -- x component of 'geom_point' in infrastructure.cells
         latitude DOUBLE PRECISION, -- y component of 'geom_point' in infrastructure.cells
-        dates_of_service TSTZRANGE, -- [date_of_first_service,date_of_last_service) in infrastructure.cells
+        geom_point geometry(Point, 4326) GENERATED ALWAYS AS ( ST_SetSRID(ST_Point(longitude, latitude), 4326) ) STORED,
         technology TEXT, -- 'type' in infrastructure.cells
         cell_name TEXT, -- 'name' in infrastructure.cells
         mno_site_id TEXT, -- 'site_id' in infrastructure.cells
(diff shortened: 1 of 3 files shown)


greenape:
For the geom_point, one could readily add this as a generated field I think.

My thoughts overall here run something like - is there too much information in this table? Is there any downside in, in effect, creating a new cells table every time you update the cells, and this table recording the metadata for each new table? (As ever what I'm not keen on is any need to use update.)

flowdb/bin/build/0020_schema_infrastructure.sql
```diff
@@ -175,3 +175,38 @@ CREATE SCHEMA IF NOT EXISTS infrastructure;
...
+        electrical_tilt NUMERIC,
+        mechanical_downtilt NUMERIC,
+        to_include BOOLEAN, -- Flag to indicate whether or not cell should be used in analysis
+        first_received DATE, -- Date of first cell info file to include this cell
+        last_received DATE, -- Date of last cell info file to include this cell
+        ingested_at TIMESTAMPTZ, -- Date/time at which this cell was first ingested
+        excluded_at TIMESTAMPTZ, -- Date/time at which this cell was first marked as excluded from analysis
+        notes TEXT, -- Free text field for adding notes related to this cell (e.g. reason for exclusion)
+        additional_metadata JSONB, -- JSON field to catch any fields provided in cell info files that don't fit into the infrastructure.cells table structure
+        PRIMARY KEY (cell_id, version),
```
greenape, line 209: Why not put a unique constraint on these and use a regular serial/uuid key? Makes it more convenient to reference from elsewhere, no?

jc-harrison: Does it? I'm not sure why a serial/uuid key would be more convenient - seems unnecessary to generate a new unique ID when (cell_id, version) already serves that purpose.

jc-harrison:
> My thoughts overall here run something like - is there too much information in this table? Is there any downside in, in effect, creating a new cells table every time you update the cells, and this table recording the metadata for each new table? (As ever what I'm not keen on is any need to use update.)

We could do it that way - create a new table each time we update the cells, and record the file date and update time in a separate metadata table instead of having the first/last received and ingested/excluded at fields in the cells table. I think we'd need to copy the entire table and append any new cells to it each time we update (so essentially the same process as updating the cell_info table as implemented, but in a new copy of the table instead of in-place), because we'd still want to keep the no-longer-active cells in the "latest" cells table even if they're no longer included in the source files, and we'd need to carry across added information (e.g. the `to_include` flag, to avoid re-including cells we have previously excluded).

Duplicating a load of data each time we update the cells feels a bit clumsy, but I'm OK with it as long as it's for good reasons. The main downside I can see is that we then have a brand new cells table each time we update - so e.g. if other tables have a foreign key on the cells table, those foreign keys all need to be re-created every time we update. And if we want to look at the history of the cell info (e.g. to check when cells were added/removed), we first need to get the set of all tables that are previous versions of the cells table.

What's your specific objection to using `update`? It should only be the `to_include`, `excluded_at`, `last_received` and `notes` fields that would ever be updated (in fact, even `to_include` is perhaps redundant since it's equivalent to `excluded_at IS NULL`) - everything else should be append-only (and we could enforce that via privileges if we want).

greenape:
I'm thinking as ever about the caching mechanism, but more generally the table being stateful is a potential source of bugs.

flowdb/bin/build/0020_schema_infrastructure.sql
jc-harrison: Thinking about this some more, maybe we should have a serial key _instead of_ `version`, since we generate `version` ourselves just to differentiate between differing records with the same cell ID, and a globally-unique serial key would do this equally well while also functioning as a standalone primary key.

flowdb/bin/build/0020_schema_infrastructure.sql
```diff
@@ -175,3 +175,43 @@ CREATE SCHEMA IF NOT EXISTS infrastructure;
         hnd_type TEXT
 
         );
+
+    -- Table to record each time the cell info is updated
+    CREATE TABLE IF NOT EXISTS infrastructure.cells_table_versions (
+        id INTEGER PRIMARY KEY GENERATED ALWAYS AS IDENTITY,
+        ingested_at TIMESTAMPTZ NOT NULL DEFAULT CURRENT_TIMESTAMP,
+        source_filename TEXT,
```
greenape, line 183: Just wondering if it would be helpful to record a hash of the file or similar as well?

greenape: And as discussed, file_date as well

greenape (changes requested):
Good to go pending extra fields

flowdb/bin/build/0020_schema_infrastructure.sql
greenape, line 183:
```suggestion
        source_filename TEXT
```

greenape (approved):

greenape:
Let's have a changelog entry as well I think

PR #6448, opened 2024-02-08
Fix no poke

### LINKED ISSUE #5763: FlowETL DataPresentSensor fails if file does not exist
When FlowETL is used to ingest CDR data from a file, the first task creates a foreign table for reading the file contents via `file_fdw`. This can be done before the file arrives - file existence is not checked when creating the foreign table.

The next task is a `DataPresentSensor` sensor which should check whether the file has arrived and contains data, and keep checking until it does. However, if the file has not yet arrived then the task will encounter a psycopg2 exception and fail (rather than rescheduling, as we'd want it to do). This can still function much like a sensor if appropriate choices of the `retries` and `retry_delay` args are used, but this is not the intended purpose of the retry mechanism.

`DataPresentSensor` should register an unsuccessful "poke" if the file is missing (as is the case if the file is present but empty), instead of failing the entire task.

greenape:
Closes #5763

### I have:

### Description

Identifying _why_ a file fdw table can't be read is as it turns out not possible. That means we can't poke only on missing files. Hence what I've done is to make the foreign table wrapper error if the table can't be read, and add an extra sensor which let's you require some number of rows be present rather than just that there's at least a header.

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1203701576305848

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## [Unreleased]
 
 ### Added
+- FlowETL sensor `NRowsPresentSensor` which checks for a specified minimum number of rows.
 
 ### Changed
+- `ForeignStagingTableOperator` will now error if the underlying file cannot be read or the command returns an error. [#5763](https://github.com/Flowminder/FlowKit/issues/5763)
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


Thingus (approved):
lgtm

jc-harrison (commented):
The approach makes sense. Although this still leaves us in the situation of using the fail/retry mechanism in lieu of a sensor; it's just that it will now be the `CreateForeignStagingTableOperator` task that fails and retries until the file appears, rather than the "wait for data" task. I wonder whether we could add a `FileExistsSensor` before creating the foreign table, using `pg_stat_file` or similar?

greenape:
Yeah I think possible - when I originally looked at the docs for stat_file I read it as limited to just the pg directories, but actually on re-reading that's only the case sans super user which the executing user here is.

greenape:
Might be better off with `pg_ls_dir` thinking about it, given that in some cases there are multiple files.

PR #6466, opened 2024-03-06
Break ties in MostFrequentLocation using last event

### LINKED ISSUE #6268: Modal location tiebreaker
As evidenced by #5600, on occasion there may be ties in `ModalLocation`. As it stands, these are broken by arbitrary Postgres sort order, which can change between versions. Rather than rely on that, we should have a known and predictable tie breaker. One might be tempted to use `random` here, but that would actually prevent modal location queries from being parallelised, so probably better to use something else.

I think an obvious way to go here would be to extend the existing tie break logic, which uses the most recent of tied daily locations, to something like the most recent _call_. This would however require daily location objects to include the last included datetime. Arguably this would be to the good anyway, as we could then dispense with the slightly awkward post-attach of the corresponding date.

greenape:
Closes #6268

### I have:

### Description

Changes `MostFrequentLocation` to break ties based on the latest event time as with modal location.


jc-harrison (approved):
Perhaps we should add a note to the changelog that MostFrequentLocation cache tables (and dependents) should be cleared because the results may change? Other than that, this looks good.

PR #6483, opened 2024-03-27
Fix missing migrations

### LINKED ISSUE #6480: FlowAuth migrations missing from docker image
The alembic migrations are not copied to the flowauth image, and as a result no migration occurs. Relatedly, the migrations don't get run when we test on CI because `make_demodata` is a parallel path to `init_db`.

greenape:
Closes #6480 

### I have:

### Description

Fix for a second time the missing flowauth migrations. These are now included in the FlowAuth package, and I've added some logging so one can see them actually being run as well.


flowauth/backend/flowauth/models.py
```diff
@@ -559,15 +559,15 @@ def init_db(force: bool = False) -> None:
...
     if force:
+        # Because we're starting fresh, we don't need to migrate and can mark the
+        # created schema as current
         current_app.logger.debug("Dropping existing db.")
         db.drop_all()
-    db.create_all()
+        db.create_all()
+        flask_migrate.stamp()
+
+    flask_migrate.upgrade()
```
Thingus, line 569: This _should_ be a noop if `force == True`, but do we want in in an `else` clause just in case something goes wrong with a migration script and we need to rebuild?

flowauth/backend/flowauth/models.py
```diff
@@ -608,9 +608,8 @@ def make_demodata():
     if current_app.config["DB_IS_SET_UP"].is_set():
         current_app.logger.debug("Database already set up by another worker, skipping.")
```
Thingus, line 609: Should there be a warning if we're going to overwrite the db with demodata?

Thingus (commented):
Do we want to exclude migrations from codecov tests?

flowauth/backend/flowauth/models.py
greenape: I reckon we want to boot this stuff out of the start up sequence and separate it all out from one another really, but not right now.

Thingus (approved):

PR #6495, opened 2024-04-16
Fix FlowETL Makefile

jc-harrison:
### I have:

### Description

Makes the following changes to the FlowETL Makefile:
- Makes the venv an order-only dependency, so that `make` will not re-compile the `*requirements.txt` files simply because the venv has not been created (as is the case in a fresh clone)
- Adds a `CI` flag which, if set, prevents requrements.txt and dev-requirements.txt from being re-compiled (to prevent inadvertent changes to the locked environment during CI runs)
- Uses `make dev-sync` to install flowetl dependencies in CircleCI (note: I considered going all the way and using `make unit-tests`/`make integration-tests` in CI, but we use additional flags to capture coverage output from the CI test runs so I think it's cleaner to keep this as a distinct step)
- Compiles `requirements.txt` and `dev-requirements.txt` directly from flowetl's `setup.py` instead of from `*.in` files
- Removes `flowpyter-task` from flowetl requirements


jc-harrison:
Using the Makefile in CI is slightly tricky, because `make` uses the file timestamps to determine whether a target needs to be rebuilt, but CircleCI operates on a fresh clone of the repo so all files will have the same timestamp. I've added a `CI` flag to block the Makefile from rebuilding the requirements files during CI (which shouldn't happen anyway, since `make` won't build the requirements files unless one of their dependencies is _newer_, which it won't be in a fresh clone), but I don't think there's any easy way to check whether the requirements files are out of date (e.g. if somebody modifies and commits an input file without re-locking, a local `make` would detect that the requirements file is out of date, but in a new clone it won't, so in CI we wouldn't hit the "requirements.txt is out of date" error. `pipenv` uses hashes to quickly check whether the Pipfile has been modified since the last lock, but `pip-tools` has no such functionality).

PR #6496, opened 2024-04-17
Avoid duplicate QA check tasks in FlowETL

### LINKED ISSUE #6494: get_qa_checks duplicates
**Describe the bug**
Suppose we have the following file structure:

```
- dags_root/
    - qa_checks/
        - my_check.sql
    - my_dag.py
```
where 'my_dag.py' is a DAG definition file that calls `flowetl.util.get_qa_checks()`. If we mount 'dags_root/' to '/opt/airflow/dags/etl/' in a FlowETL container, importing the DAG fails with:
```
Broken DAG: [/opt/airflow/dags/etl/my_dag.py] Traceback (most recent call last):
  File "/home/airflow/.local/lib/python3.8/site-packages/airflow/models/baseoperator.py", line 811, in __init__
    task_group.add(self)
  File "/home/airflow/.local/lib/python3.8/site-packages/airflow/utils/task_group.py", line 231, in add
    raise DuplicateTaskIdFound(f"{node_type} id '{key}' has already been added to the DAG")
airflow.exceptions.DuplicateTaskIdFound: Task id 'my_check' has already been added to the DAG
```

On closer inspection, `template_paths` within `get_qa_checks()` evaluates to:
```
[PosixPath('etl/qa_checks/my_check.sql'), PosixPath('qa_checks/my_check.sql')]
```

jc-harrison:
Closes #6494

### I have:

### Description

- ~Changes `get_qa_checks()` to only add `settings.DAGS_FOLDER` to the template searchpath if `dag.folder` does not already point to the correct DAG folder (which is the case if a DAG is created by the `create_dag()` function).~
- ~Changes the `default_path` in `get_qa_checks()` to always be `Path(__file__).parent`, rather than being either `Path(__file__).parent` or `Path(__file__).parent / "qa_checks"` depending on whether or not the `create_dag()` function is being used. An upshot of this is that the default QA check files no longer need the additional level of nesting that was required to ensure the template paths containd the string "qa_checks".~
- Fixes `flowetl.util.create_dag` so that the returned DAG has the correct `dag.folder`
- Removes awkward workaround in `flowetl.util.get_qa_checks` that was necessary due to the `dag.folder` sometimes being incorrect
- Adds integration tests for loading QA checks from the DAG folder

Overall, I think we still want to re-work the QA check discovery more substantially - the template-path-contains-"qa_checks" criterion and the silent addition of a path within the FlowETL module to the template searchpath both leave us open to unexpected and hard-to-debug bugs, and it would be useful to have more explicit control over which QA checks are added to a DAG. But this PR fixes the immediate issue, to enable the use of custom QA checks again.

Note: it would still be possible to get duplicates if the user defining a DAG specifies a `template_searchpath` or `additional_qa_check_paths` containing either a parent or child of the DAG folder - this PR does not prevent errors in that situation.

### DIFF
--- a/flowetl/tests/unit/test_qa_check_discovery.py
+++ b/flowetl/tests/unit/test_qa_check_discovery.py
@@ -60,7 +60,7 @@ def test_additional_checks_collected(tmpdir):
     assert len(check_operators) > len(qa_checks)
 
 
-def test_additional_checks_collected_from_dag_folder(monkeypatch):
+def test_additional_checks_collected_from_dag_folder():
     from airflow import DAG, settings
     from flowetl.util import get_qa_checks
 
@@ -69,7 +69,7 @@ def test_additional_checks_collected_from_dag_folder(monkeypatch):
     checks_folder.mkdir(parents=True)
     (checks_folder / "DUMMY_CHECK.sql").touch()
     dag = DAG("DUMMY_DAG", start_date=datetime.now())
-    monkeypatch.setattr(dag, "folder", str(dag_folder))
+    dag.fileloc = dag_folder / "DUMMY_DAG.py"
     check_operators = get_qa_checks(dag=dag)
 
     assert len(check_operators) > len(qa_checks)


jc-harrison:
I've changed to a different approach from the one I started off using. The underlying problem here was that Airflow assumes the DAG file (i.e. the file that needs to be imported to define the DAG) is the file in which the DAG constructor is called, but when using FlowETL's `create_dag` function the DAG file is actually one frame back (i.e. the file in which `create_dag` is called, not the file that defines `create_dag`). Since `dag.folder` (i.e. the parent of the DAG file) is the default template searchpath, we had to add a workaround in `get_qa_checks` for the case that `dag.folder` wasn't actually pointing to the DAG folder (which happened if the DAG was created using `create_dag`), and that workaround resulted in QA checks defined in the DAG folder being picked up twice. For the same reason, we also had to add the entire contents of the FlowETL module to the template searchpath rather than just the path to the default QA checks.

To get around this, I've updated `create_dag` to explicitly set `dag.fileloc` to be the file location from which `create_dag` is called. This feels a bit hacky, but it's exactly the same hack that's used in `DAG.__init__()`, and overall I think this is the cleanest solution short of a more substantial re-work of the QA check discovery mechanism to avoid needing to walk the entire template searchpath during DAG creation.

greenape (approved):
I find this solution distressing, but the best available option as it stands.

PR #6581, opened 2024-05-08
Deduped recieved scopes

### LINKED ISSUE #6580: Same time queries producing duplicate scopes
**Describe the bug**
When a flowkit query contains multiple sub-queries that are identical except for the date, flowAPI parses this to multiple identical scopes

**Product**
FlowAPI

**Version**
1.21.1

Thingus:
Closes #6580 

### I have:

### Description

Passes the output of `flowapi.permissions.query_to_scopes` through a `set()` before returning


flowapi/flowapi/permissions.py
```diff
@@ -210,7 +210,9 @@ async def query_to_scopes(query_dict):
     except Exception:
         raise BadQueryError
     agg_unit = await get_agg_unit(query_dict)
-    return [f"{agg_unit}:{tl_query_name}:{query_name}" for query_name in query_list]
+    return list(
```
jc-harrison, line 213: Do we need this to be a list, or could we return a set instead? (I don't think it matters much either way, but if we return a set it's perhaps clearer that the scopes will be non-duplicated)

Thingus: It's compared to other lists when we check it against the requested scopes; now I say that, a set would be better there too

jc-harrison (approved):
:+1:

PR #6627, opened 2024-05-24
Add invalid_cell_info table and change exclude constraint on cell_info

### LINKED ISSUE #6626: Disallow duplicates in `infrastructure.cell_info`
#6433 added a new table `infrastructure.cell_info` to FlowDB for recording the full history of cell information, including cells that have been excluded due to quality concerns.

The original intention behind the design of this table was for it to include _all_ cell information, including records with duplicate cell IDs. As such, the constraint that ensures no two simultaneously-valid cells have the same ID is only applied over rows where `to_include = True`, so that duplicates can be included in the table provided they are excluded from use in analysis.

On reflection, I think this was a bad design decision - when ingesting new cell information, we want to join to the previous cell information (including "excluded" cells so that we can carry over these exclusions and avoid re-including cells), and this join is made more complicated by the presence of duplicates in the cell info table. I think we would do better to make a distinction between "valid-but-excluded cells" (such as those with suspicious longitude/latitude coordinates), which should be included in `infrastructure.cell_info` to avoid re-including excluded cells in future updates, and "invalid cell records" (such as duplicates, or cell records with null cell ID - which cannot be included in `infrastructure.cell_info` due to non-null constraint), which should perhaps be kept in a separate table so we do not lose the information, but do not need to be in `infrastructure.cell_info`.

jc-harrison:
Closes #6626

### I have:

### Description

- Changes the exclude constraint on `infrastructure.cell_info` to ensure cell ID is unique across _all_ simultaneously-valid cells, not just `WHERE to_include`.
- Adds a new table `infrastructure.invalid_cell_info`, with similar structure to `infrastructure.cell_info` but fewer constraints, to hold any rows from cell information files that are not valid entries in `infrastructure.cell_info` (including cells with null or duplicate cell IDs).

----

| :rocket: | This description was created by [Ellipsis](https://www.ellipsis.dev) for commit c837b579942d793b3309a46531a6c7506dd70049  | 
|--------|--------|

### Summary:
This PR modifies the exclusion constraint on `infrastructure.cell_info` and introduces a new table `infrastructure.invalid_cell_info` to enhance data integrity and tracking of invalid cell records.

**Key points**:
- Modified `EXCLUDE` constraint on `infrastructure.cell_info` for cell ID uniqueness across all valid cells.
- Introduced `infrastructure.invalid_cell_info` table for storing invalid cell records.
- Updated `CHANGELOG.md`.

----
Generated with :heart: by [ellipsis.dev](https://www.ellipsis.dev)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## [Unreleased]
 
 ### Added
+- Added FlowDB table `infrastructure.invalid_cell_info` for recording cell information that could not be included in `infrastructure.cell_info` (including cells with null or duplicate cell IDs). [#6627](https://github.com/Flowminder/FlowKit/issues/6627)
 
 ### Changed
 - FlowDB now triggers an ANALYZE on newly created cache tables to generate statistics rather than waiting for autovacuum
@@ -20,6 +21,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Changed
 - FlowDB now enables partitionwise aggregation planning by default
 - FlowDB now uses a default fillfactor of 100 for cache table indexes
+- `EXCLUDE` constraint on FlowDB `infrastructure.cell_info` table requires unique `mno_cell_id` across _all_ simultaneously-valid cells per `cells_table_version`, regardless of `to_include`. [#6627](https://github.com/Flowminder/FlowKit/issues/6627)
 
 ### Fixed
 - Queries that have multiple of the same subquery with different parameters no longer cause duplicate scopes in tokens. [#6580](https://github.com/Flowminder/FlowKit/issues/6580)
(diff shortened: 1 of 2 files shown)


greenape (approved):

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## [Unreleased]
 
 ### Added
+- Added FlowDB table `infrastructure.invalid_cell_info` for recording cell information that could not be included in `infrastructure.cell_info` (including cells with null or duplicate cell IDs). [#6627](https://github.com/Flowminder/FlowKit/issues/6627)
```
jc-harrison, line 10:
```suggestion
- Added FlowDB table `infrastructure.invalid_cell_info` for recording cell information that could not be included in `infrastructure.cell_info` (including cells with null or duplicate cell IDs). [#6626](https://github.com/Flowminder/FlowKit/issues/6626)
```

PR #6789, opened 2024-09-09
Fix str enum

### LINKED ISSUE #6787: InOutFlow broken with python >=3.11
**Describe the bug**
`flowmachine.features.location.flows.InOutFlow` has a `direction` parameter to specify whether the query is an inflow or an outflow. We use the following enum to ensure the direction is either `"to"` or `"from"`:

```
class Direction(str, Enum):
    OUTFLOW = "from"
    INFLOW = "to"
```

The `str` mixin ensured that `Direction` objects could be used as strings, e.g. `f"pcod_{self.direction}"` formats to either `"pcod_from"`  or `"pcod_to"`.

But python 3.11 changes this behaviour (see https://blog.pecar.me/python-enum for a description of the change), so that `f"pcod_{self.direction}"` formats to `"pcod_Direction.INFLOW"` or `"pcod_Direction.OUTFLOW"`, which breaks the in/outflow SQL query.

Python 3.11 introduces a `StrEnum` class that can be used instead. So either we can make the definition of `Direction` conditional on the python version, or we can explicitly use `direction.value` instead of `direction` for string formatting.

I suspect there are other places we use the `class MyEnum(str, Enum)` hack, which may also be broken.

greenape:
Closes #6787

### I have:

### Description

Replaces usages of `str, Enum` with `StrEnum` in flowmachine, and changes the minimum python version to 3.11


Thingus (approved):
lgtm

PR #7280, opened 2026-04-28
ci: pin pipenv version in flowmachine Docker + CI to prevent hash drift

jakejellinek:
Follow-up to #7278.

## Summary

Pin pipenv version in both contexts that interact with `flowmachine/Pipfile.lock`, so the "different pipenv versions hash the same Pipfile differently" deadlock fixed in #7278 cannot recur.

## Changes

1. **`flowmachine.Dockerfile`**: replace `pip install pipenv` with `pip install pipenv==${PIPENV_VERSION}` (defaulted via `ARG` to `2026.6.1` so a future bump is a single ARG override or `--build-arg`).

2. **`.circleci/config.yml` install_flowmachine_deps**: pin to the same version, **and** include the pipenv version in the cache key (`flowmachine-deps-pipenv2026.6.1-…`). Without that second change, dependabot could float lock content forever without ever exercising `pipenv install --deploy`, because the cache-key checksum alone wouldn't change on a transitive bump — exactly the trap that hid the marshmallow `missing=` removal for an unknown amount of time.

3. **Cross-reference comments** at each pin site point at the other, so an upgrader doesn't bump one in isolation.

## Why this PR is small

This is the minimal cut. There are other reasonable preventions worth considering separately:

- **Pin pipenv across other components** (flowdb tests at `.circleci/config.yml:188-189`, etc.). Same class of issue could in principle affect them. Out of scope here because each component has its own lock + deploy path and would benefit from a tailored review.
- **Tighter upper bounds in `Pipfile`** (e.g. `marshmallow = ">=3.0.0,<4"`). A genuine reliability improvement, but the right bound is per-dep and reviewer-dependent — better as a separate discussion.
- **Generate the lock in CI rather than `--deploy`-ing it**. James flagged this in his Feb 2026 handover. Larger lift; would entirely eliminate version-drift but is a workflow change.

## Upgrading pipenv

Bump the `ARG PIPENV_VERSION` default in `flowmachine.Dockerfile`, the `pipenv==` literal in `.circleci/config.yml`, and the version segment of both `restore_cache` / `save_cache` keys, all in one commit. Comments at each site list the others.

## Test plan

- [ ] `build_flowmachine` and `install_flowmachine_deps` go green on this PR.
- [ ] Deliberately bumping the pinned version in a follow-up PR triggers a fresh install (cache miss), proving the cache key works as intended.

## Depends on

Branched off `flowmachine/refresh-pipfile-lock` (PR #7278) since it builds on those changes. Once #7278 merges, the base will move to `master` and rebase will leave just this commit.


jakejellinek:
Superseded by #7278's revised approach.

The revised #7278 already pins pipenv version in both `flowmachine.Dockerfile` and `install_flowmachine_deps` — to `2024.4.1` (the version that matches the existing lock hash) rather than to a newer pipenv that would invalidate the cache and surface latent dependency bit-rot. So this PR's pinning change is redundant.

The cache-key-includes-pipenv-version idea from this PR is still potentially useful as forward protection when pipenv is next bumped, but the right time to add it is alongside that bump (so the cache invalidation is intentional). Filing as future work rather than landing now in isolation.

Closing without merge.

